PR #16, opened 2019-02-28
Updated Community page to make 3rd link Participate in FDC3

brooklynrob:
Updated Community page in help.js to make 3rd link Participate in FDC3


brooklynrob:
Replacing with subsequent PR that has this commit and another together.

PR #17, opened 2019-02-28
Updated Community Page; Updated Navigation

brooklynrob:
Updated Community page to add link to FDC3 wiki

Updated navigation to replace About with Why FDC3

### DIFF
--- a/website/siteConfig.js
+++ b/website/siteConfig.js
@@ -33,8 +33,8 @@ const siteConfig = {
   headerLinks: [
     {doc: 'fdc3-intro', label: 'Docs'},    
     {doc: 'use-cases/use-cases-intro', label: 'Use Cases'},
-    {page: 'help', label: 'Community'},
-    {page: 'help', label: 'About'}
+    {page: 'help', label: 'FDC3 Community'},
+    {doc: 'why-fdc3', label: 'Why FDC3'}
    //, {blog: true, label: 'News'}
   ],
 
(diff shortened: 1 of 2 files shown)


website/siteConfig.js
```diff
@@ -33,8 +33,8 @@ const siteConfig = {
   headerLinks: [
     {doc: 'fdc3-intro', label: 'Docs'},    
     {doc: 'use-cases/use-cases-intro', label: 'Use Cases'},
-    {page: 'help', label: 'Community'},
-    {page: 'help', label: 'About'}
+    {page: 'help', label: 'FDC3 Community'},
+    {doc: 'why-fdc3', label: 'Why FDC3'}
```
rikoe: I was thinking of doing this slightly differently.

Move the content, but still make it an About web page, not a doc. For me this is equivalent to the FINOS programme explanation for what FDC3 is about, without necessarily taking the format of documentation.

Example:
https://docusaurus.io/en/about-slash

brooklynrob: @rikoe I like that idea. In the meantime, perhaps we can do with my proposed change temporarily? We now have live, public traffic coming to the site and I'd like to suggest we get us much "low hanging fruit" cleaned up ASAP so that the site is passable. That we have/had "Community" and "About" pointing to the same paces is something I think should be cleaned up sooner than later.

rikoe: Done!

PR #41, opened 2019-03-07
Remove unneccesary dates from filenames

tschady:


tschady:
Note: This is redundant with the 8 changes @jonathanteperJPMC just make PRs for.

PR #44, opened 2019-03-07
Rename 2019-02-26-013-user-launches-multiple-apps-from-a-single-conta…

jonathanteperJPMC:
…iner.md to 013-user-launches-multiple-apps-from-a-single-container.md


rikoe:
Addressed by #41

PR #52, opened 2019-03-12
Update fdc3-charter.md note spacing

sbloodgood:
Under Scope, line break for note section to separate from interop bulletpoint.


rikoe:
@sbloodgood If you want this change to apply to 1.0 release and not just to "next", you also need to change https://github.com/FDC3/FDC3/blob/master/website/versioned_docs/version-1.0/fdc3-charter.md

sbloodgood:
Apologies, just submitted.

rikoe:
@sbloodgood if you make a change to an existing PR, you can just push it to the same branch again, it will update automatically and include the new change as well. No need to raise a separate one! That way everything being discussed and changed stays in the same context.

PR #64, opened 2019-03-21
Switching domain (from fdc3.finos.org) to fdc3.org

maoo:
fdc3.org currently redirects to fdc3.finos.org . To complete the switch, it should be the other way around. This PR is to deploy docusaurus using fdc3.org as domain.

As soon as the PR is merged, I will make the DNS adjustments to finalize the change.

Read more on https://finosfoundation.atlassian.net/browse/ODP-79

### DIFF
--- a/CNAME
+++ b/CNAME
@@ -1,1 +1,1 @@
-fdc3.finos.org
+fdc3.org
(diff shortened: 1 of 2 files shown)


rikoe:
Will fdc3.finos.org continue to redirect to fdc3.org? IPR agreements have been signed and press releases prepared based on the current URLs. Maybe we should wait until after 1.0 for this?

PR #65, opened 2019-03-22
Fix intent examples in docs

### LINKED ISSUE #63: The Intents Overview page Using Intents samples don't match the API
The example to `open` that we have [here](https://fdc3.finos.org/docs/1.0/intents-intro#directing-a-market-data-platform-to-show-a-chart) doesn't match the API that we have defined [here](https://fdc3.finos.org/docs/1.0/api/DesktopAgent#open)

rikoe:
Fix #63 by using correct intent APIs

### DIFF
--- a/docs/intents-intro.md
+++ b/docs/intents-intro.md
@@ -19,24 +19,40 @@ FDC3 Intents define a standard set of verbs that can be used to put together com
 ## Using Intents
 Combined with [Context Data](context-intro.md) and [App Directory](appd-intro.md) standards, Intents enable rich service discovery on the desktop. For example:
 
-### Directing a market data platform to show a chart
+### Ask for a chart to be displayed
 ```javascript
-fdc3.open("my-platform","ViewChart",{
- type:"fdc3.instrument",
+const result = await fdc3.raiseIntent("ViewChart", {
+ type: "fdc3.instrument",
  name: "IBM",
- id:{
+ id: {
     ticker:"ibm"
-    }
-  });
+  }
+});
 ```
 
-### Discovering apps that can start a chat
+### Ask a specific application to display a chart
 ```javascript
-fdc3.raiseIntent("StartChat",{
- type:"fdc3.contact",
+const result = await fdc3.raiseIntent("ViewChart", {
+ type: "fdc3.instrument",
+ name: "IBM",
+ id: {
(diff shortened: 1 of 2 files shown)


docs/intents-intro.md
```diff
@@ -19,24 +19,40 @@ FDC3 Intents define a standard set of verbs that can be used to put together com
...
+ name: "IBM",
+ id: {
+    ticker:"ibm"
+  }
+}, "market-data-app");
+```
+
+### Find applications that can start a chat
+```javascript
+const intentApps = await fdc3.findIntent("StartChat");
```
RichLinnell: The `findIntents` call returns an object, not the actual apps :
```
// I know 'StartChat' exists as a concept, and want to know more about it ...
const appIntent = await agent.findIntent("StartChat");
// returns a single AppIntent:
// {
//     intent: { name: "StartChat", displayName: "Chat" },
//     apps: [{ name: "Skype" }, { name: "Symphony" }, { name: "Slack" }]
// }

// raise the intent against a particular app
await agent.raiseIntent(appIntent.intent.name, context, appIntent.apps[0].name);
```

docs/intents-intro.md
```diff
@@ -19,24 +19,40 @@ FDC3 Intents define a standard set of verbs that can be used to put together com
...
+```
+
+### Find available intents for a contact
+```javascript
+const intentsAndApps = await fdc3.findIntentsByContext({
```
RichLinnell: This is fine, but I noticed that the example in the api docs for `findIntentsByContext` says `findIntentsForContext` - we need to change that.

docs/intents-intro.md
rikoe: I know, but this isn't the API docs, I wanted to be concise and unspecific and get the idea across of what it turns. Do you think I should change it? It could still be an object with that name in my example.

RichLinnell: Cool - just checking.

docs/intents-intro.md
rikoe: Good catch, I noticed the same while doing this and then promptly forgot to change it! Will update the PR...

RichLinnell: I can do that on a separate commit if you like

PR #67, opened 2019-03-24
Add user showcase to website

rikoe:
* Based on user data in website/data/users.json
* Click through to users page that lists all FDC3 users
* Includes flag to pin users to main page (“isPinned”)
* Includes flag to distinguish users who are members (“isMember”)
* Added “Proud member of FINOS” banner above user showcase on main page

Main webpage:

image

Users page:

image


nkolba:
@rikoe can we add Citadel to the list and remove the "All FDC3 Users" CTA for now (since the lists are going to be the same).  Also, would be great if we could get similar sizings around the logo content.  The images are the same sizes, but some have quite different margins.

PR #70, opened 2019-03-26
Context channels API proposal

rikoe:
This API proposes a simple interface for dealing with channel updates / context sharing.

Some notes:
1. In this example channels are simply string names, but channels could potentially be richer objects
2. The first application to use a channel with a particular name (e.g. 'blue' or 'common') creates it.
3. Each channel represents a backing store of context objects indexed by their type. 
4. For each channel, you can have one "instrument" type, one "contact" type, etc. If a different "contact" type is pushed, this represents an updated value for that type.
5. The methods are deliberately not called "publish" and "subscribe" to indicate that there is a persistence store in play.

Some further topics to be discussed:
* Can context values be persisted between sessions?
* Should there be an option to retrieve only the current value without updates, or ignore the current value and only getting new updates.
* Should multiple different instances of the same type be supported on the same channel? (Gets complicated quickly...)
* Does it make sense for context channels to be defined as part of application definitions, the same way intents are?

Example use case (selecting a row in the blotter updates the context in other applications):

Context Sharing

### DIFF
--- a/examples/api/channels.ts
+++ b/examples/api/channels.ts
@@ -1,0 +1,22 @@
+const fdc3: DesktopAgent;
+
+async function channelExamples() {
+    
+    // ['blue', 'purple', 'green']
+    const allChannels = await fdc3.channels.list();
+
+    await fdc3.channels.update('blue', {
+        type: 'fdc3.contact',
+        name: 'Riko Eksteen',
+        id: {
+            email: 'riko@weareadaptive.com'
+        }
+    });
+
+    // called every time fdc3.contact type is updated on the blue channel
+    const contactListener = fdc3.channels.watch('blue', 'fdc3.contact', contact => {
+        console.log("Selected contact is" + contact.name);
+    });
+
+    contactListener.unsubscribe();
+};
--- a/src/api/channels.ts
+++ b/src/api/channels.ts
@@ -1,0 +1,17 @@
+interface Channels {
+
+    /**
+     * Retrieve a list of currently known channels
+     */
+    list(): Promise<string[]>;
+
+    /**
+     * Update a particular context value on a particular channel. Context values are indexed by their type.
+     */
+    update(channelName: string, context: Context): Promise<void>;
+
+    /**
+     * Watch for updates to a particular context type on a particular channel.
+     */
+    watch(channelName: string, type: string, handler: ContextHandler): Listener;
+}
--- a/src/api/interface.ts
+++ b/src/api/interface.ts
@@ -1,4 +1,5 @@
 type Context = object;
+type ContextHandler = (context: Context) => void;
 
 enum OpenError {
   AppNotFound = "AppNotFound",
@@ -174,4 +175,6 @@ interface DesktopAgent {
    * Adds a listener for incoming context broadcast from the Desktop Agent.
    */
   addContextListener(handler: (context: Context) => void): Listener;
+
+  channels: Channels;
 }


espenove:
Makes a lot of sense, liked the point about not using "pub/sub". Would be good for the discussion to clarify why we need separate context objects for each channel (and simply not only a single object per channel), I suggest we at least keep with a single object per type (else this quickly turns into a full database/mini git like store)
Persisting between session is powerful, could also be used for sharing/storing workspaces with channel/objects, but I suggest we have this optional, in other words not part of the first version of this addition.
Context channels as par of app definition: propose to start simple and not make part of standard for first version.

RichLinnell:
Great work Riko. 
I echo the comments made by Espen regarding the “depth” of the store. Storing history for each context-item type in a channel leads to unnecessary complexity. This is about sharing “current” context between applications, not providing a convenient mini-database.

rikoe:
The ability to share context by type is a very important one, arguably the main feature of this API. E.g. imagine a blotter that contains both a contact and instrument - when the row is selected, two separate context objects are updated, the instrument and the contact.

A market data app can watch only for instrument updates on the same channel, and a CRM app can watch only for contact updates.
- If all updates are sent on the channel regardless, both the market data and CRM app would have to do further filtering to ignore unwanted updates.
- If there is only one object that can be current at any point in time, you would need to use object wrapping or nesting to make it work.

We have found that two levels of grouping - by "channel" or "region", and then by object type - offers just the right mixture of granularity and simplicity to make the use cases work.

Sharing by type is part both of the Finsemble LinkerClient API, e.g.
```js
FSBL.Clients.LinkerClient.publish({
    channels: ["blue"],
    dataType:"symbol",
    data:"AAPL"
});

FSBL.Clients.LinkerClient.subscribe("symbol", data => { 
    // do something
});
```
and of the Glue42 context sharing API.

vmehtafds:
I think from understanding we have three requirements based on the call today (was there any requirement missed?):
* Limiting the routing of messages based on channel id/color/object/theme
* Getting the last cached value when some app joins the channel
* Limiting the data app gets by a type

It would be good differentiate where the responsibilities/code is separated between the DesktopAgent and the App. As an app developer and not a DesktopAgent developer this is important for us to clarify.
* Does app need to know what channel(s) they are in?
  * Ideally it would be nice that the app doesn't care what channel(s) they are in or have to care about  changing of channels as long they get the correct data.
* Does app need to know if the data they received was through a current broadcast or through last cached value?
  * Ideally an app would have one mechanism/api/callback to get the data for broadcast and initial startup/subscription to channel.
* Who does the limiting of data received by type?
  * App certainly would need to specify/know the "types" it cares about. Channel/DesktopAgent API could allow app to specify these types (hopefully only once for both broadcast and initial subscription) and then the Agent/ChannelsAPI restrict the data it sends to the app. I think most were not in the favor of of sending everything to the apps and let app ignore what it doesn't want.

donbasuno:
Nice work Riko!

Couple of questions for my understanding:

1. If an application understands both types (eg fdc3.instrument and fdc3.instrumentList), should then two listeners be created?

2. If your application handles "all" types - would that be possible with by basically creating a listener for "blue" but without specifying the type?

3. Technically, an application could listen to several colour channels?

4. Could all of the above lead to confusion and race conditions regarding what context to actually use and thus create further complications for the API?

Thanks,

Johan

rikoe:
> I think from understanding we have three requirements based on the call today (was there any requirement missed?):
> 
> * Limiting the routing of messages based on channel id/color/object/theme
> * Getting the last cached value when some app joins the channel
> * Limiting the data app gets by a type

@vmehtafds that is very well summarised.

> * Does app need to know what channel(s) they are in?

Not necessarily - but I don't like magic as a developer. If an app is filtering messages by channel, I would like it to be right there, visible and explicit in the code. It feels like "not getting certain messages because I applied a filter in some other unrelated place" is a great recipe for subtle bugs.

>   * Ideally it would be nice that the app doesn't care what channel(s) they are in or have to care about  changing of channels as long they get the correct data.

See above - I would prefer what channels an app are participating in to be explicit.

> * Does app need to know if the data they received was through a current broadcast or through last cached value?

A good question. There may be some advantage in distinguishing with an extra property that tells you. I have also seen people ask to receive only the most recent change, only future updates, or both. I didn't include this in my proposal so as not to overcomplicate the discussion.

>   * Ideally an app would have one mechanism/api/callback to get the data for broadcast and initial startup/subscription to channel.

Yes.

> * Who does the limiting of data received by type?

An application that subscribes to updates can specify which types it is interested in. How we have implemented this in the past is that the receiving application effectively "registers" with the desktop agent the types it is interested in. This greatly cuts down on the number of messages flowing through the system, as only interested applications get certain updates, and is a big advantage of this approach.

>   * App certainly would need to specify/know the "types" it cares about. Channel/DesktopAgent API could allow app to specify these types (hopefully only once for both broadcast and initial subscription) and then the Agent/ChannelsAPI restrict the data it sends to the app. I think most were not in the favor of of sending everything to the apps and let app ignore what it doesn't want.

Yes, as stated above, controlling the amount of data sent between applications is a big benefit. Similar to intents, only registered/known applications and types will be involved.

rikoe:
> 1. If an application understands both types (eg fdc3.instrument and fdc3.instrumentList), should then two listeners be created?

That is the intention - when we first implemented such an API we had one callback for all types, but then realised this callback always just contained a giant switch statement distinguishing by type. So it seemed to make the code cleaner and more isolated to have a callback per type.

This doesn't preclude having a callback for "all types" or a "set of types" of course, for the people interested in that.

> 2. If your application handles "all" types - would that be possible with by basically creating a listener for "blue" but without specifying the type?

Yes. I didn't include it in this example, but our real implementation had a handler for "all" types as well for the cases where that makes sense.

> 3. Technically, an application could listen to several colour channels?

Yes. This API enables coloured channels, but is not directly equivalent to coloured channel linking. For channel linking you would have one channel per application, but there are plenty of other use cases where a dedicated channel for a particular type of "hidden update" (i.e. under the covers) makes sense.

> 4. Could all of the above lead to confusion and race conditions regarding what context to actually use and thus create further complications for the API?

I don't see how it could lead to race conditions. What could happen is if a particular application subscribes to an update, and then updates the same piece of data inside the handler, thereby leading to a loop of updates.

But it is relatively easy to serialise updates from various applications to ensure they arrive in order.

rikoe:
Closing in favour of #90

PR #73, opened 2019-03-27
updated text

nkolba:


website/pages/en/index.js
```diff
@@ -160,7 +160,7 @@ class Index extends React.Component {
       return (
         <div className="productShowcaseSection paddingBottom">
           <h2>Who is Using FDC3?</h2>
-          <p>FDC3 is used by several industry-leading organisations...</p>
+          <p>The Financial Desktop Connectivity and Collaboration Consortium (FDC3) was founded in 2017 by <a href="https://openfin.co">OpenFin</a> and contributed to <a href="https://finos.org">FINOS</a>.  FDC3 standards are created and used by leading organizations across the financial industry</p>
```
rikoe: I don't personally think this line belongs here, as it is a bit out of context, and the front page is deliberately sparse. The About page is the perfect place for it.

nkolba: There are actually a lot of examples of this kind of explanation on sites for open source projects.  For example, on electronjs.org, just above the "Apps Build on Electron" section, you have this blurb:  "With Electron, creating a desktop application for your company or idea is easy. Initially developed for GitHub's Atom editor, Electron has since been used to create applications by companies like Microsoft, Facebook, Slack, and Docker."    If you go to chromium.org, it says this at the top of the page: "The Chromium projects include Chromium and Chromium OS, the open-source projects behind the Google Chrome browser and Google Chrome OS, respectively. This site houses the documentation and code related to the Chromium projects and is intended for developers interested in learning about and contributing to the open-source projects."

rikoe: Cool. I think this is fine for now, it could potentially move up to the banner at the top of the page, but we can iterate.

PR #90, opened 2019-04-24
Proposed specification update to support context sharing

rikoe:
This is a work-in-progress proposal to facilitate discussion for context sharing support in the FDC3 API.

Outstanding tasks:
- [x] Update context sharing description in specification
- [x] Add new methods to `src/api/interface.ts`
- [ ] Add new methods to `docs/api/DesktopAgent.md`
- [ ] Add usage examples
- [ ] Create separate PR for context channel support, which is now independent

### DIFF
--- a/docs/api/api-spec.md
+++ b/docs/api/api-spec.md
@@ -126,11 +126,52 @@ It is expected that App Directories will also curate listed apps and ensure that
 
 Like FDC3 Context Data, the Intent schemas need to be versioned.  Desktop Agents will be responsible to declare which version of the Intent schema they are using.   Applications may also assert a specific version requirement when raising an Intent.  Version negotation may be supported by a given Desktop Agent.
 
-### Send/broadcast context  
-On the financial desktop, applications often want to broadcast context to any number of applications.  Context sharing needs to support concepts of different groupings of applications as well as data privacy concerns.  Each Desktop Agent will have its own rules for supporting these features.  
-
-## Resolvers
+### Resolvers
 Intents functionality is dependent on resolver functionality to map the intent to a specific App.  This will often require end-user input.  Resolution can either be performed by the Desktop Agent (raising UI to pick the desired App for the intent) or by the app launching the intent - in which case the calling App will handle the resolution itself (using the findIntents API below) and then invoke an explicit Intent object.
+
+## Context Sharing (DRAFT)
+
+> [IMPROVE DESCRIPTION]
+>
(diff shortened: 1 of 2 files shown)


docs/api/api-spec.md
```diff
@@ -126,12 +126,78 @@ It is expected that App Directories will also curate listed apps and ensure that
...
+```ts
+const instrumentListener = desktopAgent.addTypedContextListener('fdc3.instrument', instrument => {
+    // handle instrument object
+});
+```
+
+If an application would like to retrieve the current state of the desktop at startup, they can ask for the current value for a particular type of object:
+
+```ts
+await instrument = desktopAgent.getCurrentValue('fdc3.instrument');
```
ColinEberhardt: `getCurrentValue` doesn't really tell you much about what this function does. I'd suggest `getContext` or `getContextValue` - the fact that it is current is implicit.

docs/api/api-spec.md
```diff
@@ -126,12 +126,78 @@ It is expected that App Directories will also curate listed apps and ensure that
...
+```ts
+const listener = desktopAgent.addContextListener(context => {
+    // handle context object
+});
+```
+
+If an application declares a typed context listener, the callback will only be invoked for context object updates where the type matches:
+
+```ts
+const instrumentListener = desktopAgent.addTypedContextListener('fdc3.instrument', instrument => {
```
ColinEberhardt: Similar to @nkolba - I'd vote for overloading `addContextListener` to allow both typed and non-typed subscriptions.

docs/api/api-spec.md
```diff
@@ -126,12 +126,78 @@ It is expected that App Directories will also curate listed apps and ensure that
...
+
+If an application declares a generic context listener, providing a callback, they will be notified whenever another application invokes the `broadcast` function for any object:
+
+```ts
+const listener = desktopAgent.addContextListener(context => {
+    // handle context object
+});
+```
+
+If an application declares a typed context listener, the callback will only be invoked for context object updates where the type matches:
```
ColinEberhardt: > where the type matches

Can you provide more clarity on this? is a match string equality? Or something more complex (e.g. matching a pattern or hierarchy?)

rikoe: The intention was just string equality for now. I think if a particular implementation would like to perform more complicated matching, they could, but the spec should not mandate it.

But in any case, will make this more explicit.

PR #91, opened 2019-05-07
addIntentListener handler return type

spcfran:
If IntentResolution can have a data payload, it seems there's a missing data payload coming from the intent listener handler.

- It still accepts returning `void` as the simpler use case
- Otherwise it returns a `Promise`
- The Promise resolves to an object with a `data` property, instead of just `any`, to give the API room to grow.


nkolba:
will pick this up as a wider discussing with Intents & App Directory

PR #94, opened 2019-05-09
Proposed API changes

tim-dinsdale:
For discussion, not final


src/api/channel/channel.ts
```diff
@@ -0,0 +1,179 @@
...
+     * 
+     * The `channel` property within the event will always be this channel instance.
+     */
+    public addEventListener(event: 'broadcast', listener: (event: {channel: Channel; context: Context}) => void): void;
+
+    /**
+     * Remove this event listener
+     * 
+     */
+    public removeEventListener(event: 'broadcast', listener: (event: {channel: Channel; context: Context}) => void): void;
```
ColinEberhardt: Rather than remove by reference, I'd recommend adopting the existing pattern, where `addEventListener` returns a `Listener` object that allows you to unsubscribe:

https://github.com/FDC3/FDC3/blob/master/src/api/interface.ts#L171

docs/api/api-spec.md
```diff
@@ -132,6 +132,46 @@ On the financial desktop, applications often want to broadcast context to any nu
...
 
+## Context channels
+
+The context channel api allows a set of apps to share a stateful piece of data between them, and be alerted when it changes.
+
+There are three types of channels, which are functionally identical, but have different visibility and discoverability semantics.
+
+1. The 'public' or 'desktop' ones, which have a well understood identity.
+2. The 'private' or 'app' ones, which have a transient identity and need to be revealed
+3. The 'global' one which is available to everyone.
```
rikoe: As discussed on the API WG call, can we please simply this to "public channels, of which one is called Default", and "Private channels"

docs/api/api-spec.md
```diff
@@ -132,6 +132,46 @@ On the financial desktop, applications often want to broadcast context to any nu
...
+
+Additionally there is a 'default' channel which serves as /dev/null and satisfies a possible requirements that all apps connect to one channel.
+
+To open a desktop channel, one calls
+
+    let c = channels.getDesktopChannels().find(c=>???);
```
rikoe: Is the intention that the find function is a lambda/callback? I think it will be better if you can just pass a name or indexer.

docs/api/api-spec.md
```diff
@@ -132,6 +132,46 @@ On the financial desktop, applications often want to broadcast context to any nu
...
+
+To broadcast one calls
+
+    c.broadcast(context);
+
+To subscribe one calls
+
+    c.addEventListener('broadcast', (c,e)=>{some code})
```
rikoe: If there are well-known event types, I think an enum would be great, as opposed to people having to remember strings. Although underlying parameter can still be a string though.

docs/api/api-spec.md
```diff
@@ -132,6 +132,46 @@ On the financial desktop, applications often want to broadcast context to any nu
...
+
+Private are created as thus:
+
+    let ac = AppChannel.create();
+    let channel_name = ac.id;
+    function_to_send_name_to_other_app(channel_name);
```
rikoe: As discussed on today's call, I don't like the idea of a randomly generated channel name that then has to be transferred to the other app with a separate call.

If people can pre-agree private channels, it will be simpler and more consistent with how public channels work.

docs/api/api-spec.md
```diff
@@ -132,6 +132,46 @@ On the financial desktop, applications often want to broadcast context to any nu
 ## Resolvers
 Intents functionality is dependent on resolver functionality to map the intent to a specific App.  This will often require end-user input.  Resolution can either be performed by the Desktop Agent (raising UI to pick the desired App for the intent) or by the app launching the intent - in which case the calling App will handle the resolution itself (using the findIntents API below) and then invoke an explicit Intent object.
 
+## Context channels
+
```
rikoe: In general I think we should keep the spec relatively high level - it should explain intention in clear terms, without too many code samples beyond making that intention clear, and we should try to keep technical terms like `dev/null` out of it.

The right places for usage samples is in the `examples` folder and code comments I think. (I realise that we don't currently have other API usage samples.)

src/api/channel/channel.ts
```diff
@@ -0,0 +1,179 @@
...
+     * Returns the last context that was broadcast on this channel. All channels initially have no context, until a 
+     * window is added to the channel and then broadcasts. If there is not yet any context on the channel, this method
+     * will return `null`. The context is also reset back into it's initial context-less state whenever a channel is
+     * cleared of all windows.
+     * 
+     * The context of a channel will be captured regardless of how the context is broadcasted on this channel - whether
+     * using the top-level FDC3 `broadcast` function, or using the channel-level {@link broadcast} function on this 
+     * object.
+     * 
+     * NOTE: Only non-default channels are stateful, for the default channel this method will always return `null`.
```
rikoe: `null` or `undefined`? The latter feels more like idiomatic JavaScript to me - someone could explicitly set the value to `null`.

src/api/channel/channel.ts
```diff
@@ -0,0 +1,179 @@
...
+
+    /**
+     * A user-readable name for this channel, e.g: `"Red"`
+     */
+    name: string;
+
+    /**
+     * The color that should be associated within this channel when displaying this channel in a UI, e.g: `0xFF0000`.
+     */
+    color: number;
```
rikoe: Do we want to bake the concept of a colour into the underlying FDC3 API? For me this is an implementation concern that sits on top of FDC3. E.g. a particular implementation can extend the base channel type with an extra color property if it needs it.

ColinEberhardt: Agreed, also encoding a colour as a number feels ambiguous, as it lacks details of the encoding scheme

docs/api/api-spec.md
ColinEberhardt: The find functions is from Array.find - which makes sense, no need to create a new API here.

However, the example is wrong, the Array is the ‘payload’ of the resolves promise, so you’ll need a ‘then’ or ‘await’ in this code sample.

docs/api/api-spec.md
```diff
@@ -132,6 +132,46 @@ On the financial desktop, applications often want to broadcast context to any nu
...
+The context channel api allows a set of apps to share a stateful piece of data between them, and be alerted when it changes.
+
+There are three types of channels, which are functionally identical, but have different visibility and discoverability semantics.
+
+1. The 'public' or 'desktop' ones, which have a well understood identity.
+2. The 'private' or 'app' ones, which have a transient identity and need to be revealed
+3. The 'global' one which is available to everyone.
+
+Additionally there is a 'default' channel which serves as /dev/null and satisfies a possible requirements that all apps connect to one channel.
```
zhenyakovalyov: Could you please clarify the use-case for the default channel?

src/api/channel/channel.ts
```diff
@@ -0,0 +1,156 @@
+type ChannelId = string;
+
+declare interface ChannelListener{
```
ColinEberhardt: replace with `Listener`: https://github.com/FDC3/FDC3/blob/master/src/api/interface.ts#L55

src/api/channel/channel.ts
```diff
@@ -0,0 +1,156 @@
...
+declare class DesktopChannelVisualIdentity{
+    /**
+     * A user-readable name for this channel, e.g: `"Red"`
+     */
+    name: string;
+
+    /**
+     * The color that should be associated within this channel when displaying this channel in a UI, e.g: `0xFF0000`.
+     */
+    color: number;
```
ColinEberhardt: I'd suggest using a string, as per CSS, allowing `red`, `#fff`, `rgb(255,0,255)` etc ...

src/api/channel/channel.ts
```diff
@@ -0,0 +1,156 @@
...
+     * The `channel` property within the event will always be this channel instance.
+     */
+    public addBroadcastListener(listener: (event: {channel: Channel; context: Context}) => void): ChannelListener;
+}
+
+/**
+ * A desktop channel will be global enough to have a presence across many apps. This gives us some hints
+ * to render them in a standard way.
+ */
+declare class DesktopChannelVisualIdentity{
```
ColinEberhardt: probably better defined as an interface

docs/api/api-spec.md
```diff
@@ -132,6 +132,34 @@ On the financial desktop, applications often want to broadcast context to any nu
...
+The context channel api allows a set of apps to share a stateful piece of data between them, and be alerted when it changes.
+
+There are three types of channels, which are functionally identical, but have different visibility and discoverability semantics.
+
+1. The 'public' or 'desktop' ones, which have a well understood identity. One is called 'default'.
+2. The 'private' or 'app' ones, which have a transient identity and need to be revealed
+
+The 'public' channels include a 'default' channel which serves as the backwards compatible layer with the 'send/broadcast context' above. There are some reserved channel names for future use. Currently this is just 'global'.
+
+To find a desktop channel, one calls
```
vmehtafds: I am trying to understand how apps are going to use this API. Looking forward to some pseudo code demonstrating the three use cases we talked in the meeting.

I would like to know how an app associates itself with a channel or does a DesktopAgent puts an app in a channel (i.e. in the later case app knows nothing about channels). I was imagining some sort of `joinChannel()` api somewhere.

Also how this separate `channel.broadcast()` function is going to work. Is this something an app will call? I think the idea was app doesn't even need to know about channels.  Will an app just use regular top level `fdc3.broadcast()` and the DesktopAgent intercept that and will then use the `channel.broadcast()` to route it accordingly? How are those two separate `broadcast()` play nicely with each other.

Same thing about `addBroadcastListener()` vs `addContextListener()`. What goes in the app vs the desktopAgent?

Hopefully some code around use cases will clear these things.

src/api/channel/channel.ts
tim-dinsdale: This is done

docs/api/api-spec.md
tim-dinsdale: Ok, this is resolved by calling the function addBroadcastListener (which, actually, it should have been called from the beginning, sorry about that)

docs/api/api-spec.md
tim-dinsdale: dev/null has been sent to dev/null

src/api/channel/channel.ts
tim-dinsdale: Agreed, done

docs/api/api-spec.md
tim-dinsdale: This fragment of doc is now out of date - there is no 'global', just the 'default' channel which acts as a way of retconning the old FDC3 global context into this scheme. Are you ok with that?

src/api/channel/channel.ts
tim-dinsdale: Done

src/api/channel/channel.ts
tim-dinsdale: Done

src/api/channel/channel.ts
tim-dinsdale: Agreed, done

docs/api/api-spec.md
tim-dinsdale: In terms of an app associating itself with a channel, something in the app will call `getDesktopChannels`, will find the channel it needs by checking the members of the `DesktopChannelVisualIdentity` interface, and will squirrel that away, plus the listener it got from subscribing to it.

In terms of the classic `fdc3.broadcast()` vs `channel.broadcast()`: the code is suggesting that one of the desktopChannels will be called 'default', and that if you call `fdc3.broadcast()` you will send to this channel and vice-versa. In other words, that channel is a facade to the existing single global context.

src/api/channel/channel.ts
```diff
@@ -0,0 +1,157 @@
...
+/**
+ * A desktop channel will be global enough to have a presence across many apps. This gives us some hints
+ * to render them in a standard way. It is assumed it may have other properties too, but if it has these, 
+ * this is their meaning.
+ */
+declare interface DesktopChannelVisualIdentity{
+    /**
+     * A user-readable name for this channel, e.g: `"Red"`
+     */
+    name?: string;
```
rikoe: `displayName` would be slightly more descriptive, as discussed on our recent working group call.

src/api/channel/channel.ts
```diff
@@ -0,0 +1,157 @@
+type ChannelId = string;
+
+declare interface Listener{
```
rikoe: We should reuse https://fdc3.finos.org/docs/1.0/api/DesktopAgent#listener for this, which is already ratified as part of v1.0 spec.

docs/api/api-spec.md
```diff
@@ -132,6 +132,34 @@ On the financial desktop, applications often want to broadcast context to any nu
...
+## Context channels
+
+The context channel api allows a set of apps to share a stateful piece of data between them, and be alerted when it changes.
+
+There are three types of channels, which are functionally identical, but have different visibility and discoverability semantics.
+
+1. The 'public' or 'desktop' ones, which have a well understood identity. One is called 'default'.
+2. The 'private' or 'app' ones, which have a transient identity and need to be revealed
+
+The 'public' channels include a 'default' channel which serves as the backwards compatible layer with the 'send/broadcast context' above. There are some reserved channel names for future use. Currently this is just 'global'.
```
ColinEberhardt: > There are some reserved channel names for future use. Currently this is just 'global'.

For this to be workable, _all_ reserved words need to be declared at the point this API is released, otherwise application developers aren't able to avoid naming collisions.

src/api/channel/channel.ts
```diff
@@ -0,0 +1,157 @@
...
+
+    /**
+     * Uniquely defines each channel type.
+     * 
+     * See overrides of this class for list of allowed values.
+     */
+    public readonly type: string;
+
+    /**
+     * Broadcasts the given context on this channel. This is equivilant to joining the channel and then calling the
```
ColinEberhardt: s/equivilant/equivalent

src/api/channel/channel.ts
```diff
@@ -0,0 +1,157 @@
...
+ * 
+ * All interactions with a context channel happen through the methods on here.
+ */
+declare class Channel {
+    /**
+     * Constant that uniquely identifies this channel. Will be generated by the service, and guarenteed to be unique
+     * within the set of channels registered with the service.
+     * 
+     * In the case of `desktop` channels (see {@link DesktopChannel}), these id's _should_ persist across sessions. The 
+     * channel list is defined by the service, but can be overridden by a desktop owner. If the desktop owner keeps
```
ColinEberhardt: This is introducing a new concept "desktop owner", which isn't defined in our spec:

https://fdc3.finos.org/docs/1.0/api/api-spec

Any new terminology should be clearly defined in the documentation!

rikoe:
@nkolba @tim-dinsdale Next we need to get the website documentation updated for this.

PR #101, opened 2019-07-09
added .png versions of Users - changed users.json files to .png

grizzwolf:
added .png versions of Users - changed users.json files to display .png instead of .webp - iOS doesn't currently support .webp (USER images not showing up on iPhone, but fine on desktop)


grizzwolf:
ok @rikoe - I believe this PR will fix the .png / .webp issue - and added to the README.md - let me know if any issues. Thanks!

PR #135, opened 2019-12-03
FDC3 Survey Links

grizzwolf:
Adding FDC3 Survey Links in Marquee and temp using FeatureCallout for Survey (Hid Use Cases Temporarily)


website/pages/en/index.js
```diff
@@ -137,11 +140,20 @@ class Index extends React.Component {
 
     const FeatureCallout = () => (
       <div  className="featureShowcaseSection  paddingBottom" style={{textAlign: 'center'}}>
-        <h2>Use Cases</h2>
-        <MarkdownBlock>{`Document business [use cases](${docUrl('use-cases/overview')}) that drive FDC3 interoperability standards.`}</MarkdownBlock>
+        <h2>FDC3 Survey</h2>
```
rikoe: Does this change get rid of the use cases banner? Not sure we want that... Ideally a second banner?

grizzwolf: It's a temporary change to highlight the survey for the next week (right?). I just hid the Markdown block so we can put it back up as soon as survey is done. I was asked to get this up quickly to help with draw attention to the survey. I tried a couple of other ways at first, but kept having issues getting them to work.

donbasuno: Use Cases are still available in the top toolbar so given timing here - I am ok with temporarily removing the use case banner.

Thanks,

Johan

nkolba:
@grizzwolf @rikoe @donbasuno 
I created an alternate CTA layout here:  https://github.com/FDC3/FDC3/pull/136

brooklynrob:
Here's how this looks on my laptop after this PR was approved.

image

PR #153, opened 2020-01-10
Added UC17 for review and vote

donbasuno:


jonathanteperJPMC:
All in favor! Use case ratified :)

PR #154, opened 2020-01-17
1.1 api simple

nkolba:
The purpose of this pull request is the following:

- More fully document the channels apis, adding documentation into the API Reference section
- Simplifying some aspects of the channels API.  Specifically:
   - Removing the Channels namespace, and moving some methods to the top-level fdc3 space
   - Rationalizing system and app channels - so that they are only different by convention
   - Added an error enum for channels, and made it explicit that the desktop agent could reject access to a channel based on the permissions of the app


docs/api/api-spec.md
```diff
@@ -134,31 +134,56 @@ Intents functionality is dependent on resolver functionality to map the intent t
...
+
+### Examples
 To find a system channel, one calls
 
-    let allChannels = await channels.getSystemChannels();
-    let myChannel = allChannels.find(c=>???);
+```javascript
+    //returns an array of channels
+    let allChannels = await fdc3.getSystemChannels(); 
+    let redChannel = allChannels.find(c=>{return c.id === "red";});
```
rikoe: This can be simplified to `allChannels.find(c => c.id === 'red')`, which will make it much easier to read.

docs/api/api-spec.md
```diff
@@ -134,31 +134,56 @@ Intents functionality is dependent on resolver functionality to map the intent t
...
+```
+
+To join a channel. one calls
```
rikoe: I think this needs a sub-heading: "Joining channels"

docs/api/api-spec.md
```diff
@@ -134,31 +134,56 @@ Intents functionality is dependent on resolver functionality to map the intent t
...
-    myChannel.broadcast(context);
+To get (or create) a channel reference, then interact with it
 
-To subscribe one calls
+```javascript
+    const appChannel = await fdc3.getOrCreateChannel("appChannel");
+    //get the current context of the channel
+    let current = await appChannel.getCurrentContext();
+    //add a listener
+    appChannel.addBroadcastListener((event)=>{...});
```
rikoe: This doesn't need the extra brackets around the parameter: `appChannel.addBroadcastListener(event => { ... })`

docs/api/api-spec.md
```diff
@@ -134,31 +134,56 @@ Intents functionality is dependent on resolver functionality to map the intent t
...
 
-To broadcast one calls
+Calling _fdc3.broadcast_ will now route context to the joined channel.
 
-    myChannel.broadcast(context);
+To get (or create) a channel reference, then interact with it
 
-To subscribe one calls
+```javascript
+    const appChannel = await fdc3.getOrCreateChannel("appChannel");
```
rikoe: I would put `fdc3.getOrCreateChannel('my_custom_channel')` or something similar to highlight that this is not a well-known channel name, but is bespoke.

docs/api/api-spec.md
```diff
@@ -134,31 +134,56 @@ Intents functionality is dependent on resolver functionality to map the intent t
...
+
+```javascript
+    fdc3.joinChannel(redChannel.id);
+```
 
-To broadcast one calls
+Calling _fdc3.broadcast_ will now route context to the joined channel.
 
-    myChannel.broadcast(context);
+To get (or create) a channel reference, then interact with it
```
rikoe: This needs a sub-heading: "App channels", and a further explanation about app channels being custom channels that apps can define for interaction.

nkolba:
thanks @rikoe!  I've made the recommended changes and pushed to my branch

src/api/interface.ts
```diff
@@ -179,4 +270,29 @@ interface DesktopAgent {
...
+   */
+  getSystemChannels(): Promise<Array<Channel>>;
+
+  /**
+   * Joins the app to the specified channel
+   * An app can only be joined to one channel at a time
+   * rejects with error if the channel is unavailable or the join request is denied
+   * `Error` with a string from the `ChannelError` enumeration.
+   */
+  joinChannel(channelId: string) : Promise<void>;
```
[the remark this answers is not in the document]
rikoe: I think the reason is that if it is at the top level you could just join a channel by id  without having to do a “get and search” first, so it massively simplifies the code for someone who just wants to join a channel and then use the existing top-level broadcast or context listener.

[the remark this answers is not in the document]
rikoe: I think the reason is that if it is at the top level you could just join a channel by id  without having to do a “get and search” first, so it massively simplifies the code for someone who just wants to join a channel and then use the existing top-level broadcast or context listener.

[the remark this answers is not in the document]
nkolba: Yes, exactly.

docs/api/DesktopAgent.md
```diff
@@ -138,6 +140,66 @@ agent.raiseIntent("StartChat", newContext, intentR.source);
...
+```
+Returns a Channel object for the specified channel, creating it (as an _App_ channel) - if it does not exist.
+ `Error` with a string from the [`ChannelError`](Errors#channelerror) enumeration if channel could not be created or access was denied.
+
+#### Examples
+```javascript
+
+try {
+  let myChannel = await fdc3.getOrCreateChannel("myChannel");
+  myChannel.addContextListener(listener);
```
rikoe: This is not correct. Should be: `const listener = myChannel.addContextListener(context => { })`

Issue #156, opened 2020-01-22
target parameter of raiseIntent does not correspond to name of AppMetadata


rikoe:
`raiseIntent` as defined as follows:

```ts
raiseIntent(intent: string, context: Context, target?: string): Promise<IntentResolution>;
```

`AppMetadata` as returned by `findIntentsByContext` as defined as follows:

```ts
interface AppMetadata {
  name: string;
}
```

The intent (har-har) is that you can use the app identifier returned by the discovery to target a specific app for raising an intent. But if `AppMetadata` returns a "name", this cannot be done, instead the unique app identifier should be use for the `target` parameter of `raiseIntent`, but it is not returned.

rikoe:
Closing, as `name` is the correct field to use with `raiseIntent`, I was confused.

PR #157, opened 2020-01-22
Align intent and application definition between schemas

### LINKED ISSUE #19: Align AppMetadata and IntentMetadata with other FDC3 specifications
The [`AppMetadata`][1] and [`IntentMetadata`][2] returned by [`findIntent`][3] and [`findIntentsByContext`][4] should be aligned with the [App Directory Specification][5] and [Intents Specification][6], respectively.

[1]:https://fdc3.finos.org/docs/1.0/api/DesktopAgent#appmetadata
[2]:https://fdc3.finos.org/docs/1.0/api/DesktopAgent#intentmetadata
[3]:https://fdc3.finos.org/docs/1.0/api/DesktopAgent#findintent
[4]:https://fdc3.finos.org/docs/1.0/api/DesktopAgent#findintentsByContext
[5]:https://fdc3.finos.org/1.0/appd-spec#tag/Application/paths/~1v1~1apps~1{appId}/get
[6]:https://fdc3.finos.org/docs/1.0/intents-spec

rikoe:
Fixes #19 and #156


src/api/interface.ts
```diff
@@ -21,24 +21,50 @@ enum ResolveError {
...
+  appId: string;
   name: string;
+  title?: string;
+  tooltip?: string;
+  description?: string;
+  images?: Array<AppImage>;
+  icons?: Array<Icon>;
+  intents?: Array<Intent>;
+  customConfig?: Array<{name: string, value: string}>;
+}
```
rikoe: Happy to exclude some of these properties, or include some more - it doesn't have to be an exact 1-to-1 match with the AppD schema, only metadata that will be useful for applications using the APIs.

For example, here I have left out `manifest` and `manifestType` (which are mandatory in the AppD schema), as I don't see what use they would have outside of the App Directory. But I am not against including them as well for consistency...

src/api/interface.ts
```diff
@@ -21,24 +21,50 @@ enum ResolveError {
...
 }
 
 /**
- * App metadata is Desktop Agent specific - but should support a name property.
+ * App definition as provided by the application directory
  */
-interface AppMetadata {
+interface Application {
+  appId: string;
   name: string;
```
rikoe: What is the difference and intended usage of `name` vs `appId` (the AppD schema requires both of them to be unique). Which one is supposed to be used with `raiseIntent`'s `target` parameter?

PR #158, opened 2020-01-24
compare join to being on the Channel object vs the global

nkolba:
As discussed, this shows moving the 'joinChannel' method from the global fdc3 object to a 'join' method on the Channel class. @pjbroadbent

### DIFF
--- a/docs/api/api-spec.md
+++ b/docs/api/api-spec.md
@@ -170,10 +170,14 @@ To find a system channel, one calls
 To join a channel. one calls
 
 ```javascript
-    fdc3.joinChannel(redChannel.id);
+    redChannel.join();
+
+    //join the 'default' channel to effectively reset channel membership
+    const defaultChannel = await fdc3.getOrCreateChannel('default');
+    defaultChannel.join();
 ```
 
-Calling _fdc3.broadcast_ will now route context to the joined channel.
+Calling _fdc3.broadcast_ will now route context to the `red` channel.
 
 #### App Channels
 
--- a/src/api/interface.ts
+++ b/src/api/interface.ts
@@ -96,6 +96,14 @@ declare class Channel {
   displayMetadata?: DisplayMetadata;
 
    /**
+   * Joins the app to the channel
+   * An app can only be joined to one channel at a time
+   * rejects with error if the channel is unavailable or the join request is denied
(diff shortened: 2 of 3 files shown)


docs/api/api-spec.md
```diff
@@ -170,10 +170,14 @@ To find a system channel, one calls
 To join a channel. one calls
 
 ```javascript
-    fdc3.joinChannel(redChannel.id);
+    redChannel.join();
+
+    //join the 'default' channel to effectively reset channel membership
+    const defaultChannel = await fdc3.getOrCreateChannel('default');
```
nicholasdgoodman: This example seems to counter the original documentation for `getOrCreateChannel(...)` which describes them as "these custom channels" and presumably set apart from the predefined system channels returned by `fdc3.getSystemChannels()`.

src/api/interface.ts
```diff
@@ -130,6 +138,8 @@ declare class Channel {
    * `Error` with a string from the `ChannelError` enumeration.
    */
   public addBroadcastListener(listener: (event: {channel: Channel; context: Context}) => void): Listener;
```
nicholasdgoodman: I missed this when the original PR came in, but why is the instance-based listener hook up called `addBroadcastListener` where the global-based listener subscription is called `addContextListener`?

nkolba:
Thanks for the review @nicholasdgoodman .  These are excellent points.  I completely agree on the separation of concerns.

src/api/interface.ts
nkolba: yes, we discussed this on thursday and are going to make everything 'addContextListener' across the board.

docs/api/api-spec.md
nkolba: with the new PR, there is no class differentiation between system and app channels - just  convention based.  While call 'getOrCreateChannel' for a system channel isn't necessary, it doesn't hurt anything either.

rikoe:
@nicholasdgoodman I thought you summarised this very well in terms of separation of concerns. After that I feel quite strongly like we should stick to the existing design with `fdc3.joinChannel` at the top level.

nkolba:
Closing this PR - consensus seems to be that we should keep joinChannel as is.

Issue #121, opened 2019-10-16
Modifications to the Channels API + Window selectors.


chedzoir:
There a couple of alterations that I'd suggest making to the API following my attempts at implementing it within our setup and how we expect it will actually get used

1. Can the default channel be be hidden behind a `getter` that returns a promise. That would make it's use consistent with the `getOrCreate` and `getSystemChannels`.

2. Could we modify the `addBroadcastListener` to take an optional array of context types. Then the interop broker would only notify the listener in the event of one of those specified types being transmitted. 
This gives the following advantages
* As it's possible to put anything on a channel, receiving applications will all have to put in their place their own filtering mechanisms to only receive the items that they 
* Allows applications to put different context types onto a channel that can be picked up by different consuming applications. For example, two different instruments can be transmitted by app and treated in different ways by a number of different receiving apps - some may want both, some one, some the other.
There is a downside here in that the channel will need to keep a history of the last received item by type, instead of just the last item - but that would just be a case of holding them within a map

3. Following on from 2, we would then need to add the same context type to the `getCurrentContext` method

4. In context of a requirement that we have with regards to window based channel selectors, and following on from https://github.com/FDC3/FDC3/issues/118 I would propose the following approach to managing window selectors

On the channels module in https://github.com/FDC3/FDC3/blob/master/src/api/channel/channel.ts 
add a function something along the lines of 

`public addWindowChannelListener(listener: (event: {channel: Channel; context: Context}) => void, types: string[]): DesktopAgent.Listener;`

Then the application does not need to know about or worry about the action of selecting a channel. The swapping between channels can be handled within the broker/router code.

If the user sets the channel to "off" then the application just doesn't receive anything
A `DesktopAgent.Listener` is returned should the application decide that it wishes explicitly stop receiving anything regardless.

pjbroadbent:
**Context filtering**
In terms of adding a context argument to `addBroadcastListener`, that could certainly be useful - but would make it inconsistent with `addIntentListener` - would we also add a context argument to intent listeners too? I'm also assuming you're referring to both the "channel" `addBroadcastListener` and the "top level" `addBroadcastListener`?

There's also an implicit assumption here that context filtering would only ever happen on the `type` field - that application's won't want to say "I'll allow any context type that has an ISIN code in it's `id` section", for example. To clarify I'd prefer context filtering to be entirely `type`-based, as implied by the proposed changes, but worth considering any side-effects of that approach.

**Window-level context listeners**
[...]

PR #161, opened 2020-01-25
Add context filtering to API and improve docs

rikoe:
Addresses #121.

A summary of the type changes I introduced:

* I introduced a new type for consistent context handling:
```ts
type ContextHandler = (context: Context) => void;
```

* Changed the channel broadcast listener to be consistent with the rest of the API:
 ```ts
addBroadcastListener(listener: (event: {channel: Channel; context: Context}) => void): Listener;
```
becomes:
```ts
```
The channel parameter is completely redundant here, as we already now what the channel is from the owning object - if a reference is needed, it can be captured with a closure. This brings the handler in line with the one on the desktop agent, which makes for much easier consumption.

* Added a new overload for the top-level context listener and channel context listener:
```ts
addContextListener(contextType: string, handler: ContextHandler): Listener;
```

* Added an optional parameter when getting the current context:
```ts
getCurrentContext(contextType?: string): Promise<Context|null>;
```

Apart from that I have factored out and improved the relevant documentation sections on the website.

### DIFF
--- a/src/api/interface.ts
+++ b/src/api/interface.ts
@@ -4,6 +4,7 @@
  */
 
 type Context = object;
+type ContextHandler = (context: Context) => void;
 
 enum OpenError {
   AppNotFound = "AppNotFound",
@@ -90,10 +91,10 @@ declare class Channel {
   public readonly type: string;
 
   /**
-   * Channels may be visualized and selectable by users. DisplayMetaData may be used to provide hints on how to see them.
+   * Channels may be visualized and selectable by users. DisplayMetadata may be used to provide hints on how to see them.
    * For app channels, displayMetadata would typically not be present
    */
-  displayMetadata?: DisplayMetadata;
+  public readonly displayMetadata?: DisplayMetadata;
 
    /**
    * Broadcasts the given context on this channel. This is equivalent to joining the channel and then calling the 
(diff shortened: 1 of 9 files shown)


src/api/interface.ts
```diff
@@ -102,34 +103,36 @@ declare class Channel {
...
-   * This broadcast will be received by all windows that are members of this channel, *except* for the window that
-   * makes the broadcast. This matches the behaviour of the top-level FDC3 `broadcast` function.
    * `Error` with a string from the `ChannelError` enumeration.
    */
   public broadcast(context: Context): Promise<void>;
 
   /**
    * Returns the last context that was broadcast on this channel. All channels initially have no context, until a 
-   * window is added to the channel and then broadcasts. If there is not yet any context on the channel, this method
-   * will return `null`. The context is also reset back into it's initial context-less state whenever a channel is
```
rikoe: @nkolba I have removed the second part - seems like an implementation detail to me. FDC3 has no concept of "windows", so it shouldn't be in there.

src/api/interface.ts
```diff
@@ -102,34 +103,36 @@ declare class Channel {
    * Note that this function can be used without first joining the channel, allowing applications to broadcast on
    * channels that they aren't a member of.
    * 
-   * This broadcast will be received by all windows that are members of this channel, *except* for the window that
-   * makes the broadcast. This matches the behaviour of the top-level FDC3 `broadcast` function.
```
rikoe: @nkolba I have removed this part - seems like an implementation detail to me. FDC3 has no concept of "windows", so it shouldn't be in there.

Issue #118, opened 2019-08-26
Turning off context sharing via Channels API


vmehtafds:
### Description of Problem:
We have a use case that we currently support with our internal Channels where a user can turn off context sharing via the UI. We have exposed that as OFF channel in our channel selector.
![image](https://user-images.githubusercontent.com/45774571/63723333-2d3e9f00-c823-11e9-996d-80b425c2c14c.png)

In the FDC3 land, we have our apps which use the top level FDC3 API ([broadcast](https://fdc3.finos.org/docs/1.0/api/DesktopAgent#broadcast) and [addContextListener](https://fdc3.finos.org/docs/1.0/api/DesktopAgent#addcontextlistener)) to send and receive context. They are unaware of Channels API and would work if they wrapped in a TitleBar that has Channels support vs a TitleBar that doesn't have channels support.
The TitleBar is what interacts with the Channels api and exposes a UI (dropdown select) to pick channel and that sets the channel for the app. The TitleBar acts as the router for the traffic.

Currently the way to have the off channel or turning off the context sharing based on some UI in the TitleBar (i.e. an OFF option in the dropdown) is by making the app aware of the TitleBar and removing the context listener when that OFF option is selected. This prevents our initial goal of app being unaware of Channels.

### Potential Solutions:
There are few solutions.
* The TitleBar could create a AppChannel with an unique id per window and use it as the "OFF" channel for that window. As it is an unique id that no other window is aware of, no one can broadcast or listen to context shared by that window. - No API changes required. This is supported with current API.
* Similar to `default` channel, create a new System-level `off` channel.
* Add a `myChannel.turnOff()` API

1christianhall:
As discussed briefly in the call today the question is whether it is the API's job or the implementer's to provide a way to mute/filter a channel to ensure that the channel is used or not.

The gut feel from folks on the call is that the channels API assumes that if a channel is being used, it is needed...so in your example above, you'd simply "detach" your title bar in this case from the channel when the GUI indicated the widget no longer should participate on a channel.

nkolba:
This feels like an implementation feature.  

Also, it seems that at least some of the issue is due to the ambiguity around the "default" channel.  The purpose of the default is to provide backwards compatibility for the 1.0 broadcast/addContextListener apis.  Maybe an approach would be to add a _addChannelListener_ method that only listened to system channels (not default)?  Another approach would be to extend addContextListener with an options arg that supported configuration of the listener.  E.g.

[...]

PR #168, opened 2020-02-07
updated spec with more detailed explanation of the 'default' channel

nkolba:
addressing questions raised in https://github.com/finos/FDC3/issues/118

### DIFF
--- a/docs/api/api-spec.md
+++ b/docs/api/api-spec.md
@@ -150,9 +150,35 @@ The 'system' channels include a 'default' channel which serves as the backwards 
 ### Joining Channels
 Apps can join channels.  An app can only be joined to one channel at a time.  When an app joins a channel it will automatically recieve the current context for that channel, except if the channel joined is the 'default'.
 
-When an app is joined to a channel, calls to fdc3.broadcast and listeners added through fdc3.addContextListener will be routed to that channel.  If an app is not explicitly joined to a channel, it is on the 'default' channel.  It is up to the desktop agent to determine the behavior of the 'default' channel, and if context is automaticaly broadcast over it or not.
+When an app is joined to a channel, calls to fdc3.broadcast and listeners added through fdc3.addContextListener will be routed to that channel.  If an app is not explicitly joined to a channel, it is on the 'default' channel.  
+The `default` channel has special behavior that makes it different from other channels.  See a detailed description [below](#default-channel-behavior).
 
 It is possible that a call to join a channel could be rejected.  If for example, the desktop agent wanted to implement controls around what data apps can access.  
+
+### *default* Channel Behavior
+Joining channels in FDC3 is intended to be a behavior initiated by the end user. For example: by color linking or apps being grouped in the same workspace.  Most of the time, it is expected that apps will be joined to a channel by mechanisms outside of the app.  Always, there SHOULD be a clear UX indicator of what channel an app is joined to. 
+
+The one exception is the `default` channel.  Because apps are automatically added to the `default` channel, the message routing behavior is different in order to prevent poor user experiences from unexpected context behavior.  The behavior for the `default` channel is:
+
+- All broadcasts (`fdc3.broadcast` when the app is on default, or `default.broadcast`) update the `default` context
+- `fdc3.addContextListener` will NOT receive context events when the app is joined to the `default` channel
+- Explicit listening on the `default` channel through `default.addContextListener` and calling `default.getCurrentContext` will return context updates for the channel.
+
+#### Examples
+
+An app gets the current context of the `default` as a fallback if no other context is set.
+
+```js
+let dChan = await fdc3.getOrCreateChannel("default");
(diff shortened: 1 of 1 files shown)


docs/api/api-spec.md
```diff
@@ -150,10 +150,36 @@ The 'system' channels include a 'default' channel which serves as the backwards
...
 
 It is possible that a call to join a channel could be rejected.  If for example, the desktop agent wanted to implement controls around what data apps can access.  
 
+### *default* Channel Behavior
+Joining channels in FDC3 is intended to be a behavior initiated by the end user. For example: by color linking or apps being grouped in the same workspace.  Most of the time, it is expected that apps will be joined to a channel by mechanisms outside of the app.  Always, there SHOULD be a clear UX indicator of what channel an app is joined to. 
+
+The one exception is the `default` channel.  Because apps are automatically added to the `default` channel, the message routing behavior is different in order to prevent poor user experiences from unexpected context behavior.  The behavior for the `default` channel is:
+
+- All broadcasts (`fdc3.broadcast` when the app is on default, or `default.broadcast`) update the `default` context
+- `fdc3.addContextListener` will NOT receive context events when the app is joined to the `default` channel
```
rikoe: I don't mind this, but it is kind of unexpected behaviour, so we should be really clear that this is the case, and discuss it with the API working group.

It is also different from how `broadcast` and `addContextListener` worked in version 1.0, so we should clarify that this is a breaking change, and consider making the version 2.0.0 instead of 1.1.0 to indicate this.

docs/api/api-spec.md
```diff
@@ -150,10 +150,36 @@ The 'system' channels include a 'default' channel which serves as the backwards
 ### Joining Channels
 Apps can join channels.  An app can only be joined to one channel at a time.  When an app joins a channel it will automatically recieve the current context for that channel, except if the channel joined is the 'default'.
 
-When an app is joined to a channel, calls to fdc3.broadcast and listeners added through fdc3.addContextListener will be routed to that channel.  If an app is not explicitly joined to a channel, it is on the 'default' channel.  It is up to the desktop agent to determine the behavior of the 'default' channel, and if context is automaticaly broadcast over it or not.
+When an app is joined to a channel, calls to fdc3.broadcast and listeners added through fdc3.addContextListener will be routed to that channel.  If an app is not explicitly joined to a channel, it is on the 'default' channel.  
+The `default` channel has special behavior that makes it different from other channels.  See a detailed description [below](#default-channel-behavior).
```
rikoe: The special behaviour of the `default` channel should also be documented clearly in the API documentation, not only in the spec, as people are more likely to read the docs than the whole spec.

The spec doesn't clarify if the `default` channel is returned by `getSystemChannels`?

nkolba: Agreed about updates to the API docs - will do.

Re: `getSystemChannels` - there is definitely a gap here.   There are arguments either way - would be great to get the opinions of others who have already built or are using channel linking today.

docs/api/api-spec.md
nkolba: Definitely this is significant and needs to be vetted by the WG.  I'm OK with adding language calling out that this may be a breaking change for some (or event many) 1.0 implementations.  However, I don't believe that this is a breaking change for the spec requiring a 2.x designation - since the 1.0 spec intentionally leaves this behavior [open-ended](https://fdc3.finos.org/docs/1.0/api/api-spec#send-broadcast-context).

vmehtafds: Consider an simple application on 1.0 without any UI/frame/functionality for channels and all it does it `fdc3.broadcast` and `fdc3.addContextListener`.
Can it move to 1.1 without changing code and work as expected? Based on the above, the app would stop receiving any context but it can still send context on `default` channel. Is that correct? That seems confusing.

I think the difference in behavior of `fdc3.addContextListener` and `fdc3.broadcast` around `default` channel is confuisng.

vmehtafds: All my apps have just following two calls in their code base:
`fdc3.broadcast` and `fdc3.addContextListener`. They know nothing about being in a channel or not (which I think is very important).
I have a frame around that does channel management. 
My apps can be in that custom frame (where they get channels) or in regular DesktopAgent/Windows frame (where there is no channels).

If I am understanding it correctly, basically any platform that doesn't implement Channels API, is useless for interop purposes because they can send stuff but not receive.

nicholasdgoodman: I think this line in particular needs clarification or rewording, mostly because the phrase "when the app is joined" itself is ambiguous. This can be interpreted as:

- Context listeners will not receive a context updated event **at the moment** the channel is joined
- Context listeners will not receive context updated events **while the channel is joined**

nkolba:
@vmehtafds @nicholasdgoodman @rikoe   I've made a new PR [here](https://github.com/finos/FDC3/pull/171)  that takes a different approach to resolving these issue.  Please you please review and add your feedback?  

thanks!

PR #169, opened 2020-02-09
Split out API files and docs better

rikoe:
- Rename api-intro and api-spec files
- Move api docs files into "ref" subfolder for better URLs
- Split out types from single Interface.ts file into multiple ts files
- Add fdc3 global constant that implements DesktopAgent interface
- Improve structuring of reference docs in tree
- Doc fixes/improvements

image

image


src/api/Channel.ts
```diff
@@ -0,0 +1,67 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright 2019 FINOS FDC3 contributors - see NOTICE file
+ */
+
+/**
+ * Object representing a context channel.
+ */
+declare class Channel {
+    /**
+     * Constant that uniquely identifies this channel. Will be generated by the service, and guarenteed to be unique
```
nkolba: In the case of App Channels, isn't the id set by the caller of `getOrCreateChannel` ?

rikoe: This is a good catch. I didn't change this file, I just moved the file. This is one reason why I wanted to split out the code, so these type of issues can be spotted more easily. What should the comment be @nkolba then I can change it?

nkolba: Could just be the first sentence: “ Constant that uniquely identifies this channel.”

PR #170, opened 2020-02-09
Rename and publish context schema files

rikoe:
- Change schema file extensions
- Change schema `$id` property to use correct target URI (without version, but "next" instead)
- Change "enum" for "type" to be "const", which is the more correct JSON schema usage
- Some changes to some types to bring in line with FinObj types
- Copy to `website/static/schemas/next` folder as part of build
- Update links in context schema spec to point to published schemas


src/context/schemas/organization.schema.json
```diff
@@ -1,18 +1,18 @@
...
   "properties": {
-    "type": { "enum": ["fdc3.organization"] },
+    "type": { "const": "fdc3.organization" },
     "id": {
       "type": "object",
       "properties": {
         "LEI": { "type": "string" },
         "PERMID": { "type": "string" },
-        "FDSID": { "type": "string" }
+        "FDS_ID": { "type": "string" }
```
rikoe: Changed to `FDS_ID` with an underscore to bring in line with https://fo.finos.org/docs/objects/organization.

src/context/schemas/organization.schema.json
```diff
@@ -1,18 +1,18 @@
...
       }
     }
   },
-  "required": ["id","name"]
+  "required": ["id"]
```
rikoe: I thought we decided `name` was required for organization, but according to https://fo.finos.org/docs/objects/organization it is optional?

src/context/schemas/position.schema.json
```diff
@@ -1,12 +1,12 @@
...
   "type": "object",
   "allOf": [{ "$ref": "context.schema#/" }],
   "properties": {
-    "type": { "enum": ["fdc3.position"] },
+    "type": { "const": "fdc3.position" },
     "instrument": { "$ref": "instrument.schema#/" },
     "holding": { "type": "number", "minimum": 0 }
   },
-  "required": ["instrument", "holding"]
+  "required": ["instrument"]
```
rikoe: It doesn't make sense to me that the `holding` property on position is optional, which is why it was previously required, but according to https://fo.finos.org/docs/objects/position it is optional.

src/context/schemas/organization.schema.json
nkolba: it seems more consistent for just id to be required

src/context/schemas/position.schema.json
nkolba: agreed.  the object is not meaningful without a holding value.

src/context/schemas/contact.schema.json
```diff
@@ -0,0 +1,17 @@
...
+  "$id": "https://fdc3.finos.org/schemas/next/contact.schema.json",
+  "type": "object",
+  "allOf": [{ "$ref": "context.schema.json#/" }],
+  "properties": {
+    "type": { "const": "fdc3.contact" },
+    "id": {
+      "type": "object",
+      "properties": {
+        "email": { "type": "string" },
+        "FDS_ID": { "type": "string" }
```
nkolba: FDS_ID is something I'd associate with a 'person' object, rather than a contact object.  A 
'contact' is someone in my CMS, email, chat, etc.  A 'person' is a reference data object.

rikoe: Should I remove it again @nkolba (given that it is part of the FO definition)?

src/context/schemas/position.schema.json
rikoe: I will change it back.

src/context/schemas/contact.schema.json
nkolba: I’d be in favor of removing then working with FO to clarify on their side

donbasuno: I don't think we should introduce a new "person" object and thus create a separation between that and "contact". We are referencing an individual and there is only one of you and not different parts. It's not like "instrument" and "organization" where you need to reference specific parts for eg trading. I would of course be in favor to keep it in and line up with FO.

nkolba: @donbasuno -I'm fine with leaving as is.

PR #176, opened 2020-02-26
Mark chart iq/context metadata

markChartIQ:
The Context schema should indicate that the ‘id’ field is intended specifically for system or database identification fields and this pattern should be followed when extending context objects.

Updated the 'id' object in the Context schema:
- added comment around 'id' field purpose
- moved 'additionalProperties' field up one level
- added a data object
-- for clients to add non-identifier metadata. Context body.


markChartIQ:
@rikoe I was not able to add a reviewer to this.

markChartIQ:
@rikoe I am fine with this being scheduled for a later release as it is not urgent.

src/context/schemas/context.schema.json
```diff
@@ -4,11 +4,13 @@
...
   "properties": {
     "type": { "type": "string" },
-    "name": { "type": "string" },
+    "name": { "type": "string" }, 
     "id": {
+      "$comment": "Reserved for identifiers",
       "type": "object",
-      "additionalProperties": { "type": "string" }
-    }
+      "additionalProperties": { "type": "any" }
```
rikoe, line 11: I don't think we can or should change the identifiers list be of type any, the idea has always been that they are strings.

src/context/schemas/context.schema.json
```diff
@@ -4,11 +4,13 @@
...
+    },
+    "additionalProperties": { "type": "any" }
```
rikoe, line 13: I think this is good to add, in addition should we add a `$comment` like above to indicate that other data goes here?

kriswest:
@rikoe We've discussed this and think you can close the PR and issue (we can't).

Issue #177, opened 2020-02-26
Generic List Context Object Schema - 2.0


markChartIQ:
#178  Context Object Request

### Description of Problem:
List objects are currently created as a 1-off for each schema (ex. ContactList). As more context objects are added to FDC3, a generic list object will allow those implementing to have greater flexibility in messaging. 

### Potential Solutions:
A suggested generic list object has been created for review

Mixed lists or collections of various objects are not supported in this object.

chedzoir:
As was discussed earlier, a simple implementation of a list could be something like

```
export interface ContextList {
  
  // Set to Context List
  type: string;

  // type of objects that the list contains
  subType: string;

  // the entries in the list, all of which should be context items of the type defined in listOf
  entries: Context[];
  
}
```

An example could be

```
{
  "type":"fdc3.contextList",
  "subType": "fdc3.instrument",
  "entries":[
    {
      "type":"fdc3.instrument",
      "id":{
        "ticker":"GBPUSD"
      }
    },
    {
      "type":"fdc3.instrument",
      "id":{
        "ticker":"EURUSD"
      }
    }
  ]
}
```

Then when it comes to listening for a list of contexts, there's a couple of ways that it could be done

One would be to register intents and listeners with a context string of `fdc3.contextList<fdc3.instrument>` 

Alternatively, you could instead of a string use an object of the form

```
{
  "type":"fdc3.contextList",
  "subType": "fdc3.instrument"
}
```

The former is more elegant, but would need a little parsing on desktop agent start up / intent registration to make calls such as `raiseIntent(intent: string, context: Context, target?: string): Promise<IntentResolution>;` efficient. When those calls are made, we need to identify that the context is a context list and has a sub type that we'd need to look up. But if `Context` is more formally defined than it is now and if we change the signature to `raiseIntent(intent: string, context: Context|ContextList, target?: string): Promise<IntentResolution>;` then it becomes workable.

kriswest:
@chedzoir thanks for your contribution, what you propose is cleaner than whats in the PR associated with this issue.

[...]

PR #178, opened 2020-02-26
Mark chart iq/context type list

markChartIQ:
#177


markChartIQ:
@rikoe This was not intended to be proposed for 1.1. I created an issue #177 and hope to have it discussed as part the next release.

kriswest:
After discussion during the working group, an alternative format for the contextList object was proposed (which replaces the `list.contextType` field with a top-level `subType` field. However, after some discussion, there doesn't seem to be strong support for this proposal (including only half-hearted support at Cosaic) - nor significant objections to continuing to construct list types such as instrumentList or portfolio as named types in their own right.

Hence, we are of the opinion that this PR can be closed.

rikoe:
Thanks @kriswest. Closing.

PR #188, opened 2020-03-12
initial additions to ref data for intents

nkolba:


docs/intents/ref/StartCall.md
```diff
@@ -6,4 +6,40 @@ hide_title: true
...
+```js
+const contact = {
+    type: 'fdc3.contact',
+    name: 'Jane Doe',
+    id: {
+        email: 'jane@mail.com'
+    }
+}
+
+fdc3.raiseIntent('fdc3.StartCall', contact)
```
rikoe: I don't think these should be prefixed with `fdc3`. The decision as per the intents spec was that standardised intents belonging to fdc3 would not have a prefix.

docs/intents/ref/StartChat.md
```diff
@@ -32,7 +32,7 @@ const contact = {
     }
 }
 
-fdc3.raiseIntent('StartChat', contact)
+fdc3.raiseIntent('fdc3.StartChat', contact)
```
rikoe: Same here.

docs/intents/ref/StartCall.md
nkolba: agreed.  updated.

Issue #210, opened 2020-06-18
Error in the documentation of raiseIntent (example contains an error)


kriswest:
### Area of Issue
[ ] App Directory
[X] API
[ ] Context Data
[ ] Intents
[ ] Use Cases
[ ] Other

### Issue Description: 
The documentation for rasieIntent is incorrect: 
- in the example it references a non-existent function (findIntents - plural). 
- findIntent (if that was the intended example) returns an AppIntent - example seems to be based on an IntentResolution
- AppIntent contains multiple apps
- The intent should be raised with the original context passed to findIntent (rather than the undefined `newContext`)

PR #211, opened 2020-06-18
:ambulance: :books: Correcting raiseIntent examples

kriswest:
Addresses issue #210

### DIFF
--- a/src/api/DesktopAgent.ts
+++ b/src/api/DesktopAgent.ts
@@ -102,10 +102,10 @@ interface DesktopAgent {
     /**
      * Raises an intent to the desktop agent to resolve.
      * ```javascript
-     * //raise an intent to start a chat with a given contact
-     * const intentR = await agent.findIntents("StartChat", context);
-     * //use the IntentResolution object to target the same chat app with a new context
-     * agent.raiseIntent("StartChat", newContext, intentR.source);
+     * //Find apps to resolve an intent to start a chat with a given contact
+     * const appIntent = await fdc3.findIntent("StartChat", context);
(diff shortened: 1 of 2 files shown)


src/api/DesktopAgent.ts
```diff
@@ -102,10 +102,10 @@ interface DesktopAgent {
...
-     * const intentR = await agent.findIntents("StartChat", context);
-     * //use the IntentResolution object to target the same chat app with a new context
-     * agent.raiseIntent("StartChat", newContext, intentR.source);
+     * //Find apps to resolve an intent to start a chat with a given contact
+     * const appIntent = await fdc3.findIntent("StartChat", context);
+     * //use the returned AppIntent object to target one of the returned chat apps with the context
+     * await fdc3.raiseIntent("StartChat", context, appIntent.apps[0].name);
      * ```
      */
     raiseIntent(intent: string, context: Context, target?: string): Promise<IntentResolution>;
```
rikoe, line 111: @kriswest Thanks for this! I think `Promise<IntentResolution>` should actually change too right?

kriswest: @rikoe yes you are correct, will update PR

kriswest: @rikoe I thought so at first, but no I think thats correct - rather the example was written as if `fdc3.findIntents()` would return an IntentResolution, when `fdc3.findIntent()` should return AppIntent.

rikoe: Oh yes, I was getting confused with `findIntent` and `findIntentsByContext`, but this is `raiseIntent`...

PR #254, opened 2020-10-09
chore(build): add npm package and docusaurus website publish

rikoe:
- Use docusaurus.yml workflow from ODP
- Only run npm package build for non-website changes
- Add npm publish workflow triggered by GitHub release
- Add "endOfLine": "auto" to prettier settings, so build works on Windows
- Set npm package version to 1.1.0-beta


.github/workflows/docusaurus.yml
```diff
@@ -0,0 +1,83 @@
...
+        # Configure git client
+        echo "Git repo - $REPO"
+        echo "Git user - $GIT_USERNAME"
+        echo "Event name - ${{github.event_name }}"
+        git config --global user.email "$GIT_USERNAME@users.noreply.github.com"
+        git config --global user.name "$GIT_USERNAME"
+        echo "machine github.com login $GIT_USERNAME password $GITHUB_TOKEN" > ~/.netrc
+        # Run docusaurus-publish
+        GIT_USER=$GIT_USERNAME yarn --cwd website publish-gh-pages
+    - name: Create comment
```
maoo, line 74: @rikoe - please remove the `Create comment` step, as it fails (sometimes) probably due to permission issues; I couldn't find a solution yet, but adding a comment to the PR is something that the PR submitter can do by adding content to the description.

.github/workflows/package.yml
```diff
@@ -0,0 +1,51 @@
...
+# Only trigger on:
+# - master branch
+# - PR or Pull Request event types
+# - Exclide Docusaurus files: this file, docs/** and website/**
+on:
+  push:
+    branches:
+      - master
+    paths-ignore:
+      - '.github/workflows/docusaurus.yml'
```
maoo, line 12: Here there should probably be the following paths:
```
- src
- test
- package.json
- tsconfig.json
- yarn.lock
```

Same in the block below.

.github/workflows/release.yml
```diff
@@ -0,0 +1,42 @@
...
+      uses: actions/checkout@v2
+    - name: Configure Node
+      uses: actions/setup-node@v1
+      with:
+        node-version: '12'
+        registry-url: 'https://registry.npmjs.org'
+    - name: Install dependencies
+      uses: bahmutov/npm-install@v1
+    - name: Publish
+      run: yarn publish --access public
```
maoo, line 22: I'd suggest to run tests first, before doing the release, and same below. You could create another job called `test` and the 2 existing jobs would define `needs: test` to set their dependency to it.

.github/workflows/package.yml
rikoe: @maoo I did consider this, but I think it's not ideal. It's easy to think of changes to other files than the ones you mentioned above, that could impact the build (or adding new files at the root level that could, e.g. an eslint file).

README.md goes into the npm package, for example. Similarly changes to the license file, or editor config. And if the build files themselves change.

So I went with the "exclude docusaurus files/folders and include everything else" approach.

maoo (approved):
LGTM!

PR #267, opened 2020-10-19
Allow joining multiple channels

sgd2z:
PR for Issue https://github.com/finos/FDC3/issues/242

### DIFF
--- a/docs/api/spec.md
+++ b/docs/api/spec.md
@@ -166,9 +166,9 @@ There are two types of channels, which are functionally identical, but have diff
  
 
 ### Joining Channels
-Apps can join channels.  An app can only be joined to one channel at a time.  When an app joins a channel it will automatically recieve the current context for that channel.
+Apps can join channels. When an app joins a channel, it will automatically receive the current context for that channel.
 
-When an app is joined to a channel, calls to fdc3.broadcast and listeners added through fdc3.addContextListener will be routed to that channel.  If an app is not joined to a channel these methods will be no-ops, but apps can still choose to listen and broadcast to specific channels via the methods on the `Channel` class.  
+When an app is joined to a channel, calls to fdc3.broadcast and listeners added through fdc3.addContextListener will be routed to that channel.  If an app is not joined to any channels these methods will be no-ops, but apps can still choose to listen and broadcast to specific channels via the methods on the `Channel` class.  
 
 It is possible that a call to join a channel could be rejected.  If for example, the desktop agent wanted to implement controls around what data apps can access.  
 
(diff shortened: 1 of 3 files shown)


src/api/DesktopAgent.ts
```diff
@@ -165,4 +171,12 @@ export interface DesktopAgent {
...
    */
   getCurrentChannel(): Promise<Channel>;
+
+  /**
+   * Returns an Array of `Channel` objects for the current channel membership.
+   *
+   * Returns `null` if the app is not joined to any channels.
+   */
+  getCurrentChannels(): Promise<Array<Channel>>;
+
```
sgd2z, line 181:
```suggestion
```

docs/api/spec.md
```diff
@@ -192,11 +192,11 @@ const context = await globalChannel.getCurrentContext("fdc3.instrument");
...
 ```js
-// check for current fdc3 channel
-let joinedChannel = await fdc3.getCurrentChannel()
-//current channel is null, as the app is not currently joined to a channel
+// check for current fdc3 channels
+let joinedChannels = await fdc3.getCurrentChannels()
+// joinedChannels is null, as the app is not currently joined to any channels
 
-const globalChannel = await fdc3.getSystemChannels.filter(c => c.id === "global")
+const globalChannel = (await fdc3.getSystemChannels).filter(c => c.id === "global")
```
rikoe, line 199: Good spot 👍

docs/api/spec.md
```diff
@@ -227,6 +227,16 @@ fdc3.joinChannel(redChannel.id);
 
 Calling _fdc3.broadcast_ will now route context to the joined channel.
 
+#### Leaving channels
+
+To leave a channel. one calls
```
rikoe, line 232: "To a leave a channel, one calls:" (change period to comma, and add a colon)

docs/api/spec.md
```diff
@@ -227,6 +227,16 @@ fdc3.joinChannel(redChannel.id);
...
+
+```js
+fdc3.leaveChannel(redChannel.id);
+```
+
+Calling _fdc3.broadcast_ will now route context to the joined channel.
```
rikoe, line 238: "to all of the joined channels"?

src/api/DesktopAgent.ts
```diff
@@ -165,4 +171,11 @@ export interface DesktopAgent {
    * Returns `null` if the app is not joined to a channel.
    */
   getCurrentChannel(): Promise<Channel>;
+
+  /**
+   * Returns an Array of `Channel` objects for the current channel membership.
+   *
+   * Returns `null` if the app is not joined to any channels.
```
rikoe, line 178: As per @mattjamieson's comment, let's change this to an empty array.

rikoe (changes requested):

src/api/DesktopAgent.ts
sgd2z: I changed the output of getCurrentChannels (plural) to an empty array. Leaving getCurrentChannel as it is

nkolba:
@sgd2z  how would this impact *fdc3.addContextListener* ?   Would an app listening for context and joined to multiple channels be able to tell which context came from what channel?

kriswest:
@nkolba We've had multiple channels in Finsemble for several years now, clients have rarely if ever wanted to know which channel something came from, however they do sometimes wish to know the component that originated the message (which this change would not affect).

mattjamieson (changes requested):
Approach looks good to me. We should also add/update the DesktopAgent methods in `src/api/methods.ts` for the ES6 API.

Issue #213, opened 2020-06-24
Context Equivalent of RaiseIntent


sgd2z:
### Use Case:
User selects intent as opposed to intent being required in the API call. 

For example: 
User clicks on a contact. With a contact, the user can have multiple possible Intents. So the user can select one of those intents - "StartChat", "Email", "StartCall, "AddToAddressBook", "OpenInCRM" etc.

### Workflow Description
User clicks on a UI representation of a context
System shows UI with possible Intents/Resolvers
User selects an Intent then raiseIntent is automatically called with the intent and context

### Workflow Examples
User clicks on a contact
System shows a UI to select possible intents and Applications that can resolve these intents.
Caller gets back response data which includes the selected Intent & IntentResolution

### Additional Information
This is currently difficult because it requires two API calls:
- findIntentsByContext 
- raiseIntent 
As an app developer it is not immediately obvious how to combine these APIs in a convenient way. findIntentsByContext does result in a list of intents and then the developer could potentially call that, get the response and then show a UI and raiseIntent from that UI. It would be much more convenient to have that dealt with under the hood so it becomes the responsibility of the desktop integrator as opposed to the app developer.

### Potential solutions
Modify raiseIntent to work without an intent specified:
e.g. 
```
raiseIntent(null, context): Promise<IntentResolution>;
```

Add a new API:
```
raiseContext(context): Promise<IntentResolution>;
```

rikoe:
@sgd2z passing null for the first parameter to skip the second feels like an anti-pattern. What I would propose is the following:

```ts
  // two find intent methods
  findIntent(intent: string, context?: Context): Promise<AppIntent>;
  findIntentsByContext(context: Context): Promise<Array<AppIntent>>;

  // two symmetrical raise intent methods
  raiseIntent(intent: string, context: Context, target?: string): Promise<IntentResolution>;
  raiseIntentForContext(context: Context): Promise<IntentResolution>
```

Would `raiseIntentForContext` need an optional `target` parameter as well? I am tempted to say, based on intended usage, no.

Anyone else like to opine?

lspiro-Tick42:
This would be the first call (I think) which requires an FDC3 API implementation to show a UI. This doesn't seem like a good idea. 
Could we just include a simple JavaScript sample about how to do it, using the findIntent and raiseIntent methods ?

rikoe:
[...]

PR #268, opened 2020-10-19
Context Equivalent of RaiseIntent - Issue 213

sgd2z:
PR for Issue https://github.com/finos/FDC3/issues/213

### DIFF
--- a/docs/api/ref/DesktopAgent.md
+++ b/docs/api/ref/DesktopAgent.md
@@ -20,7 +20,7 @@ interface DesktopAgent {
   findIntent(intent: string, context?: Context): Promise<AppIntent>;
   findIntentsByContext(context: Context): Promise<Array<AppIntent>>;
   raiseIntent(intent: string, context: Context, target?: string): Promise<IntentResolution>;
-  raiseContext(context: Context): Promise<IntentResolution>;
+  raiseIntentForContext(context: Context, target?: string | AppMetadata): Promise<IntentResolution>;
   addIntentListener(intent: string, handler: ContextHandler): Listener;
   
   // channels
@@ -342,17 +342,17 @@ const appIntent = await fdc3.findIntent("StartChat", context);
 await fdc3.raiseIntent("StartChat", context, appIntent.apps[0].name);
 ```
 
-### `raiseContext`
+### `raiseIntentForContext`
 
 ```ts
-raiseContext(context: Context): Promise<IntentResolution>;
+raiseIntentForContext(context: Context, target?: string | AppMetadata): Promise<IntentResolution>;
 ```
-Raises a context to the desktop agent to resolve. Raise context deals with the case where a context can have multiple associated intents. Similar to raiseIntent without a target, it provides the opportunity for Intent and target selection, which can result in a call to raiseIntent with a target under the hood to provide Intent resolution.
+Allows for selection of an intent for a specific context for the desktop agent to resolve. Raising intent for a context deals with the case where a context can have multiple associated intents. Similar to raiseIntent without a target, it provides the opportunity for Intent and target selection, which can result in a call to raiseIntent or raiseIntentForContext with a target under the hood to provide Intent resolution.
 
 #### Example
 
 ```js
-const intentResolution = await fdc3.raiseContext(context);
+const intentResolution = await fdc3.raiseIntentForContext(context);
 ```
 
 
--- a/docs/api/ref/DesktopAgent.md
+++ b/docs/api/ref/DesktopAgent.md
@@ -20,7 +20,7 @@ interface DesktopAgent {
   findIntent(intent: string, context?: Context): Promise<AppIntent>;
   findIntentsByContext(context: Context): Promise<Array<AppIntent>>;
   raiseIntent(intent: string, context: Context, target?: string): Promise<IntentResolution>;
-  raiseIntentForContext(context: Context, target?: string | AppMetadata): Promise<IntentResolution>;
+  raiseIntentForContext(context: Context, target?: Target): Promise<IntentResolution>;
   addIntentListener(intent: string, handler: ContextHandler): Listener;
   
   // channels
@@ -345,7 +345,7 @@ await fdc3.raiseIntent("StartChat", context, appIntent.apps[0].name);
 ### `raiseIntentForContext`
 
 ```ts
-raiseIntentForContext(context: Context, target?: string | AppMetadata): Promise<IntentResolution>;
+raiseIntentForContext(context: Context, target?: Target): Promise<IntentResolution>;
 ```
 Allows for selection of an intent for a specific context for the desktop agent to resolve. Raising intent for a context deals with the case where a context can have multiple associated intents. Similar to raiseIntent without a target, it provides the opportunity for Intent and target selection, which can result in a call to raiseIntent or raiseIntentForContext with a target under the hood to provide Intent resolution.
 
@@ -358,6 +358,5 @@ const intentResolution = await fdc3.raiseIntentForContext(context);
 
 #### See also
 * [`IntentResolution`](IntentResolution)
-
 
 
--- a/docs/api/ref/DesktopAgent.md
+++ b/docs/api/ref/DesktopAgent.md
@@ -20,6 +20,7 @@ interface DesktopAgent {
   findIntent(intent: string, context?: Context): Promise<AppIntent>;
   findIntentsByContext(context: Context): Promise<Array<AppIntent>>;
   raiseIntent(intent: string, context: Context, target?: string): Promise<IntentResolution>;
+  raiseContext(context: Context): Promise<IntentResolution>;
   addIntentListener(intent: string, handler: ContextHandler): Listener;
   
   // channels
@@ -340,6 +341,20 @@ const appIntent = await fdc3.findIntent("StartChat", context);
 //use the returned AppIntent object to target one of the returned chat apps with the context
 await fdc3.raiseIntent("StartChat", context, appIntent.apps[0].name);
 ```
+
+### `raiseContext`
+
+```ts
+raiseContext(context: Context): Promise<IntentResolution>;
+```
+Raises a context to the desktop agent to resolve. Raise context deals with the case where a context can have multiple associated intents. Similar to raiseIntent without a target, it provides the opportunity for Intent and target selection, which can result in a call to raiseIntent with a target under the hood to provide Intent resolution.
+
+#### Example
+
+```js
+const intentResolution = await fdc3.raiseContext(context);
+```
+
 
 #### See also
 * [`IntentResolution`](IntentResolution)
--- a/docs/api/spec.md
+++ b/docs/api/spec.md
@@ -115,7 +115,7 @@ Raising an Intent will return a Promise-type object that will resolve/reject bas
 - The intent was ambiguous and the resolver experienced an error.
 
 ##### Resolution Object
-If the raising of the context or intent resolves (or rejects), a standard object will be passed into the resolver function with the following format:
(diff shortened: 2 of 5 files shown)


mattjamieson:
Is there any scenario where we would want to `raiseContext` with a `target`? I can see it being potentially useful where the target is known but the Desktop Agent has UI (or other) features for intent resolution which can be used prior to passing the intent+context to the target.

```typescript
raiseContext(context: Context, target?: string);
```

rikoe:
> Is there any scenario where we would want to `raiseContext` with a `target`? I can see it being potentially useful where the target is known but the Desktop Agent has UI (or other) features for intent resolution which can be used prior to passing the intent+context to the target.
> 
> ```ts
> raiseContext(context: Context, target?: string);
> ```

@mattjamieson I agree, it doesn't hurt, and is consistent with `raiseIntent`. Although to be consistent with #272 it should probably be `raiseContext(context: Context, target?: string | AppMetadata)` 😉

mattjamieson:
> [...] to be consistent with #272 it should probably be `raiseContext(context: Context, target?: string | AppMetadata)` 😉

👍, might also be nice to define:
```typescript
type Target = string | AppMetadata;
```

docs/api/ref/DesktopAgent.md
```diff
@@ -20,6 +20,7 @@ interface DesktopAgent {
   findIntent(intent: string, context?: Context): Promise<AppIntent>;
   findIntentsByContext(context: Context): Promise<Array<AppIntent>>;
   raiseIntent(intent: string, context: Context, target?: string): Promise<IntentResolution>;
+  raiseContext(context: Context): Promise<IntentResolution>;
```
rikoe, line 23: I would prefer if we can call this `raiseIntentForContext`. 

There are two reasons for this:
1. You can't raise a "context", that doesn't make sense intuitively.
2. `raiseIntent` and `raiseIntentForContext` would follow the same pattern as `findIntent` and `findIntentsByContext`, one of which takes an intent name, and the other just a context. It's good symmetry.

docs/api/ref/DesktopAgent.md
```diff
@@ -341,6 +342,20 @@ const appIntent = await fdc3.findIntent("StartChat", context);
 await fdc3.raiseIntent("StartChat", context, appIntent.apps[0].name);
 ```
 
+### `raiseContext`
+
+```ts
+raiseContext(context: Context): Promise<IntentResolution>;
+```
+Raises a context to the desktop agent to resolve. Raise context deals with the case where a context can have multiple associated intents. Similar to raiseIntent without a target, it provides the opportunity for Intent and target selection, which can result in a call to raiseIntent with a target under the hood to provide Intent resolution.
```
rikoe, line 350: This description doesn't make sense to me, mostly I think because it is not clear what "raising a context" means if you don't understand that under the covers the desktop is looking up intent(s) for the context and asking the user to pick one.

I think it is good if FDC3 (at this stage) only have one "thing" that can be "raised", but there are different ways of doing so, i.e. based on name, and based on context.

docs/api/spec.md
```diff
@@ -78,29 +78,31 @@ Examples of End Points include:
 ### Open an Application by Name
 Linking from one application to another is a critical basic workflow that the web revolutionized via the hyperlink.  Supporting semantic addressing of applications across different technologies and platform domains greatly reduces friction in linking different applications into a single workflow.
 
-### Raising Intents
+### Raising Intents or Contexts
```
rikoe, line 81: This could be changed back to just `Raising Intents` - there is not an additional concept to understand.

docs/api/spec.md
```diff
@@ -78,29 +78,31 @@ Examples of End Points include:
...
 Often, we want to link from one app to another to dynamically create a workflow.  Enabling this without requiring prior knowledge between apps is a key goal of FDC3.
 
-Intents provide a way for an app to request functionality from another app and defer the discovery and launching of the destination app to the Desktop Agent.  There are multiple models for interop that Intents can support.
+Intents provide a way for an app to request functionality from another app and defer the discovery and launching of the destination app to the Desktop Agent.  There are multiple models for interop that Contexts and Intents can support.
```
rikoe, line 84: Here as well, this paragraph really refers to Intents, not to Contexts (which can't be raised in and of themselves) - just sticking to Intents as the thing that can be raised will make this clearer.

docs/api/spec.md
```diff
@@ -78,29 +78,31 @@ Examples of End Points include:
...
 
 - **Chain**:  In this case the workflow is completely handed off from one app to another (similar to linking).   Currently, this is the primary focus in FDC3
 - **Client-Service**: A Client invokes a Service via the Intent, the Service performs some function, then passes the workflow back to the Client.  Typically, there is a data payload type associated with this intent that is published as the standard contract for the intent.
 - **Remote API**: An app wants to remote an entire API that it owns to another App.  In this case, the API for the App cannot be standardized.  However, the FDC3 API can address how an App connects to another App in order to get access to a proprietary API.
 
+A Context can be associated with multiple intents. For example, an instrument could be associated with ViewChart, ViewNews, ViewAnalysis or some other intent. Raising a context provides the opportunity for the system or the user to select the appropriate Intent for the selected Context and resolve that Intent.
```
rikoe, line 90: "Raising an intent by context provides" .... "for the _provided_ Context" (the context isn't being selected, the intent is).

docs/api/spec.md
```diff
@@ -78,29 +78,31 @@ Examples of End Points include:
...
+
 #### Intent Resolution
-Raising an Intent will return a Promise-type object that will resolve/reject based on a number of factors.
+Raising an Intent or a Context will return a Promise-type object that will resolve/reject based on a number of factors.
```
rikoe, line 93: The title says "Intent Resolution", but it then talks about "Raising an Intent or Context" - removing the concept of raising "contexts" removes this ambiguity.

docs/api/spec.md
```diff
@@ -78,29 +78,31 @@ Examples of End Points include:
...
 
 ##### Resolve
-- Intent was resolved unambigiously and the recieving app was launched successfully.
-- Intent was ambigious, a resolution was chosen by the end user and the chosen application was launched succesfully.
+- Intent was resolved unambiguously and the receiving app was launched successfully.
+- Intent was ambiguous, a resolution was chosen by the end user and the chosen application was launched successfully.
```
rikoe, line 97: I think in this case the intent resolution return type should tell you which intent was ultimately selected by the user, otherwise this information is lost.

docs/api/spec.md
```diff
@@ -125,7 +127,18 @@ try {
...
+```js
+try {
+    const result = await fdc3.raiseContext(context);
+    if (result.data) {
+        const orderId = result.data.id;
+    }
+}
+catch (er){
+    console.log(er.message);
+}
```
rikoe, line 141: I suggest expanding `IntentResolution` with an `intentName` property, which will match the provided name when using `raiseIntent`, but will be populated with the selected `intentName` when resolved by the user or desktop agent based on context alone.

rikoe (changes requested):

sgd2z:
> Is there any scenario where we would want to `raiseContext` with a `target`? I can see it being potentially useful where the target is known but the Desktop Agent has UI (or other) features for intent resolution which can be used prior to passing the intent+context to the target.
> 
> ```ts
> raiseContext(context: Context, target?: string);
> ```

Interesting idea. We should discuss in the meeting. I was thinking that raiseContext would eventually result in a raiseIntent call with a target and only added a raiseContext without a target. For drag and drop I was thinking of private temporary channel broadcast.

docs/api/ref/DesktopAgent.md
sgd2z: That's a good suggestion. Also makes the explaining it simpler. I'll update the PR and docs.

nkolba: maybe 'raiseIntentsByContext' ?

kriswest (changes requested):
WG feedback: Addition to API makes sense, some changes requested for @sgd2z to make: 
- update function name to raiseIntentForContext
- add optional target parameter for filtering purposes
- ensure that docs are clear that IntentResolver UI should ask you to select an intent (which the resolver doesn't do at the moment), then an app to resolve it (unless a target was specified)

docs/api/spec.md
rikoe: > I suggest expanding `IntentResolution` with an `intentName` property, which will match the provided name when using `raiseIntent`, but will be populated with the selected `intentName` when resolved by the user or desktop agent based on context alone.

Just to confirm, will consider this as a separate Issue/PR 👍

sgd2z:
PR has been updated with feedback and based on our discussions

docs/api/spec.md
sgd2z: https://github.com/finos/FDC3/issues/284

docs/api/spec.md
```diff
@@ -100,20 +100,22 @@ Intents provide a way for an app to request functionality from another app and d
...
 ##### Reject
-- An app matching the intent was not found.
-- A match was found, but the recieving app failed to launch.
+- No app matching the intent or any possible intents for the specific context were not found.
+- A match was found, but the receiving app failed to launch.
 - The intent was ambiguous and the resolver experienced an error.
 
 ##### Resolution Object
-If the raising of the intent resolves (or rejects), a standard object will be passed into the resolver function with the following format:
+If the raising of the context or intent resolves (or rejects), a standard object will be passed into the resolver function with the following format:
```
sgd2z, line 118:
```suggestion
If the raising of the intent resolves (or rejects), a standard object will be passed into the resolver function with the following format:
```

src/api/DesktopAgent.ts
```diff
@@ -126,6 +127,14 @@ export interface DesktopAgent {
     target?: Target
   ): Promise<IntentResolution>;
 
+  /**
+   * Raises a context to the desktop agent to resolve with one of the possible Intents for that context.
+   * ```javascript
+   * await fdc3.raiseIntentForContext(context);
+   * ```
+   */
+  raiseIntentForContext(context: Context, target?: string | AppMetadata): Promise<IntentResolution>;
```
sgd2z, line 136:
```suggestion
  raiseIntentForContext(context: Context, target?: Target): Promise<IntentResolution>;
```

docs/api/ref/DesktopAgent.md
```diff
@@ -20,6 +20,7 @@ interface DesktopAgent {
   findIntent(intent: string, context?: Context): Promise<AppIntent>;
   findIntentsByContext(context: Context): Promise<Array<AppIntent>>;
   raiseIntent(intent: string, context: Context, target?: string): Promise<IntentResolution>;
+  raiseIntentForContext(context: Context, target?: string | AppMetadata): Promise<IntentResolution>;
```
sgd2z, line 23:
```suggestion
  raiseIntentForContext(context: Context, target?: Target): Promise<IntentResolution>;
```

docs/api/ref/DesktopAgent.md
```diff
@@ -341,6 +342,20 @@ const appIntent = await fdc3.findIntent("StartChat", context);
 await fdc3.raiseIntent("StartChat", context, appIntent.apps[0].name);
 ```
 
+### `raiseIntentForContext`
+
+```ts
+raiseIntentForContext(context: Context, target?: string | AppMetadata): Promise<IntentResolution>;
```
sgd2z, line 348:
```suggestion
raiseIntentForContext(context: Context, target?: Target): Promise<IntentResolution>;
```

rikoe (approved):

rikoe (approved):

Issue #241, opened 2020-09-15
Ready Event before FDC3 API can be called


sgd2z:
## Enhancement Request
We would like to propose adding a ready event before FDC3 API can be called and a ready state global.

```
function fdc3stuff() {
  // Make some fdc3 API calls here
}

if (window.fdc3Ready) {
  fdc3stuff();
} else {
  window.addEventListener("fdc3Ready", fdc3stuff);
}
```

### Use Case:
Calling the DesktopAgent API generally requires that the Desktop Agent is ready to broadcast/receive data which generally relies on some underlying infrastructure to be ready. 

For example: 
In a Desktop Integration Platform like Finsemble where communication happens between different components, it is necessary for the communication channels of the platform to be ready before fdc3 API can be called. Also, the fdc3 API itself relies on a a service or other infrastructure outside the component, which needs to be ready before it can be used. So calling any fdc3 API as soon as a component starts up might fail.

Just adding a ready event however is not enough because sometimes, the ready event has already occurred before any checks are made. So we would like to propose adding a global ready state that can be checked in addition to the ready event.

rikoe:
I would prefer:

```ts
await window.fdc3.onReady()
```

to adding another top-level global primitive - this is already a smell so let's not make it worse.

In addition, an async function feels more like modern JS as opposed to a window-level event.

sgd2z:
@rikoe while that is simple, in real life use cases does not work. You cannot ensure that window.fdc3 even exists, which is why it must be window event based. Also I'm not sure top level await is even a thing yet so your await must go into an async function anyway.

rikoe:
Fair enough - I think we should discuss this during the WG meeting. I don't understand the implications of this fully. But it would be great if we can find another way to do it. We have a request to port the API to .NET etc., so we don't want to rely on the window object too much. According to the spec, `window.fdc3` is a SHOULD not a MUST.

When we discussed this in the past, people were in favour of an import pattern is opposed to a global object pattern, perhaps we should revisit this.

sgd2z:
Using an import pattern might not work for desktop integrators like us because we preload all the default functionality into the browser as part of the platform and we use a similar pattern for our own ready event. And there is no guarantee that the preload finishes execution before your own code uses it, which is why an event is needed. However to make sure that you didn't miss the event, you need to also check state. I think the same pattern would apply to .NET

if (FDC3.ready) { fn() } else FDC3.addEventListener("ready", fn)

[...]

PR #269, opened 2020-10-19
FDC3 ready event - Issue 241

sgd2z:
PR for Issue https://github.com/finos/FDC3/issues/241

### DIFF
--- a/docs/api/spec.md
+++ b/docs/api/spec.md
@@ -40,7 +40,7 @@ Other interfaces defined in the spec are not critical to define as concrete type
 #### API Access
 The FDC3 API can be made available to an application through a number of different methods.  In the case of web applications, a Desktop Agent SHOULD provide the FDC3 API via a global accessible as _window.fdc3_. Implementors MAY additionally make the API available through modules, imports, or other means. 
 
-The global `window.fdc3` must only be available after the API is ready to use. To prevent the API from being used before it is ready, implementors may provide a global `fdc3Ready` event. Here is code demonstrating the use of the FDC3 API and the ready event:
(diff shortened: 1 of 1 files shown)


docs/api/spec.md
```diff
@@ -37,10 +37,23 @@ Other interfaces defined in the spec are not critical to define as concrete type
...
+  // Make some fdc3 API calls here
+}
+
+if (window.fdc3) {
+  fdc3Stuff();
+} else {
+  window.addEventListener("fdc3Ready", fdc3Stuff);
+}
+```
+
```
rikoe, line 56: Is it enough to just add this to the spec here? What about the website documentation and the typings? (Not sure if it is possible to add typings for a window-level event, just checking...)

docs/api/spec.md
```diff
@@ -37,10 +37,23 @@ Other interfaces defined in the spec are not critical to define as concrete type
 - `DisplayMetadata`
 
 
-
 #### API Access
 The FDC3 API can be made available to an application through a number of different methods.  In the case of web applications, a Desktop Agent SHOULD provide the FDC3 API via a global accessible as _window.fdc3_. Implementors MAY additionally make the API available through modules, imports, or other means. 
 
+The global `window.fdc3` must only be available after the API is ready to use. To prevent the API from being used before it is ready, implementors may provide a global `fdc3Ready` event. Here is code demonstrating the use of the FDC3 API and the ready event:
```
rikoe, line 43: Should this be changed to "implementors SHOULD Provide", to be consistent with the rest of the spec? If we are making this addition to version 1.2, it should probably have the same mandate as `window.fdc3`?

docs/api/spec.md
sgd2z: Can you point me to where the website documentation is?

rikoe: 🤔  doesn't seem there is any!

rikoe (approved):

Issue #240, opened 2020-09-15
Add additional properties to AppMetadata type


sean-ciq:
### Use Case:
An application developer, in a situation where more than one App Directory exists and the names of applications are not unique, needs more information about applications returned from `findIntent()` or `findIntentsByContext()` to successfully raise an intent to a specified `target`. We recommend bringing in fields from the AppD application definition that differentiate different apps with the same name:
- `appId`: the fully qualified appID differentiates based on version and the appDirectory and hence should be globally unique
- `version`: the value that differentiates different copies of an app within a single app directory 

### Workflow Description
1. Application calls `findIntent()` or `findIntentsByContext()`
2. Application receives a response containing a list of AppIntent objects, themselves containing an AppMetadata object for each application that details are being returned on.
3. AppMetadata objects contain values that can *uniquely* identify each app (given that multiple AppD's may be in use, each of which could contain multi[le versions of an application with a single `name` value).

### Workflow Examples
1. User clicks on a contact
2. UI displays the available applications.
3. There are multiple applications available with the name "Messenger"
4. The user selects any "Messenger" from the UI
5. Instead of passing a `name` value, the application passes an AppMetadata object containing an `appId` value or `name` and `version`, uniquely identifying the app
6. Desired app version is launched.

### Additional Information
Note the following definitions in the AppD specification for an application record:

> `name`: The name of the application. The name should be unique within an FDC3 App Directory instance. The exception to the uniqueness constraint is that an App Directory can hold definitions for multiple versions of the same app. The same appName could occur in other directories. We are not currently specifying app name conventions in the document.
> `version`: Version of the application. This allows multiple app versions to be defined using the same app name. This can be a triplet but can also include things like 1.2.5 (BETA)

Integrating additional properties defined in the App D specification provides an additional way for applications to determine and specify intent targets. The argument for returning the entire AppMetadata object from `findIntents()` & `findIntentsByContext()` and others is based upon deference to the application. If an application receives a list of available apps, there is currently no identifier other than the `name` property. This inability to properly identify other applications by one or more properties breaks down when:
- there are multiple app directories
- there are multiple versions of an app within the app directory 

### Proposed Additions

Current `AppMetadata` interface:
```Typescript
interface AppMetadata {
  name: string;
  title?: string;
  tooltip?: string;
  description?: string;
  icons?: Array<string>;
  images?: Array<string>;
}
```

New interface:
```Typescript
interface AppMetadata {
  name: string;
  appId?: string;
  version?: string;
[...]

PR #273, opened 2020-10-21
Add appId & version to AppMetadata

sean-ciq:
This PR contributes appId & version to the definition of an AppMetadata object as described in #240 

Specifically this PR changes
- The AppMetadata type definition and interface
- Updates documentation

### DIFF
--- a/src/api/AppMetadata.ts
+++ b/src/api/AppMetadata.ts
@@ -10,6 +10,12 @@ export interface AppMetadata {
   /** The unique app name that can be used with the open and raiseIntent calls. */
   readonly name: string;
 
+  /** The unique application identifier located within a specific application directory instance.  */
+  readonly appId?: string;
+
+  /** The Version of the application. */
+  readonly version?: string;
+
   /** A more user-friendly application title that can be used to render UI elements  */
   readonly title?: string;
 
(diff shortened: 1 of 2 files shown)


src/api/AppMetadata.ts
```diff
@@ -10,6 +10,12 @@ export interface AppMetadata {
   /** The unique app name that can be used with the open and raiseIntent calls. */
   readonly name: string;
 
+  /** The unique application identifier located within a specific application directory instance.  */
```
kriswest, line 13: Add examples of a fully qualified appId to the comments

kriswest (changes requested):
WG feedback - PR/change consider 'harmless' and should pass. Could we add some example appIds to comments

src/api/AppMetadata.ts
rikoe: Do we have such examples, or would we just be speculating?

rikoe (approved):

src/api/AppMetadata.ts
sean-ciq: @kriswest I decided to go with consistency. I added an example to the comment based upon the documentation here https://fdc3.finos.org/docs/1.1/app-directory/discovery

PR #277, opened 2020-10-28
Reject/throw as appropriate in ES6 exported methods if window.fdc3 is not available

mattjamieson:
Improve error handling in exported methods by:
- rejecting with the message `'FDC3 DesktopAgent not available at 'window.fdc3'.` if the method returns a `Promise` (e.g. `open`), and
- throwing with the same message if the method is synchronous (e.g. `broadcast`).

At present, developers would see the significantly less helpful:
> `Uncaught TypeError: Cannot read property 'broadcast' of undefined`

This can be improved further in future by leveraging the ready event pattern proposed in #269.

### DIFF
--- a/src/api/methods.ts
+++ b/src/api/methods.ts
@@ -7,28 +7,43 @@ import {
   Listener,
 } from '..';
 
+const unavailableError = new Error(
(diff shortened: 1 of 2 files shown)


rikoe (approved):
👏

PR #282, opened 2020-11-13
Support nullable return type for getCurrentChannel

garyadaptive:
According to the specification (https://fdc3.finos.org/docs/1.1/api/ref/DesktopAgent#getcurrentchannel), `getCurrentChannel` should allow for a nullable return type, but the current type signature does not support this.

Included changes:
1. In references to `getCurrentChannel`, change the return type to `Promise<Channel | null>`


rikoe:
@cla-bot[bot] check

rikoe (approved):

PR #285, opened 2020-11-18
Clarify docs for broadcast functions

kriswest:
While working with a vendor app, we noted that they were worried about message loops. The documentation should clarify that its the job of the desktop agent implemention to prevent an application from receiving its own broadcasts


kriswest:
@rikoe minor docs addition. Any idea why the linter doesn't like it? Error message didn't make a whole lot of sense to me.

rikoe:
> @rikoe minor docs addition. Any idea why the linter doesn't like it? Error message didn't make a whole lot of sense to me.

@kriswest I have recently noticed on another project that prettier is very aggressive about markdown files, and it's not possible to change the rules.

Can you please add a `.prettierignore` file to the root, with *.md as an exclusion? Should make the problem go away.

kriswest:
@rikoe markdown issues resolved. Had to cleanup some trailing spaces in other docs that were also getting rejected by prettier.

rikoe (approved):

Issue #231, opened 2020-07-28
Multiple instances of the same application 


pgn-vole:
### Question Area
[ ] App Directory
[X ] API
[ ] Context Data
[ ] Intents
[ ] Use Cases
[ ] Other

### Question

One of the basic use case I'm looking to enable is to allow user to open multiple instances of a single application, for instance multiples Chart or Tickets. Looking the current specification it seems that this use case is not covered, for example the `raiseIntent` API `target` parameter accepts an application name and not of application instance uuid (or something similar).
Have you put some thought already into that use case?

lspiro-Tick42:
Managing application, application instances, discovery and also dynamically publishing new Intents (so applications can register them outside of the App Directory) are important parts of the Interop approach in Glue42. We have worked with DB in the Plexus project to produce an API that allows for this functionality. The API is defined [here](https://github.com/finos/plexus-interop-desktop-api/blob/master/src/client-api.ts)

I have a task to review this API and suggest what bits (if any) might be useful to bring into FDC3 API.

My opinion is that a proper Request/Response API is much better than trying to add return values to Intents, but that is a longer discussion

nkolba:
@pgn-vole 
Handling app instances is a Desktop Agent implementation concern (not appD) as far as the standards go.  Different platforms and technologies have different notions of what an application instance is and how to reference one.  You can see an open source FDC3 implementation that handles this [here](https://github.com/finos/fdc3-desktop-agent).  Since this implementation is a ChromeExtension - app instances are just separate tab instances (in Chrome) that happen to map to the same appD definition.

The resolution of multiple instances in the intent handler looks like this:
![image](https://user-images.githubusercontent.com/1150874/88973131-a68d1000-d284-11ea-86fd-d306ce7eab04.png)

lspiro-Tick42:
Nick, I don't think the interesting question is about Desktop Agent vs AppD, I think the question is the extent to which applications writing to FDC3 want to have 'control' over application instances.

For example if an application calls the FDC3 Open method, in our API's (and we will be proposing extensions to add this for FDC3) the application caller should receive a handle to an application instance, which they can use to close the instance. We also believe it should be possible for applications to use that instance handle to invoke intents on that instance.

We will further propose that the instance handle can be used to inquire what methods are supported by that application and invoke those methods using the Plexus Interop API, but that will be the subject of a separate issue

pgn-vole:
@nkolba thanks for sharing this example implementation. 
[...]

Issue #201, opened 2020-05-26
Data Driven Intents


nkolba:
## Enhancement Request

Intents in FDC3 should be able to return data payloads.

### Use Cases:
Some use cases for data resolving intents include:

#### Symbol Resolution
As an end user, I may want to initiate an FDC3 workflow from an app providing news or other unstructured data.  For example, "International Business Machines" appearing in a body of text.  Resolving the string to structured instrument data - as well as disambiguating the instrument the end user may want to work with - is critical to routing the end user to the right workflow.  

#### Price Quotes
As an end user, I will have content that is not directly hooked to a  pricing system.  Enabling inline pricing in any app and with the pricing system of my choice would provide great value in alleviating the need for me to context switch between multiple apps.

#### Placing Orders
As an end user, I want to leverage the OMS of my choice, and to be able to slot in a new system - without relying on backend integrations between vendors or being forced to rekey to fill last mile gaps.  An end user should be able to receive an inquiry in a chat, place an order with the OMS of choice, and send a confirmation # back to the requester without leaving the chat application.

### Workflow Description

- An intent is raised as it is today.
- This intent is resolved against known applications providing the intent.
- If there is ambiguity, the end user chooses which app they wish to resolve the intent.
- The app is launched (if needed) and is passed the intent and any accompanying context
- The resolved app returns the requested data and this is passed back to the calling app as the resolution of the intent.

### Workflow Examples
For pricing:
A grid provider has enabled realtime pricing for their watch list app.  They add a button to their app to "get prices".  When selected, this will raise a 'getPrice' event for the instrumentList.  The end user will be prompted to choose the pricing service they want to use.  The prices will be returned (and the intent resolved) and the grid will update its UI. 

### Additional Information

Related issue https://github.com/finos/FDC3/issues/127

Some previously floated ideas around implementation:
- Returned data would be treated as Context Data and standardized (for standard intents)
- The intent definition for data intents would be expanded to specify the context data type returned

To consider:
- Streaming data

nkolba:
Questions:
Do we need to distinguish between background (no UI) and visual (with UI) intent resolvers?

How do we handle delayed/failed return of data?  

How do we manage return data structures for different asset classes?

kriswest:
The name for this issue doesn't match the proposal well, which describes 'IntentResolutions with data' or 'Resolving Intents with Data' (rather than intents that were driven by data, although I concede they might have been driven by a desire to retrieve data).

rikoe:
[...]

PR #279, opened 2020-11-11
Adds a new Target type, updates open() and raiseIntent()

sean-ciq:
This PR adds a new Target type definition to src/api. It also adjusts the `target` argument currently in `open()` and `raiseIntent()` to accept that new type.

Changes include:

1. Adding TargetType.ts to `src/api`
2. `open()` and `raiseIntent()` now accept `target: Target` instead of `target: string | AppMetadata`
3. Changes `IntentResolution.souce` to use `Target` type

PR #287, opened 2020-11-22
App instance proposal + connect to app instance for intent results

nkolba:
This is an initial PR to introduce the concept of AppInstances to address data returning intents - as outlined in issue #201 as well as questions around app instances (e.g. issue #231).  

Key points of this PR are:

-   introduction of an *AppInstance* class that provides a reference to a specific instance of an app and allows direct broadcasting / listening of context with that instance
- `IntentResolution` *source* properties can be used to obtain an AppInstance
- `ContextHandler` functions now get an additional *source* arg that provides a reference to the instance sending the context or intent
- Instances provide lifecycle information

### DIFF
--- a/docs/api/ref/AppInstance.md
+++ b/docs/api/ref/AppInstance.md
@@ -1,0 +1,177 @@
+---
+id: AppInstance
+sidebar_label: AppInstance
+title: AppInstance
+hide_title: true
+---
+# `AppInstance`
+
+```ts
+interface AppInstance {
+  //properties
+  readonly instanceId : string;
+  readonly status : 'ready' | 'loading' | 'unregistered';
+  
+  //methods
+  addContextListener(handler: ContextHandler): Listener;
+  addContextListener(contextType: string, handler: ContextHandler): Listener;
+  broadcast(context: Context): void;
+  onStatusChanged(handler : (newVal : string, oldVal :string) => {}) : void;
+
+}
+```
+
+Represents a specific instance of an app.  
+
+AppInstances can be used to directly listen for and broadcast context between two apps connected to a desktop agent.
+
+An AppInstance is obtained by calling the `getAppInstence` method on the `DesktopAgent` using a source token provided by either an IntentResolution or as an argument to a ContextHandler. 
+
+#### See also
+
+* [`IntentResolution`](IntentResolution)
+* [`ContextHandler`](ContextHandler)
+* [`DesktopAgent.getAppInstance`](DesktopAgent#getappinstannce)
+
+
+## Properties
+
+### `instanceId`
+
+```ts
+public readonly instanceId: string;
+```
+
+Uniquely identifies the instance. This identifier is internal to the desktop agent.  It should correspond to the source prop returned for IntentResolution and ContextHandlers.
+
+### `status`
+
+```ts
+public readonly status: 'ready' | 'loading' | 'unregistered';
+```
+
+Represents the lifecycle status of the AppInstance.  
+
+- *ready* indicates that the instance is loaded and connected to the DesktopAgent.
+- *loading* indicates that the instance is in the process of loading and not yet connected and ready to recieve context or intents.
+- *unregistered* indicates that the instance has disconnected from the DesktopAgent and is no longer available. 
+
+
+## Methods
+
+
+### `addContextListener`
+
+```ts
+public addContextListener(handler: ContextHandler): Listener;
+```
+
+Adds a listener for incoming contexts from the app instance.
+
+```ts
+public addContextListener(contextType: string, handler: ContextHandler): Listener;
+```
+
+Adds a listener for incoming contexts from the app instance using the specified _context type_.
+
+#### Examples
+Raise an intent then add a context listener on the resolved instance.
+```ts
+const listener = instance.addContextListener(context => {
+    if (context.type === 'fdc3.contact') {
+        // handle the contact
+    } else if (context.type === 'fdc3.instrument') => {
+        // handle the instrument
+    }
+});
+
+
+```
+
+Adding listeners for specific types of context that is broadcast from the instance:
+
+```ts
+const contactListener = instance.addContextListener('fdc3.contact', contact => {
+    // handle the contact
+});
+
+const instrumentListener = instance.addContextListener('fdc3.instrument', instrument => {
+    // handle the instrument
+});
+
+// later
+contactListener.unsubscribe();
+instrumentListener.unsubscribe();
+```
+
+#### See also
+* [`Listener`](Listener)
+* [`ContextHandler`](ContextHandler)
+* [`IntentResolution`](IntentResolution)
+
+
+### `broadcast`
+
+```typescript
+public broadcast(context: Context): void;
+```
+
+Broadcasts a context directly to the app instance. 
+
+If the instance is not available, the method will return an `Error` with a string from the [`InstanceError`](InstanceError) enumeration.
+
+#### Example
+
+```javascript
+const instrument = {
+    type: 'fdc3.instrument',
+    id: {
+        ticker: 'AAPL'
+    }
+};
+
+try {
+    instance.broadcast(instrument);
+} catch (err: InstanceError) {
+    // handler errror
+}
+```
+
+#### See also
+* [`InstanceError`](ChannelError)
+* [`addContextListener`](#addcontextlistener)
+
+### `onStatusChannged`
+
+``` typescript
+public onStatusChanged(handler : (newVal : string, oldVal :string) => {}) : void;
+```
+
+Sets a listener for status change events on an app instance.
+
+Possible status values are:
+
+- ready
+- loading
+- unregistered
+
+### Example
+
+```javascript
+
+// get an app instance
+const instance = await fdc3.getAppInstance(source);
+
+// if status is ready, add a context listener
+if (instance.status === 'ready'){
+    const listener = instance.addContextListener(someContext, myListener);
+}
+
+// listen for status change, if status is unregistered, then unsubscribe the context listener
+instance.onStatusChanged((status) => {
+    if (status === 'unregistered'){
+        listener.unsubscribe();
+    }
+});
+
+```
--- a/docs/api/ref/AppInstance.md
+++ b/docs/api/ref/AppInstance.md
@@ -10,7 +10,7 @@ hide_title: true
 interface AppInstance {
   //properties
   readonly instanceId : string;
-  readonly status : 'ready' | 'loading' | 'unregistered';
+  readonly status : AppInstanceStatus;
   
   //methods
   addContextListener(handler: ContextHandler): Listener;
@@ -18,6 +18,12 @@ interface AppInstance {
   broadcast(context: Context): void;
   onStatusChanged(handler : (newVal : string, oldVal :string) => {}) : void;
 
+}
+
+enum AppInstanceStatus {
+  Ready = 'ready',
+  Loading = 'loading',
+  Unregistered = 'unregistered',
 }
 ```
 
@@ -47,7 +53,7 @@ Uniquely identifies the instance. This identifier is internal to the desktop age
 ### `status`
 
 ```ts
-public readonly status: 'ready' | 'loading' | 'unregistered';
+public readonly status: AppInstanceStatus;
 ```
 
 Represents the lifecycle status of the AppInstance.  
@@ -144,7 +150,7 @@ try {
 ### `onStatusChannged`
 
 ``` typescript
-public onStatusChanged(handler : (newVal : string, oldVal :string) => {}) : void;
+public onStatusChanged(handler : (newVal : AppInstanceStatus, oldVal : AppInstanceStatus) => {}) : void;
 ```
 
 Sets a listener for status change events on an app instance.
@@ -163,13 +169,13 @@ Possible status values are:
 const instance = await fdc3.getAppInstance(source);
 
 // if status is ready, add a context listener
-if (instance.status === 'ready'){
+if (instance.status === AppInstanceStatus.Ready){
     const listener = instance.addContextListener(someContext, myListener);
 }
 
 // listen for status change, if status is unregistered, then unsubscribe the context listener
 instance.onStatusChanged((status) => {
-    if (status === 'unregistered'){
+    if (status === AppInstanceStatus.unregistered){
         listener.unsubscribe();
     }
 });
--- a/docs/api/ref/DesktopAgent.md
+++ b/docs/api/ref/DesktopAgent.md
@@ -9,7 +9,7 @@ hide_title: true
 ```ts
 interface DesktopAgent {
   // apps
-  open(name: string, context?: Context): Promise<void>;
+  open(name: string, context?: Context): Promise<AppInstance | null>;
   getAppInstance(instanceId : string ) : Promise<AppInstance>;
   
   // context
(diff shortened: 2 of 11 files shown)


src/api/DesktopAgent.ts
```diff
@@ -165,4 +166,38 @@ export interface DesktopAgent {
...
+   *  //handle immediate context data payload (if any)
+   *  if (intentRes.data){
+   *    updatePrice(intentRes.data);
+   *  }
+   * 
+   * 
+   * ```
+   * 
+   */
+  getAppInstance(instanceId : string ) : Promise<AppInstance>;
```
mattjamieson, line 202: For `getCurrentChannel` we specify returning null if the channel doesn't exist (although the return type doesn't allow it, and I think there is a separate PR to fix that). Should we do that here too?:

```js
getAppInstance(instanceId: string) : Promise<AppInstance | null>;
```

docs/api/ref/DesktopAgent.md
```diff
@@ -294,7 +295,44 @@ redChannel.addContextListener(channelListener);
...
 
+### `getAppInstance`
 
+  ```ts
+  getAppInstance(instanceId : string ) : Promise<AppInstance>;
+  ```
+
+ Returns the instance of an app for a given identifier.
+  
+  An instanceId can be obtained from
```
mattjamieson, line 306: In #279 we're adding:
```ts
type Target = string | AppMetadata;
```

If/when we merge that we should clarify the language here as I think it only makes sense to `getAppInstance` with a `string` appId and not an `AppMetadata` object?

docs/api/ref/AppInstance.md
```diff
@@ -0,0 +1,177 @@
...
+
+```ts
+interface AppInstance {
+  //properties
+  readonly instanceId : string;
+  readonly status : 'ready' | 'loading' | 'unregistered';
+  
+  //methods
+  addContextListener(handler: ContextHandler): Listener;
+  addContextListener(contextType: string, handler: ContextHandler): Listener;
```
mattjamieson, line 17: I still think we should fix the arg order for all the `addContextListener` methods 😁

docs/api/ref/AppInstance.md
```diff
@@ -0,0 +1,177 @@
...
+title: AppInstance
+hide_title: true
+---
+# `AppInstance`
+
+```ts
+interface AppInstance {
+  //properties
+  readonly instanceId : string;
+  readonly status : 'ready' | 'loading' | 'unregistered';
```
mattjamieson, line 13: Can we type the status?
```ts
type AppInstanceStatus = 'ready' | 'loading' | 'unregistered';
```

docs/api/ref/AppInstance.md
```diff
@@ -0,0 +1,177 @@
...
+  
+  //methods
+  addContextListener(handler: ContextHandler): Listener;
+  addContextListener(contextType: string, handler: ContextHandler): Listener;
+  broadcast(context: Context): void;
+  onStatusChanged(handler : (newVal : string, oldVal :string) => {}) : void;
```
mattjamieson, line 19: As above:
```ts
onStatusChanged(handler : (newVal : AppInstanceStatus, oldVal :AppInstanceStatus) => void) : void;
```

mattjamieson (commented):
Great to start to define `AppInstance` as a concept, definitely seems like something we've been touching on more and more in conversations lately.

Added some comments inline.

docs/api/ref/DesktopAgent.md
```diff
@@ -10,6 +10,7 @@ hide_title: true
 interface DesktopAgent {
   // apps
   open(name: string, context?: Context): Promise<void>;
```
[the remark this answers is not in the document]
nkolba: Yes that would make sense.  I think returning the instance itself would be appropriate for `open` and consistent with `window.open` in the DOM.

docs/api/ref/DesktopAgent.md
nkolba: I think the instanceId here is different from AppMetadata being provided for the `Target`.  

In the case of Target, we are addressing the app - regardless of instance - and matching on either name/appId  or metadata and can return any instance (or even spawn a new one).

docs/api/ref/AppInstance.md
nkolba: 2.0?

docs/api/ref/AppInstance.md
nkolba: Yes.  I was being lazy...

docs/api/ref/DesktopAgent.md
mattjamieson: I agree - it's just that #279 changed `IntentResolution.source` to be of type `Target` so we probably need to be clear that this usage is only for the `string` variant.

Or:

docs/api/ref/AppInstance.md
rikoe: They actually cause a lot of problems, being out of order. The code you have to write to deal with it is really ugly.

rikoe (commented):
Looks good to me @nkolba - should we remove the `data` property from `IntentResolution` as part of this?

docs/api/spec.md
```diff
@@ -155,6 +163,36 @@ On the financial desktop, applications often want to broadcast context to any nu
 ## Resolvers
 Intents functionality is dependent on resolver functionality to map the intent to a specific App.  This will often require end-user input.  Resolution can either be performed by the Desktop Agent (raising UI to pick the desired App for the intent) or by the app launching the intent - in which case the calling App will handle the resolution itself (using the findIntents API below) and then invoke an explicit Intent object.
 
+## App Instances
+Using FDC3 APIs, an app can get a reference to an instance of another app that it has either raised an intent to or recieved an intent or context from.  The `AppInstance` API allows apps to listen for and broadcast context directly.  
+
+### Broadcasting and Listening for Context
+An AppInstance MUST support both *broadcast* and *addContextListener* methods.  Calling *broadcast* on an AppInstance MUST send the context only *to* that instance.  The *addContextListener* method MUST listen for context events only *from* that instance.
```
pgn-vole, line 170: > *addContextListener* method MUST listen for context events only *from* that instance.

How can the application instance, when using `fdc3.broadcast`, differenciate between broadcasting to a channel, that it has joined previously, and broadcasting to an actual "appInstance listener"?
I understand that the application instance can do that differenciation with the `fdc3.addContextListener` API because the handler does provide a source property, which implies that the new context is not coming from a channel broadcasting. But it is not clear on how this distinction can be done with `fdc3.broadcast`.

nkolba: @pgn-vole instance messages can only be handled/passed via addContextListener and broadcast on the instance.  So, fdc3.broadcast would never get routed to an instance.addContextListener handler and vice-versa.  This is consistent with broadcast/handler behavior on channels.  The only exception being when an app has actually joined a channel - in which case - the channel context is the same as the global broadcast.  I will look amending the spec to make that more clear.  

BTWY - The source property won't provide this differentiation - you will always have a single instance source - whether this is coming from a channel, a direct message from an instance, or a global broadcast.  The distinction is only in where the app sets the listener.

src/api/DesktopAgent.ts
nkolba: @mattjamieson after some more thought, I think we should leave as is.  If the instance doesn't exist, you can still get an instance reference and get lifecycle events if it loads, set handlers on it, etc.

docs/api/ref/DesktopAgent.md
nkolba: k.  maybe it would make more sense if there was a type of InstanceId - instead of "TargetSource".  e.g 

```
type InstanceId = string;
type Target = InstanceId | AppMetadata;
```
This would make the type names more aligned and accurate.  (what you are targeting above would either be an exact InstanceId or a match on AppMetadata)

docs/api/ref/AppInstance.md
nkolba: @rikoe @mattjamieson - if we made this change in 1.2, what would be the impact to existing apps?  Can some of the platform owners weigh in?  @kriswest @lspiro-Tick42 @tim-dinsdale

docs/api/ref/DesktopAgent.md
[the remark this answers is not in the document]
nkolba: @mattjamieson updated PR

docs/api/ref/AppInstance.md
nkolba: @mattjamieson updated PR

docs/api/ref/AppInstance.md
nkolba: @mattjamieson updated PR

docs/api/spec.md
pgn-vole: Thanks for the explanation. So for an application instance to broadcast a message to any of its listener it must beforehand retrieve its own instance and then broadcast via it like the below?

_Listener:_ 
```js
const chart = await fdc3.raiseIntent('ViewChart');
const instance = await fdc3.getAppInstance(chart.source);

instance.addContextListener("someContext", handler);
```

_"ViewChart" Application:_ 
```js
const instance = await fdc3.getAppInstance(myInstanceId); // myInstanceId is exposed in some way by the desktop agent

instance.broadcast({type: "someContext"}) //broadcast its app instance listeners
fdc3.broadcast({type: "someContext"})// broadcast to currently joined channel
```

nkolba: Not exactly. An app shouldn't broadcast on its own instance - but instead, broadcast on the instance of a subscribing app (retrieved from the intentListener or contextListener). So, in the "ViewChart" example above, you would do this instead:

```js
fdc3.addIntentListener("ViewChart",async (context, source) => {
   const sourceInstance = await fdc3.getAppInstance(source);
   sourceInstance.broadcast({type: "someContext"})
});
```

pgn-vole: Thanks that makes more sense. I got confused by the section "Context last value cache" which made me assume that there would be a single context per application instance, however it seems that instead there will be a context held for each relation provider/subscriber.

docs/api/ref/AppInstance.md
```diff
@@ -0,0 +1,183 @@
...
+  Loading = 'loading',
+  Unregistered = 'unregistered',
+}
+```
+
+Represents a specific instance of an app.  
+
+AppInstances can be used to directly listen for and broadcast context between two apps connected to a desktop agent.
+
+An AppInstance is obtained by calling the `getAppInstence` method on the `DesktopAgent` using a source token provided by either an IntentResolution or as an argument to a ContextHandler.
```
sean-ciq, line 34: just a small typo: `getAppInstence` should be `getAppInstance`

docs/api/ref/AppInstance.md
thorsent: Speaking of jumping through hoops to make function signatures work, the API would be easier to implement and more consistent if addContextListener() were to return a `Promise<Listener>` rather than a `Listener`.

docs/api/spec.md
kriswest: We had similar issues with initial comprehension here, due to the slightly unconventional (but perhaps necessary) approach.

@nkolba When you say that (the app should) "broadcast on the instance of a subscribing app (retrieved from the intentListener or contextListener)" how does it do that retrieval? It would have needed to register a listener of some sort for the addition of those listeners in order to retrieve their IDs. That seems too complex to be practical.

Perhaps what @pgn-vole suggests is simpler (retrieving its own AppInstance and broadcasting on that) OR perhaps some object is returned by addIntentListener for that purpose (this does not solve the open case however).

kriswest (changes requested):
This is a great start @nkolba. 

The biggest outstanding issues I/we see are: 
- the approach is unconventional (though necessary), leading to some problems with comprehension. 
  - This might be resolved through documentation and naming conventions (e.g. replacing `source` in IntentResolution with `appInstanceId`. 
- The return types of `open` and `raiseIntent` are inconsistent. `open` returns a `Promise<AppInstance | null>` while `raiseIntent` returns an `IntentResolution` containing a `source` string (presumed to have changed from the apps name to appInstanceId). These should be harmonized.
- There is no way to retrieve the AppMetadata associated with an InstanceId - this could be solved by: 
  - including the AppMetadata in the AppInstance objects returned 
  - OR by adding the InstanceId to the AppMetadata object and returning that
  - OR by creating an InstanceMetadata object that wraps the InstanceId and an AppMetadata object
- There is some inconsistency arising from the fact that raiseIntent accepts a `Target` (see https://github.com/finos/FDC3/pull/279) to be renamed `TargetApp` (https://github.com/finos/FDC3/pull/315) but broadcast and addContextListener are called on the AppInstance.
  - The proposal should incorporate the merged changes in https://github.com/finos/FDC3/pull/279 and proposed improvement in https://github.com/finos/FDC3/pull/315
  - we should consider whether it would be more consistent to: 
    - further refine the `TargetApp` to allow use of `InstanceMetadata` or an `AppMetadata` with instance id
    - OR allow raiseIntent on the AppInstance object (as the proposal does with broadcast and addContextListener)
    - OR add a target option to the desktop agent's broadcast and addContextListener functions instead of creating an AppInstance object at all (i.e. redefine this proposal using instance metadata only and not an object.
- The proposed method of responding to an app that has called `addContextListener` on an AppInstance (retrieving that app's own `AppInstance` and broadcasting to it) has issues. 
  - The responding app currently has no way to get the `InstanceId` of the subscribing app in order to retrieve it.
    - In the case of `fdc3.open`, to resolve the `open` issues, an app would have to register a listener for subscriptions and then retain Ids or `AppInstance` objects to use for this purpose.
    - In the case of `fdc3.raiseIntent` the `ContextHandler` argument on `addIntentListener` would need to have an additional argument added to it to receive the source instanceId.  
    - This feels far too complex to be practical. 
  - As this is the primary usecase being addressed this is perhaps the most important issue to address in the proposal.

kriswest:
On further review, we also note that there is a danger of an app losing the association of a raised intent with a response with this approach, which implements communications between two instances after an intent is raised, but doesn't necessarily relate those comms to a specific intent. What happens if an app raises more than one intent and targets the same instance? The same conduit would have to be used to reply to both intents, with messages potentially being interleaved and needing additional context metadata to relate them to the specific intents.

However, there already exists the concepts of app channels that could be used to implement the comms between instances. As an application can create and join more than one channel it can uniquely tie a stream of responses to a single event. Alternatively, it can also reuse a channel (e.g. if more than one app requests the same pricing stream). 

Using channels would also eliminate the duplication of functionality that the proposed AppInstances create (you can already `broadcast` and `addContextListener` on the desktop agent and channels, this proposal would create a third set of these functions). Working with a channel returned by `raiseIntent` or `open` would also be easier to understand conceptually.

src/api/methods.ts
```diff
@@ -6,11 +6,12 @@ import {
...
+import { AppInstance } from './AppInstance';
 
-export const open: (name: string, context?: Context) => Promise<void> = (
-  name,
-  context
-) => {
+export const open: (
+  name: string,
+  context?: Context
+) => Promise<AppInstance | null> = (name, context) => {
```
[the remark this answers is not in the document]
nkolba: Intent resolution was handled in this way to preserve backwards compatibility.  Given that this change is being pushed to a 2.x, then breaking changes to IntentResolution could be on the table.  Have the intent resolve to an AppInstance (consistent with open) seems to make sense.

docs/api/spec.md
```diff
@@ -141,16 +141,24 @@ catch (er){
...
+There are a wide range of workflows where decoupled intents and/or context passing do not provide rich enough interactivity and applications are better off exposing proprietary APIs.  In these cases, an App can use the *source* property available on the resolution of an intent and when recieving context to get a direct reference to the calling app. For example: 
 
 ```js
 const chart = await fdc3.raiseIntent('ViewChart');
-// construct a vendor wrapper for the App
-const chartApp = fin.Application.wrap(chart.source);
-// do some vendor-specific stuff
+
+//get an AppInstance object back from the DesktopAgent
+const instance = await fdc3.getAppInstance(chart.source);
```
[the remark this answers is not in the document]
nkolba: see comment above.  Also, note that 'source' was never specifically defined as app 'name' - left up to implementation by design.

kriswest:
Superceded by: 
- #432 
- #433 
- #450

PR #300, opened 2021-01-21
Updating README.md

agitana:
Hi @rikoe I've added a description of the PMC, fixed a couple of typos and re-arranged some sections.
I wonder if the roadmap needs updating?

Best,
Aitana


README.md
```diff
@@ -14,19 +14,40 @@ The mission of the Financial Desktop Connectivity and Collaboration Consortium (
 ## About this Repository
 FDC3 is a standard.  This repository houses the documentation and specifications for the ratified FDC3 standards. 
 
+## Contribute to the FDC3 Standard
```
rikoe, line 17: @agitana I think how to use the standard is more important than how to contribute, so the contribution section should therefore come after the usage section.

rikoe (changes requested):

README.md
agitana: thanks @rikoe I've rearranged that section

rikoe (approved):

PR #319, opened 2021-02-08
Icons spec

### LINKED ISSUE #257: Application definition "icons" property
### Question Area
[X] App Directory
[ ] API
[ ] Context Data
[ ] Intents
[ ] Use Cases
[ ] Other

##  Question

Could you please clarify how does an application launcher should interpret the icons property of applications within the app directory? 

It is specified for this property to have the below shape:
```jsx
{
 // other app properties
 icons : [
   {  
     icon: "https://foo.com/icon.png"
   }
 ]
}
```

Since `icons` is a list, which item of that list should the launcher use and as an application developer how do I know which item of that list the launcher is going to pick?

I had a look at 2 open source implementations [1] and the logic appears flimsy to me considering that they do arbitrarily pick the first item of that list.

I would think in the first place that multiple icons would be useful so that different sizes of the same icon or even different icons could be exposed to launchers.  However to enable this capability some metadata properties would have to be added to the "icon" model. At the moment it does support a single property "icon" which is meant to be just the url.

1: https://github.com/ChartIQ/finsemble-fdc3/blob/master/src/components/intentResolver/src/app.tsx#L204
   https://github.com/finos/fdc3-desktop-agent/blob/master/src/content.ts#L338

pgn-vole:
Improves App Metadata icons specification. It is heavily based on [PWA manifest's icon spec](https://developer.mozilla.org/en-US/docs/Web/Manifest/icons).

Fixes #257


src/app-directory/specification/appd.yaml
```diff
@@ -347,9 +347,15 @@ components:
     Icon:
       description: Icon holder
       properties:
-        icon:
+        src:
```
pgn-vole, line 350: This is a breaking change. Considering that this is meant to be part of the "2.0" release I thought that it would be  I acceptable. 
If not, I suggest to either 
 - not introduce `src`  and use the `icon` property as image url
 - deprecate icon in favour of a newly added `src` property. In that case Desktop Agent may treat `icon` as it would treat `src`

docs/api/ref/Icon.md
```diff
@@ -0,0 +1,55 @@
...
+    "size": "72x72"
+  }
+]
+```
+
+## Properties
+
+### `src`
+
+The fully qualified url to the icon.
```
pgn-vole, line 41: Please note that in the PWA manifest spec the url may be relative. However in the context of FDC3 I couldn't find where a Desktop agent may retrieve the base url of an application so that when given a relative url it could construct the fully qualified url.
That is the reason why only fully qualified url are allowed here.

pgn-vole:
@rikoe - would you mind having a look at that PR when have some free time?

rikoe:
> @rikoe - would you mind having a look at that PR when have some free time?

@pgn-vole plan to! this has been earmarked for v2.0 so we can't merge yet, will take a look and discuss at standard working group once 1.2 is out.

rikoe:
@pgn-vole do you mind resolving the conflicts (rebasing), then we can merge this one in?

pgn-vole:
@rikoe it should be good now.

rikoe (approved):

mattjamieson (approved):

Issue #306, opened 2021-01-27
Support for ready event in v1.2 FDC3 npm package imports


rikoe:
Add support for the new v1.2 ready event when resolving the fdc3 object in ES6 imports exposed via the npm package.

rikoe:
Implemented as part of #360

PR #327, opened 2021-02-25
Docs: fdc3ready event docs

kriswest:
Improving docs for ready event added in #269  
Needs details adding on ES6 use of ready event to be dealt with in #306 

@rikoe

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,8 +112,6 @@ if (window.fdc3) {
 }
 ```
 
-[comment]: # (insert instructions for fdc3Ready event use via ES6 syntax here)
-
 ## Native
 The FDC3 standard does not define wire formats for communication. Hence, for native applications to be FDC3-enabled, they need to make use of a library (e.g. a DLL in .Net or Jar file in Java) that provides them with an implementation of the FDC3 API. FDC3-enabled native applications are therefore specific to particular desktop container frameworks (or other suitable environments) that provide the necessary libraries. 
 
--- a/README.md
+++ b/README.md
@@ -98,6 +98,22 @@ const channel = await getOrCreateChannel('myChannel')
 channel.broadcast({ ... })
 ```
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -98,6 +98,22 @@ const channel = await getOrCreateChannel('myChannel')
...
+
+if (window.fdc3) {
+  fdc3Stuff();
+} else {
+  window.addEventListener("fdc3Ready", fdc3Stuff);
+}
+```
+
+[comment]: # (insert instructions for fdc3Ready event use via ES6 syntax here)
+
```
rikoe, line 116: @kriswest I think we should leave this out of the README for now. It is not meant to be exhaustive, the README just has some very basic examples. I think the right place to properly document it is through the documentation website, if we try to duplicate all the docs in the readme it will get very big.

mattjamieson: The ES6 methods all throw/reject if the global is not available anyway, so I think whatever we do there is additive anyway. I'm thinking about just:
```ts
import { onReady, broadcast } from '@finos/fdc3';
```

README.md
```diff
@@ -98,6 +98,20 @@ const channel = await getOrCreateChannel('myChannel')
 channel.broadcast({ ... })
 ```
 
+To prevent the API from being used before it is ready, a global `fdc3Ready` event is provided:
```
rikoe, line 101: Sorry @kriswest - my previous comment was meant for everything in the readme, I didn't intend to show how we obtain fdc3 object in it - just some brief usage examples.

To keep the readme front page concise, I think let's cover fdc3 ready in the spec, and in docs, but not here.

kriswest: In the original request you mentioned the 'getting started section' I wasn't sure what that was and interpreted as the readme... Will strip this out, but let me know if there is somewhere else it should be

rikoe (approved):

Issue #326, opened 2021-02-24
Mark items as deprecated in 1.2 release for removal in future major version


rikoe:
The proposal is to mark the following features as deprecated in FDC3 1.2, for removal as part of the next major version of FDC3:
- [x] `IntentResolution.data` property (#219)
- [x] `global` channel (#308)
- [x] problematic signature of `addContextListener` (#215)

PR #341, opened 2021-03-08
feat(spec): add deprecation notices to API spec

rikoe:
- deprecation of 'global' channel
- deprecation of IntentResolution.data property

Relates to #326

### DIFF
--- a/docs/api/spec.md
+++ b/docs/api/spec.md
@@ -208,11 +208,13 @@ Context channels allows a set of apps to share a stateful piece of data between 
 
 There are two types of channels, which are functionally identical, but have different visibility and discoverability semantics.
 
-1. The 'system' ones, which have a well understood identity. One is called 'global'.
+1. The 'system' channels, which have a well understood identity.
 
-    > **Deprecation notice:** In future versions of FDC3, there won't be a `global` channel, see [below](#the-global-channel).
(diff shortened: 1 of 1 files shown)


docs/api/spec.md
```diff
@@ -202,6 +209,9 @@ Context channels allows a set of apps to share a stateful piece of data between
 There are two types of channels, which are functionally identical, but have different visibility and discoverability semantics.
 
 1. The 'system' ones, which have a well understood identity. One is called 'global'.
```
kriswest, line 211: I suggest removing this and incorporating into the statement below.
```suggestion
1. The 'system' ones, which have a well understood identity. 
```

docs/api/spec.md
```diff
@@ -202,6 +209,9 @@ Context channels allows a set of apps to share a stateful piece of data between
...
+
+    > **Deprecation notice:** In future versions of FDC3, there won't be a `global` channel, see [below](#the-global-channel).
```
kriswest, line 213:
```suggestion
    > **Deprecation notice:** Earlier versions of FDC3 include the concept of a 'global' channel for backwards compatibility with FDC3 1.0. In future, there won't be a `global` channel, see [below](#the-global-channel).
```

kriswest (approved):
LGTM - I had one suggestion that you can incorporate (as I think it's cleaner) or leave at your discretion.

PR #375, opened 2021-04-27
resolves #361 (support raise Intent without a context) 

### LINKED ISSUE #361: Support raiseIntent without a context
## Enhancement Request

The signature for raiseIntent currently requires a context object, however, there are use-cases where one may not be required. Change:

`raiseIntent(intent: string, context: Context, app?: TargetApp): Promise<IntentResolution>;`
to:
`raiseIntent(intent: string, context?: Context | null, app?: TargetApp): Promise<IntentResolution>;`

which would allow calls of the form:
- `raiseIntent(intent, context, target)`
- `raiseIntent(intent, null, target)`
- `raiseIntent(intent)`

### Use Case:
An intent expresses action, which is useful to combine with a context, but not always required. 

E.g.
- `raiseIntent("ViewNews")` might open an intent resolver allowing the user to select an application in which to view (unfiltered) news.
- `raiseIntent("ViewNews", target)` might open a specific application (e.g. the Factset terminal) with multiple features and display the News panel.
- `raiseIntent("CreateOrder")` might open a blank order ticket.
- `raiseIntent("StartCall")` might open a VoIP application with the dialling form or contact selector focused.

kriswest:
…by adding an explicit Null context type

N.B. I couldn't run `yarn typegen` as the schema is not yet deployed to the web.

N.B. 2 I got some weird output from git on this commit @rikoe 
image

### DIFF
--- a/docs/api/ref/DesktopAgent.md
+++ b/docs/api/ref/DesktopAgent.md
@@ -378,7 +378,7 @@ raiseIntent(intent: string, context: Context, app?: TargetApp): Promise<IntentRe
 ```
 Raises a specific intent against a target app.
 
-The desktop agent will resolve the correct app to target based on the provided intent name and context data.
+The desktop agent will resolve the correct app to target based on the provided intent name and context data type. If you wish to raise an Intent without a context, use the `fdc3.null` context type. This type exists so that apps can explicitly declare support for raising an intent without context.
 
 If multiple matching apps are found, the user may be presented with an app picker.
 Alternatively, the specific app to target can also be provided (if known).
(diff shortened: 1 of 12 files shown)


rikoe:
I would still prefer the context type used for this to be `fdc3.nothing` instead of `fdc3.null` - can we discuss this in an upcoming meeting @kriswest?

Reason being that `null` is a technical term, while the context types are more business-oriented. And I believe it makes the intention clearer.

@mattjamieson @ggeorgievx @donbasuno any thoughts?

kriswest:
`fdc3.null` changed `fdc3.nothing` throughout - this should now be ready to merge (post-review)

mattjamieson (approved):

rikoe (approved):

PR #392, opened 2021-05-22
Docs: fixing broken link

davidwatkins73:

PR #393, opened 2021-05-24
Docs: fixing broken link

kriswest:
Add https://github.com/finos/FDC3/pull/392 to the 1.2 version of docs.


rikoe (approved):
@kriswest this is the right way to do it - you have to manually update the previously taken snapshot of the versioned docs.

PR #454, opened 2021-09-15
Clarify existing behavior of joinChannel and addIntentListener when joining a channel

### LINKED ISSUE #418: fdc3.joinChannel docs should explain expected behavior on joinChannel
### Area of Issue
[ ] App Directory
[ ] API
[ ] Context Data
[ ] Intents
[ ] Use Cases
[X] Other (documentation)

### Issue Description: 
The documentation for `fdc3.joinChannel()` does not clarify that the desktop agent should immediately call any context listeners already registered via `fdc3.addContextListener` with the current context as defined in the spec:
> When an app joins a channel it will automatically recieve the current context for that channel.
This presumably implies the same behaviour as the [Channel API's `getCurrentContext` function](https://fdc3.finos.org/docs/api/ref/Channel#getcurrentcontext), where, if a type is passed the last context of that type broadcast on the channel is returned, rather than just the last context. However, this is not clarified in either the spec or documentation.

Both the specification and documentation should be updated to make the expected behaviour clear.

kriswest:
Resolves #418 

Also: 
- shuffles sentence order in spec for clarity (doesn't change meaning)
- switches examples using 'agent.' to 'fdc3.' for consistency
- fixes a couple of minor formatting inconsistencies

### DIFF
--- a/docs/api/ref/DesktopAgent.md
+++ b/docs/api/ref/DesktopAgent.md
@@ -55,8 +55,9 @@ addContextListener(contextType: string | null, handler: ContextHandler): Listene
  */
 addContextListener(handler: ContextHandler): Listener;
 ```
-Adds a listener for incoming context broadcast from the Desktop Agent. If the consumer is only interested in
-a context of a particular type, they can use the relevant overload that allows the type to be specified.
+Adds a listener for incoming context broadcasts from the Desktop Agent. If the consumer is only interested in a context of a particular type, they can they can specify that type. If the consumer is able to receive context of any type or will inspect types received, then they can pass `null` as the `contextType` parameter to receive all context types. 
+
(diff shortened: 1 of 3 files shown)


docs/api/ref/DesktopAgent.md
```diff
@@ -55,8 +55,9 @@ addContextListener(contextType: string | null, handler: ContextHandler): Listene
  */
 addContextListener(handler: ContextHandler): Listener;
 ```
-Adds a listener for incoming context broadcast from the Desktop Agent. If the consumer is only interested in
-a context of a particular type, they can use the relevant overload that allows the type to be specified.
+Adds a listener for incoming context broadcasts from the Desktop Agent. If the consumer is only interested in a context of a particular type, they can they can specify that type. If the consumer is able to receive context of any type or will inspect types received, then they can pass `null` as the `contextType` parameter to receive all context types.
```
mattjamieson, line 58: "they can they can" -> "they can"

mattjamieson (changes requested):

mattjamieson (approved):

PR #462, opened 2021-09-23
fix: update Methods addContextListener to support new signature

mattjamieson:


src/api/Methods.ts
```diff
@@ -67,13 +67,16 @@ export function addIntentListener(intent: string, handler: ContextHandler): List
...
 
-export function addContextListener(contextTypeOrHandler: string | ContextHandler, handler?: ContextHandler): Listener {
+export function addContextListener(
+  contextTypeOrHandler: string | null | ContextHandler,
+  handler?: ContextHandler
+): Listener {
   if (typeof contextTypeOrHandler !== 'function') {
     return throwIfNoGlobal(() =>
       window.fdc3.addContextListener(contextTypeOrHandler as string, handler as ContextHandler)
     );
```
kriswest, line 77: I don't think you need the cast to string (as it could also be null). If removed, prettier also wants it back on a single line.

```suggestion
  if (typeof contextTypeOrHandler !== 'function') {
    return throwIfNoGlobal(() => window.fdc3.addContextListener(contextTypeOrHandler, handler as ContextHandler));
```

src/api/Methods.ts
```diff
@@ -67,13 +67,16 @@ export function addIntentListener(intent: string, handler: ContextHandler): List
   return throwIfNoGlobal(() => window.fdc3.addIntentListener(intent, handler));
 }
 
-export function addContextListener(contextTypeOrHandler: string | ContextHandler, handler?: ContextHandler): Listener {
+export function addContextListener(
+  contextTypeOrHandler: string | null | ContextHandler,
+  handler?: ContextHandler
+): Listener {
   if (typeof contextTypeOrHandler !== 'function') {
```
kriswest, line 74:
```suggestion
  //Handle (deprecated) function signature that allowed contextType argument to be omitted
  if (typeof contextTypeOrHandler !== 'function') {
```

kriswest (approved):
Added a comment and minor suggestion, otherwise LGTM.

kriswest:
@mattjamieson This PR is still marked as a draft, could you commit the suggestions and move it on?

I think deal with the export issue in a separate PR

rikoe (approved):

kriswest (approved):
LGTM

Issue #452, opened 2021-09-07
Channels, Feeds & Transactions Discussion group meeting 9th Sept 2021


kriswest:
## Group overview
There are currently 13 open issues in the FDC3 repository, under consideration for the 2.0 release, that relate to the existing concept of channels (both the 'system channels' used via the desktop agent API, and 'app channels' interacted with through the Channels API) and new concepts we've discussed introducing: transactions (a standardized way to make a request and receive a response back) and feeds (a standardized way to request a stream of data, such as a pricing stream).

### Relevant issue tags
Current open issues that relate to the above concepts with the label: 
[![](https://user-images.githubusercontent.com/1701764/120999635-422ebd00-c781-11eb-8033-7f6ff251605b.png)](https://github.com/finos/FDC3/labels/channels%20feeds%20%26%20transactions)

## Meeting Date
Thursday 9th Aug 2021 - 9am EDT / 2pm BST

### WebEx info
- [Meeting Link](https://finos.webex.com/finos/j.php?MTID=mf2940a718824a41e0747dada7d34e271)
- Meeting number: 127 494 1096
- Password: 3cqFgG7fMT2

*More ways to join*
- Join by video system:
   - Dial 1274941096@finos.webex.com
   - You can also dial 173.243.2.68 and enter your meeting number.
- Join by phone
   - +1-415-655-0003 US Toll
   - +44-20319-88141 UK Toll
- Access code: 127 494 1096

## Meeting notices
- FINOS **Project leads** are responsible for observing the FINOS guidelines for [running project meetings](https://github.com/finos/community/blob/master/governance/Meeting-Procedures.md#run-the-meeting). Project maintainers can find additional resources in the [FINOS Maintainers Cheatsheet](https://odp.finos.org/docs/finos-maintainers-cheatsheet/).

- **All participants** in FINOS project meetings are subject to the [LF Antitrust Policy](https://www.linuxfoundation.org/antitrust-policy/), the [FINOS Community Code of Conduct](https://github.com/finos/community/blob/master/governance/Code-of-Conduct.md) and all other [FINOS policies](https://github.com/finos/community/tree/master/governance#policies).

[...]

PR #464, opened 2021-09-28
Resolves #434 Clarify broadcast of complex context types

### LINKED ISSUE #434: Clarify how apps should handle broadcast and receipt of multiple context types
## Enhancement Request
When sharing context over channels, you can `addContextListner` for all context or for particular types of context (which will filter the messages you receive). However, the specification does not advise on whether an application, intended to link to multiple other applications consuming a variety of different types, should broadcast multiple times with different types or expect the other applications to listen for its preferred type and extract what they need for it. 

### Use Case:
A particular application (e.g. an OMS) might broadcast details of orders (lets call that `fdc3.order` for the sake of the example). A type representing an order might be composed of several other types, including:
- A Position (`fdc3.position`)
- An Instrument (`fdc3.instrument`)
- An Organisation (`fdc3.organization`)
- A Contact (`fdc3.contact`)

That application might be linked (via a System channel) to other components that use particular types as context, e.g.
- a CRM (`fdc3.contact`),
- an account management tool (`fdc3.organization`)
- a pricing chart (`fdc3.instrument`)
- a news feed (`fdc3.instrument`)
- a risk tool (`fdc3.position`)
- a notification or order update tool (`fdc3.order`)
- etc.

### Possible solutions
Should the OMS:
- Broadcast just the fdc3.order type? 
  - This encodes all the required information for all the other components
  - However the other components would have to be expecting it and/or ready to extract the types they need from it
- Broadcast all 5 types separately one after another
  - Each component can listen for only the type they require
  - Generates additional load on the system by multiplying each broadcast
  - Potentially need to rebroadcast every type or need to know which other types are expected to be re-broadcast.

This question may need to be answered for components that use different types to be able to work together without prior coordination. That answer could form advice in the specification OR could update it (e.g. to have context listeners added to channels automatically extract types they are listening for).

Note: it is possible to write a general purpose function for searching for and extracting a specified type from another type that is composed from it:

```javascript
/** Recursively search the input context and return an array of matches */
let extractContextTypeHelper = (contextReceived, typeDesired, list) => {
    if (contextReceived?.type == typeDesired) {
        //if we found the type we want return it
        return [contextReceived];
    } else if (contextReceived instanceof Array){
        //if we received an array we need to check each item, return a match, or return a list of each entries children to check
        for (let i=0; i < contextReceived.length; i++){
            if (contextReceived[i]?.type == typeDesired) {
                //if we found the type we want return it
                return contextReceived[i];
            }
            let found = extractContextTypeHelper(contextReceived[i], typeDesired, [])
            if (found) {
                list = list.concat(found);
            } 
        }
        return extractContextTypeHelper(list, typeDesired, []);
    } else if (typeof contextReceived == "object"){
        //if we received an object thats not the right type, extract its fields and add to our list
        let keys = Object.keys(contextReceived);
        for (let i=0; i < keys.length; i++) {
            let found = extractContextTypeHelper(contextReceived[keys[i]], typeDesired, []);
            if (found) {
                list = list.concat(found);
            } 
        }

        return list;
    }

    return list;
}

/** Return the first match found or null if none is found */
let extractContextType = (contextReceived, typeDesired) => {
    let found = extractContextTypeHelper(contextReceived, typeDesired, []);
    if (found && found.length > 0) {
        return found[0];
    }
    return null;
};

/** Example of usage */

let example = {
  type: "fdc3.order",
  instrument: {
    type: "fdc3.instrument",
    name: "AAPL",
    id: {
      ticker: "AAPL"
    }
  },
  amount: 100,
  direction: "buy",
  client: {
    type: "fdc3.client",
    organization: {
      type: "fdc3.organization",
      name: 'Cargill, Incorporated',
      id: {
        LEI: 'QXZYQNMR4JZ5RIRN4T31',
        FDS_ID: '00161G-E'
      }
    },
    contact: {
      type: 'fdc3.contact',
      name: 'Jane Doe',
      id: {
        email: 'jane.doe@mail.com'
      }
    }
  }
};

extractContextType(example, "fdc3.instrument")
extractContextType(example, "fdc3.contact")  
extractContextType(example, "fdc3.organization")
extractContextType(example, "fdc3.order")
```

kriswest:
Adds advice to the spec and documentation about how to broadcast complex context types, where other apps on the channel may wish to/are set up to process simpler types that are the fields of your context type. The consensus from #452 was to broadcast each type you want others to respond to separately.  
E.g. when broadcasting an `fdc3.position`,  which has an `fdc3.instrument` field, broadcast instrument and then position.

### DIFF
--- a/docs/api/ref/Channel.md
+++ b/docs/api/ref/Channel.md
@@ -136,6 +136,8 @@ If the broadcast is denied by the channel or the channel is not available, the m
 
 Channel implementations should ensure that context messages broadcast by an application on a channel should not be delivered back to that same application if they are joined to the channel.
 
+If you are working with complex context types composed of other simpler types (as recommend by the [Context specification](../../context/spec#assumptions)) then you should broadcast each individual type that you want other apps to be able to respond to. Doing so allows applications to filter the context types they receive by adding listeners for specific context types.
+
 #### Example
 
 ```javascript
--- a/docs/api/ref/DesktopAgent.md
+++ b/docs/api/ref/DesktopAgent.md
@@ -104,6 +104,8 @@ Publishes context to other apps on the desktop.  Calling `broadcast` at the `Des
 
 DesktopAgent implementations should ensure that context messages broadcast to a channel by an application joined to it should not be delivered back to that same application.
 
+If you are working with complex context types composed of other simpler types (as recommend by the [Context specification](../../context/spec#assumptions)) then you should broadcast each individual type that you want other apps to be able to respond to. Doing so allows applications to filter the context types they receive by adding listeners for specific context types.
+
 #### Example
 ```js
 const instrument = {
--- a/docs/api/spec.md
+++ b/docs/api/spec.md
@@ -272,6 +272,13 @@ joinedChannel = await fdc3.getCurrentChannel()
 ### Direct Listening and Broadcast on Channels
 While joining channels automates a lot of the channel behavior for an app, it has the limitation in that an app can belong to only one channel at a time.  Listening and Broadcasting to channels using the _Channel.addBroadcastListener_ and the _Channel.broadcast_ APIs provides an app with fine-grained controls for specific channels.  This is especially useful for working with dynamic _App Channels_.
 
+### Broadcasting and listening for multiple context types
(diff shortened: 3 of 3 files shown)


mattjamieson (approved):
I still think we can do better on this, perhaps by providing a mechanism for developers to register context adapters to extract the simple types from the composite type. Though maybe that's an opportunity for vendor platform differentiation. :grin:

PR #467, opened 2021-10-01
Update copyright notice

kriswest:
Updating the copyright notice to include claims for FDC3 contributors and FINOS

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -1,5 +1,5 @@
-FDC3 - FINOS
+FDC3 - Fintech Open Source Foundation (https://www.finos.org/)
+
 Copyright 2017-2019 OpenFin info@openfin.co
-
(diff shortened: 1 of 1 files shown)


mattjamieson (approved):

rikoe (approved):

kriswest:
@openfin-gavin we're updating the repository copyright notice and were advised that we can remove the dates from the notices to avoid them going out of date. Could you confirm that we're ok to do that to the OpenFin claim as well for consistency? If not we'll leave as is.

openfin-gavin:
> @openfin-gavin we're updating the repository copyright notice and were advised that we can remove the dates from the notices to avoid them going out of date. Could you confirm that we're ok to do that to the OpenFin claim as well for consistency? If not we'll leave as is.

I'm checking on this

PR #473, opened 2021-10-08
Ensure system channels are not retrieved until FDC3 API is ready

kriswest:
@donbasuno pointed out that the workbench isn't waiting for the FDC3 ready event before retrieving system channels, meaning it can fail to retrieve them in some Desktop Agent implementations. This occurs, despite the fact that the workbench doesn't render its UI until FDC3 ready has fired, because the channels are retrieved on import of the Channels page, which imports the Channels store, which in turn retrieves the channels in its constructor. 

This PR fixes that issue and appears to ensure that the Channels are more stably retrieved from the FDC3 Desktop Agent Chrome extension, and should also resolve issues in the Finsemble 5 FDC3 add-on.

### DIFF
--- a/toolbox/fdc3-workbench/src/App.tsx
+++ b/toolbox/fdc3-workbench/src/App.tsx
@@ -158,13 +158,6 @@ export const App = observer(() => {
 				setFdc3Available(true);
 			} catch (e) {}
 		})();
-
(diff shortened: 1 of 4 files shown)


donbasuno (approved):
Ship it captain!

PR #478, opened 2021-10-15
Workbench: add error handling to getInfo call to prevent errors in OpenFin

kriswest:
fdc3.getInfo was introduced in FDC3 1.1 and, despite checking for the existence of the function, calls to it were throwing errors in the current OpenFin FDC3 service. This PR swaps the check for the existence of the function for error handling on calls to it, which should prevent further issues.

If `getInfo` is not implemented or throws an error the workbench will display version info as:
image

### DIFF
--- a/toolbox/fdc3-workbench/src/components/Header.tsx
+++ b/toolbox/fdc3-workbench/src/components/Header.tsx
@@ -50,7 +50,16 @@ export const Header = (props: { fdc3Available: boolean }) => {
 		if (props.fdc3Available) {
 			//getInfo is not available in FDC3 < v1.2, handle any errors thrown when trying to use it
 			try {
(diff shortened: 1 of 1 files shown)


donbasuno (changes requested):
Minor issue, but I think eg "Not Specified" is better than "unknown"?

kriswest:
I went with 'unknown' as this also needs to apply to cases where the desktop agent can't be found or an error occurs when calling the function. 'Not specified' would make sense if the call came back, but without the info.

I've tweaked the PR so that it will display "unknown" on errors and "not specified" whenever information is omitted. E.g. if the function is not implemented or errors out you now get:
image

Whereas for ommited info (e.g. in dev build of finsemble where its version number is unknown):
image

PR #495, opened 2021-11-11
Added the ability to return data from a raised intent

### LINKED ISSUE #432: Make it possible to return data from a raised intent
Output from the Channels, Feeds & Transactions discussion group: #420
Supercedes #201 & #372 

## Enhancement Request
It should be possible to return data from an application that is resolving a raised intent and for the application that raised that intent to receive it back. This was part of the original idea for intents (see https://fdc3.finos.org/docs/api/spec#raising-intents) , but was never correctly implemented in the spec such that it was possible to populate the `data` element in the `IntentResolution` data structure. 

### Use Case:
There are multiple existing use-cases that would benefit from being able to return data from a raised intent:

- Use Case 4: Client-side FX Trader Credit Check
  https://fdc3.finos.org/docs/use-cases/uc-4
- Use Case 9: Call Transcription to CRM
  https://fdc3.finos.org/docs/use-cases/uc-9
- Use Case 15: Sales Floor Base Workflow
  https://fdc3.finos.org/docs/use-cases/uc-15

Other examples include:
- raising an order (or other similar) ticket and receiving back a ticket number
- performing a credit check and receiving back a response
- retrieving a single price or another piece of market data from another application

Note: Channels are a poor fit for a single response:
- Overhead for creating/retrieving a channel, which has to happen at both ends
- You would probably need to return the name of the response channel, which is in-of-itself a use-case for a transaction
- Possible synchronization issues (was the response already sent before I joined?)

kriswest:
resolves #432 
Adds the ability for apps resolving intents to also return data that may be retrieved from the `IntentResolution` by the app that raised the intent.
- Introduces an IntentHandler type, similar to ContextHandler, but with an optional return type of Promise<Context>
- Adds a `getData()` function to IntentResolution
- Removes the deprecated `data` field of IntentResolution
- Adds details and examples to the spec and API reference

A different PR will deal with adding support for specifying the desired return type when resolving an intent through the API and for adding metadata to do so to the AppD.

### DIFF
--- a/docs/api/ref/DesktopAgent.md
+++ b/docs/api/ref/DesktopAgent.md
@@ -30,7 +30,7 @@ interface DesktopAgent {
   findIntentsByContext(context: Context): Promise<Array<AppIntent>>;
   raiseIntent(intent: string, context: Context, app?: TargetApp): Promise<IntentResolution>;
   raiseIntentForContext(context: Context, app?: TargetApp): Promise<IntentResolution>;
-  addIntentListener(intent: string, handler: ContextHandler): Listener;
+  addIntentListener(intent: string, handler: IntentHandler): Listener;
 
   // channels
   getOrCreateChannel(channelId: string): Promise<Channel>;
@@ -71,27 +71,39 @@ const contactListener = fdc3.addContextListener('fdc3.contact', contact => { ...
 #### See also
 * [`Listener`](Types#listener)
 * [`Context`](Types#context)
+* [`ContextHandler`](Types#contexthandler)
 
 
 
 ### `addIntentListener`
 
 ```ts
-addIntentListener(intent: string, handler: ContextHandler): Listener;
+addIntentListener(intent: string, handler: IntentHandler): Listener;
 ```
- Adds a listener for incoming Intents from the Agent.
+Adds a listener for incoming Intents from the Agent. The handler function may return void or a promise that should resolve to a context object representing any data that should be returned to app that raised the intent. If an error is thrown by the handler function, or the promise returned is rejected, then the Desktop Agent MUST reject the promise returned by the `getData()` function of the `IntentResolution`. 
 
 #### Examples
 
 ```js
+//Handle a raised intent
 const listener = fdc3.addIntentListener('StartChat', context => {
-  // start chat has been requested by another application
+    // start chat has been requested by another application
+    return;
+});
+
+//Handle a raised intent and return Context data via a promise
+fdc3.addIntentListener("trade", (context) => {
+    return new Promise<Context>((resolve) => {
+        // go place trade
+        resolve({result: "trade placed successfully"});
+   });
 });
 ```
 
 #### See also
 * [`Listener`](Types#listener)
 * [`Context`](Types#context)
+* [`IntentHandler`](Types#intenthandler)
 
 
 
@@ -389,7 +401,7 @@ Alternatively, the specific app to target can also be provided. A list of valid 
 
 If you wish to raise an Intent without a context, use the `fdc3.nothing` context type. This type exists so that apps can explicitly declare support for raising an intent without context.
 
-Returns an `IntentResolution` object with details of the app that was selected to respond to the intent.
+Returns an `IntentResolution` object with details of the app that was selected to respond to the intent. If the application that resolves the intent returns a promise of Context data, this may be retrieved via the `getData()` function of the IntentResolution object. If an error is thrown by the handler function, or the promise returned is rejected, then the Desktop Agent MUST reject the promise returned by the `getData()` function of the `IntentResolution`. 
 
 If a target app for the intent cannot be found with the criteria provided, an `Error` with a string from the [`ResolveError`](Errors#resolverrror) enumeration is returned.
 
@@ -432,7 +444,7 @@ Alternatively, the specific app to target can also be provided, in which case an
 
 Using `raiseIntentForContext` is similar to calling `findIntentsByContext`, and then raising an intent against one of the returned apps, except in this case the desktop agent has the opportunity to provide the user with a richer selection interface where they can choose both the intent and target app.
 
-Returns an `IntentResolution` object with a handle to the app that responded to the selected intent.
+Returns an `IntentResolution` object with details of the app that was selected to respond to the intent. If the application that resolves the intent returns a promise of Context data, this may be retrieved via the `getData()` function of the IntentResolution object. If an error is thrown by the handler function, or the promise returned is rejected, then the Desktop Agent MUST reject the promise returned by the `getData()` function of the `IntentResolution`. 
 
 If a target app for the intent cannot be found with the criteria provided, an `Error` with a string from the [`ResolveError`](Errors#resolveerror) enumeration is returned.
 
--- a/docs/api/ref/Metadata.md
+++ b/docs/api/ref/Metadata.md
@@ -126,21 +126,44 @@ The Interface used to describe an Intent within the platform.
 
 ```ts
 interface IntentResolution {
-  source: TargetApp;
+/**
+   * The application that resolved the intent.
+   */
+  readonly source: TargetApp;
   /**
-  * @deprecated not assignable from intent listeners
-  */
-  data?: object;
-  version: string;
+   * The version number of the Intents schema being used.
+   */
+  readonly version?: string;
+  /**
+   * Retrieves a promise that will resolve to data returned by the
+   * application that resolves the raised intent. The promise will 
+   * reject if an error is thrown by the intent handler or the promise
+   * returned by the intent handler is reject. If the intent handler 
+   * does not return a promise this function will return null.
+   */
+  getData(): Promise<Context> | null;
 }
 ```
 
 IntentResolution provides a standard format for data returned upon resolving an intent.
 
-#### Example
+#### Examples
 ```js
-// resolve a "Chain" type intent
-const intentResolution = await fdc3.raiseIntent("intentName", context);
+//resolve a "Chain" type intent
+let resolution = await agent.raiseIntent("intentName", context);
+
+//resolve a "Client-Service" type intent with a data response
+let resolution = await agent.raiseIntent("intentName", context);
+try {
+	   const result = await resolution.getData();
+    if (result) {
+        console.log(`${resolution.source} returned ${JSON.stringify(result)}`);
+    } else {
+        console.error(`${resolution.source} didn't return data`
+    }
+} catch(error) {
+    console.error(`${resolution.source} returned an error: ${error}`);
+}
 ```
 
 #### See also
--- a/docs/api/spec.md
+++ b/docs/api/spec.md
@@ -42,7 +42,7 @@ The global `window.fdc3` must only be available after the API is ready to use. T
 
 ```js
 function fdc3Stuff() {
-  // Make some fdc3 API calls here
+  // Make fdc3 API calls here
 }
 
 if (window.fdc3) {
@@ -99,11 +99,13 @@ Intents provide a way for an app to request functionality from another app and d
 - **Remote API**: An app wants to remote an entire API that it owns to another App.  In this case, the API for the App cannot be standardized.  However, the FDC3 API can address how an App connects to another App in order to get access to a proprietary API.
 
 #### Intents and Context
-When raising an Intent a specific context may be provided. The type of the provided context may determine which applications can resolve the Intent.
+When raising an Intent a specific context may be provided as input. The type of the provided context may determine which applications can resolve the Intent.
 
 A Context type may also be associated with multiple Intents. For example, an `fdc3.instrument` could be associated with `ViewChart`, `ViewNews`, `ViewAnalysis` or other Intents. In addition to raising a specific intent, you can raise an Intent for a specific Context allowing the Desktop Agent or the user (if the Intent is ambiguous) to select the appropriate Intent for the selected Context and then to raise that Intent for resolution.
 
 To raise an Intent without a context, use the `fdc3.nothing` context type. This type exists so that applications can explicitly declare that they support raising an intent without a context (when registering an Intent listener or in an App Directory).
+
+An optional context object may also be returned as output by an application resolving an intent. For example, an application resolving a `CreateOrder` intent might return a context representing the order and including an ID, allowing the application that raised the intent to make further calls using that ID.  
 
 #### Intent Resolution
 Raising an Intent will return a Promise-type object that will resolve/reject based on a number of factors.
@@ -119,27 +121,28 @@ Raising an Intent will return a Promise-type object that will resolve/reject bas
 
 ##### Resolution Object
 
-> **Deprecation notice**
->
-> It is not currently possible to provide a value for the `data` property described below,
-as intent listeners don't currently offer a way to return values.
->
-> Future versions of FDC3 plan to remove the optional `data` property from the intent resolution object,
-and include a more robust mechanism for intents that need to return data back to the caller.
-
 If the raising of the intent resolves (or rejects), a standard object will be passed into the resolver function with the following format:
 
 ```js
 {
-    source: String;
-    data?: Object;
-    version: String;
+  /**
+   * The application that resolved the intent.
+   */
+  readonly source: TargetApp;
+  /**
+   * The version number of the Intents schema being used.
+   */
+  readonly version?: string;
+  /**
+   * Retrieves a promise that will resolve to data returned by the
+   * application that resolves the raised intent. The promise will 
+   * reject if an error is thrown by the intent handler or the promise
+   * returned by the intent handler is reject. If the intent handler 
+   * does not return a promise this function will return null.
+   */
+  getData(): Promise<Context> | null;
 }
 ```
-- *source* = identifier for the Application resolving the intent (null if the intent could not be resolved)
-- *data* = return data structure - if one is provided for the given intent
-- *version* = the version number of the Intents schema being used
-
 
 For example, to raise a specific Intent:
 
@@ -152,7 +155,7 @@ catch (er){
 }
 ```
 
-or to raise an unspecified Intent for a specific context, where the user will select an intent from a resolver dialog:
+or to raise an unspecified Intent for a specific context, where the user may select an intent from a resolver dialog:
 ```js
 try {
     const result = await fdc3.raiseIntentForContext(context);
@@ -162,6 +165,21 @@ try {
 }
 catch (er){
     console.log(er.message);
+}
+```
+
+Raise an intent and retrieve data from the IntentResolution:
+```js
+let resolution = await agent.raiseIntent("intentName", context);
+try {
+    const result = await resolution.getData();
+    if (result) {
+        console.log(`${resolution.source} returned ${JSON.stringify(result)}`);
+    } else {
+        console.error(`${resolution.source} didn't return data`
+    }
+} catch(error) {
+    console.error(`${resolution.source} returned an error: ${error}`);
 }
 ```
 
--- a/src/api/DesktopAgent.ts
+++ b/src/api/DesktopAgent.ts
@@ -5,7 +5,7 @@
 
 import { AppIntent } from './AppIntent';
 import { Channel } from './Channel';
-import { ContextHandler, TargetApp } from './Types';
+import { ContextHandler, IntentHandler, TargetApp } from './Types';
 import { IntentResolution } from './IntentResolution';
 import { Listener } from './Listener';
 import { Context } from '../context/ContextTypes';
@@ -122,25 +122,40 @@ export interface DesktopAgent {
   /**
    * Raises a specific intent for resolution against apps registered with the desktop agent.
    *
-   * The desktop agent will resolve the correct app to target based on the provided intent name and context data. If multiple matching apps are found, the user may be presented with an app picker.
+   * The desktop agent MUST resolve the correct app to target based on the provided intent name and context data. If multiple matching apps are found, the user MAY be presented with a Resolver UI allowing them to pick one, or another method of Resolution applied to select an app.
    * Alternatively, the specific app to target can also be provided. A list of valid target applications can be retrieved via `findIntent`.
    *
-   * Returns an `IntentResolution` object with details of the app that was selected to respond to the intent.
+   * Returns an `IntentResolution` object with details of the app that was selected to respond to the intent. If the application that resolves the intent returns a promise of Context data, this may be retrieved via the `getData()` function of the IntentResolution object. If an error is thrown by the handler function, or the promise returned is rejected, then the Desktop Agent MUST reject the promise returned by the `getData()` function of the `IntentResolution`.
    *
    * If a target app for the intent cannot be found with the criteria provided, an `Error` with a string from the `ResolveError` enumeration is returned.
    *
    * ```javascript
    * // raise an intent for resolution by the desktop agent
-   * // a resolver UI will be displayed if more than one application can resolve the intent
+   * // a resolver UI may be displayed if more than one application can resolve the intent
    * await fdc3.raiseIntent("StartChat", context);
+   *
    * // or find apps to resolve an intent to start a chat with a given contact
    * const appIntent = await fdc3.findIntent("StartChat", context);
    * // use the returned AppIntent object to target one of the returned chat apps by name
    * await fdc3.raiseIntent("StartChat", context, appIntent.apps[0].name);
    * // or use one of the AppMetadata objects returned in the AppIntent object's 'apps' array
    * await fdc3.raiseIntent("StartChat", context, appIntent.apps[0]);
+   *
    * //Raise an intent without a context by using the null context type
    * await fdc3.raiseIntent("StartChat", {type: "fdc3.nothing"});
+   *
+   * //Raise an intent and retrieve data from the IntentResolution
+   * let resolution = await agent.raiseIntent("intentName", context);
+   * try {
+   * 	   const result = await resolution.getData();
+   *     if (result) {
+   *         console.log(`${resolution.source} returned ${JSON.stringify(result)}`);
+   *     } else {
+   *         console.error(`${resolution.source} didn't return data`
+   *     }
+   * } catch(error) {
+   *     console.error(`${resolution.source} returned an error: ${error}`);
+   * }
    * ```
    */
   raiseIntent(intent: string, context: Context, app?: TargetApp): Promise<IntentResolution>;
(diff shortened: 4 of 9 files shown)


nkolba:
@kriswest  great to see data intents being picked up again!  A lot to think about here, but my first question  would be why change out the _data_ property for the _getData_ method?  _raiseIntent_  returns a promise, so not clear what the extra async call buys us.  Beyond adding an extra call, downsides I'd be concerned about would be:
- the desktop agent now has to maintain state to pass context back with an intent, as they will need to hold the return context for that specific raiseIntent until _getData_ is called.  
- previously, the intentResolution type was a JSON data structure - it now has a method - which limits flexibility of implementations.

Would be very helpful to hear more of the thinking behind this.

thanks!

-Nick

kriswest:
Hi @nkolba,  although I raised the PR the decision to use this approach came from the [Channels, Feeds & Transactions group](https://github.com/finos/FDC3/issues?q=is%3Aissue+label%3A%22channels+feeds+%26+transactions%22). You can find more documentation of the discussion on the issue: https://github.com/finos/FDC3/issues/432

TL;DR: the data return might not be immediate, there could be operations to perform or user interactions to wait for (e.g. raising a `CreateOrder` intent might display and fill out an order ticket, which the user needs to check and confirm). If the data is directly included in the IntentResolution that cannot be returned until the operation completes. Whereas, if you can retrieve a promise for the data, it can be returned immediately, ensuring that the app that raised the intent knows which app is supposed to resolve it. 
 
Regarding the Desktop Agent holding the returned context until `getData` is called, that's true, although it could be passed from any _service_ implementing the desktop agent to the specific _client_ (in the window that raised the intent) so that the only reference is a resolved promise that can be garbage collected with the IntentResolution when no longer needed. I originally proposed that `data` become that promise (which I believe helps with your first point, if not your second), but others felt it would make more sense to provide a function to retrieve that promise.

bertrand-s:
I must say I share @nkolba concerns. To be fair I am not clear on the added value of a 2 step process.
As a developer, when I see a function returning a promise I expect the promise to resolve when the function has executed. In this proposal the main promise would be resolved when the intent is dispatched and then we would have to call the 2nd method to get another promise that would resolve when the actual function has been executed. This seems very complicated. What would be the use case of awaiting the first promise and not the 2nd one? 
I feel like we have two cases:
- either we want to raise an intent as fire and forget (so we don't even need the first promise)
- or we want to know when the intent is actually processed (even if there is not actual return value) as it allows to chain subsequent actions.
If we are afraid of performance issues, then why not having an extra argument to the intent method - e.g.:
```js
raiseIntent(intent: string, context: Context, app?: TargetApp, fireAndForget=true): Promise<IntentResolution>;
```
Besides, promises indeed introduce context, and personally I would rather prefer the context to be handled by the Desktop agent rather than by the apps as it would be easier to fix issues. For instance,
- how to manage backward compatibility? i.e. old apps that expect fire-and-forget intents and that will not return anything?
- what happens if an app crashes while fulfilling an intent?
In both cases I feel like we would need some kind of timeouts - that could either be handled by the desktop agents or by the FDC3 wrapper library (so on the caller side)

src/api/IntentResolution.ts
```diff
@@ -3,23 +3,44 @@
...
+  /**
+   * Retrieves a promise that will resolve to data returned by the
+   * application that resolves the raised intent. The promise MUST
+   * reject with a string from the `DataError` enumeration if an error
+   * is thrown by the intent handler, it rejects the returned promise,
+   * or it does not return a promise.
    */
-  readonly data?: object;
-  readonly version: string;
+  getData(): Promise<Context>;
```
bertrand-s, line 45: As discussed, couldn't we rename this field ```getResult()``` as it would allow to handle potential future use cases where the result is not a data structure (cf. discussions on notifications)

That said, I don't think we need to have a method here (i.e. a Promise would be enough as you suggested). I guess most implementation will keep the promise as a private hidden field that would be returned by getData/getResult method. Indeed we don't want to keep the data in the Agent until the method is called - so the agent is likely to send the result once received and store it in a private field in the client structure to make it available through getData... (so there is no point in this extra complexity..)

thorsent:
@nkolba & @bertrand-s (to add to @kriswest) these points are quite valid and were discussed heavily.

The compelling case for a two-stage process was made by several participants, pointing specifically to a common use case in our industry:
- An intent to place a trade should result in (1) a confirmation that the trade was received by the OMS, and (2) a data response that contains a trade identifier (which is  often assigned asynchronously). This situation was further clarified in that OMS's often prompt a modal dialog after stage (1) and that, therefore, intent-raising apps should not be frozen during that stage.

The group was swayed, that the additional complexity was warranted given the prevalence of this use case in our industry.

In regards to a function call vs. a plain promise - the initial design did return a plain promise but it was argued that a function that returns a promise would provide a more familiar developer interface. We reworked the approach to compare these approaches and the group agreed that the resulting function call did result in more readable code. This is certainly a stylistic consideration though.

FWIW, an addendum PR is forthcoming that introduces streams into this concept which I think will offer a holistic view for consideration, nomenclature sweeps, etc.

nkolba:
@thorsent @kriswest  @bertrand-s  In reply to Terry's comments.  I still don't get what we're solving for with _getData()_ .

1. the desktop agent is free to provide end users with any level of feedback they want to that the intent has been routed.  For example, in the DesktopAgent chrome extension, the intent dialog closes after a selection is made and the tab of the app processing the intent is brought into focus.  ie. in your example, the OMS the intent was routed to could be brought into focus with the modal for the end user to ack.  
2. there is nothing requiring blocking on the app raising the intent while it is waiting for the intent to resolve.  e.g in the case of data intents, an app could just display a localized loading animation

Has anyone tried implementing these workflows with the _data_ prop?  I did some initial implementations last year - as part of the PR I had made and did not find these concerns to be issues.

-Nick

src/api/Types.ts
```diff
@@ -6,5 +6,21 @@
...
 export type TargetApp = string | AppMetadata;
+/**
+ * Describes a callback that handles a context event.
+ * Used when attaching listeners for context broadcasts.
+ */
 export type ContextHandler = (context: Context) => void;
+/**
+ * Describes a callback that handles a context event and may return a
+ * promise of a Context object to be returned to the application that
+ * raised the intent.
```
bertrand-s, line 23: I guess we could also add that if void is returned by the IntentHandler a Promise will still be still returned to the intent caller - but it will only mean that the intent was properly dispatched and its execution started.

src/api/IntentResolution.ts
kriswest: My only concern on using `getResult()` is that it will require a union return type (`Promise<Context | Channel>`), typescript is good at those, but other languages are not and although we use Typescript to define the API it should ideally be implementable in other languages. 

I would tend to agree on a promise vs. a function returning a promise. Using a function does mean either retaining the data in the desktop agent service or a hidden field, as opposed to it living in the promise itself. However, I have had to demonstrate to a couple of people recently that you can `await` promises, not just functions returning promises, so I was willing to accept that developers might be more familiar with calling a function.

src/api/Types.ts
kriswest: That was the intent (pun intended ;-) ). I'll attempt to clarify.

```suggestion
 * raised the intent. Any response other than Promise indicates that the handler does not return data (which will result in the Desktop Agent rejecting the IntentResolution.getData() promise with the appropriate error message).
```

kriswest: Does that help @bertrand-s ?

kriswest:
Hi @bertrand-s 
> - how to manage backward compatibility? i.e. old apps that expect fire-and-forget intents and that will not return anything?

This proposal is backwards compatible as nothing requires you to call or await the `getData()` function's promise, ensuring the fire-and-forget intents work exactly the same way as they do now. 

> - either we want to raise an intent as fire and forget (so we don't even need the first promise)

Raising an intent is inherently an async process, as it involves the delivery of a request/data to another system running in a separate process, potentially mediated by user interaction in a resolver UI.

As it stands currently (and since FDC3 1.0) you don't *have* to await the `raiseIntent` promise for the `IntentResolution` either. You only need to do so if you need the info in the IntentResolution (such as the app that resolved the intent) or to know if the resolution was successful (indicated by the promise resolving or rejecting with a [ResolveError](https://fdc3.finos.org/docs/api/ref/Errors#resolveerror)). This proposal doesn't change that.

Returning data from intent is a two-step process (both steps of which are async operations as described above): 
1. resolution/delivery of the intent to an application
2. handling by that application/return of data

If we combine those steps so that they are mediated by a single promise, that will introduce a change in behaviour, as the IntentResolution can't be returned until the handler has finished running, which again might be mediated by a second user interaction (e.g. submitting a form or some other interaction). It would also mean a combined error channel for both resolution and the return of data; if we successfully delivered the intent and context to an app, but it returned an error we would have to reject the promise of the intent resolution AND would then lack the data structure for expressing which app was responsible for the error. This relates to one of your comments:

> what happens if an app crashes while fulfilling an intent?
In both cases I feel like we would need some kind of timeouts - that could either be handled by the desktop agents or by the FDC3 wrapper library (so on the caller side)

The advantage of separating the promises is that, if we trust our Desktop Agent and its resolver UI, we don't need a timeout on the raisedIntent and can find out where it sent our request from the IntentResolution. We can then await the data promise (if we're expecting data), which will reject if an error is thrown. We can implement our own timeout on that wait, without having to have the desktop agent implement it (although that is something we can discuss - the discussion group liked the idea of leaving that to the app raising the intent/expecting data).  


@nkolba 
> the desktop agent is free to provide end users with any level of feedback they want to that the intent has been routed. For example, in the DesktopAgent chrome extension, the intent dialog closes after a selection is made and the tab of the app processing the intent is brought into focus. ie. in your example, the OMS the intent was routed to could be brought into focus with the modal for the end user to ack.

Web containers managing many windows on a multi-screen desktop represent a significantly different form factor to a browser. Focusing a tab is indeed a salient action in a tabbed browser, but far less so in a desktop where it is already visible. I have also recently been reminded by other participants that an Intent resolver UI is not the only possible way to handle intent resolution and that we need to support appropriate feedback through the API. Further, this feedback is intended for the app raisingIntent, and in this case receiving data back, rather than the app resolving the intent. We also considered use-cases where the resolving app has no UI (e.g. a service for retrieving data) and any and all feedback is in the raising application. 

> there is nothing requiring blocking on the app raising the intent while it is waiting for the intent to resolve. e.g in the case of data intents, an app could just display a localized loading animation.

If you don't use an async construct such as a promise and put the data directly in the `IntentResolution` then that does require that you block until the handler has run (+any required user interactions) and returned data that you can then put in the resolution. A promise of the data was introduced specifically to prevent the need for that blocking. 

If, on the other hand, the objection relates to having to call `getData()` to retrieve the promise, then I'm actually with you on that. Nothing really requires a function to return it as you can await any expression that evaluates to a promise (i.e. you can just as easily `await resolution.dataPromise` as `await resolution.getData()`) however other participants felt using a function would be more familiar to most developers (it could conceivable also be easier to implement in other languages, although I have no specific examples to support that).

kriswest:
P.S. Thank you both (@bertrand-s and @nkolba) for scrutinizing the PR! I'm happy to make minor changes to it and to clarify design decisions in docs if you think it necessary. Alternative proposals can also be submitted and we can make time to compare them to this one in a further meeting (let me know if you want to do that).

src/api/IntentResolution.ts
kriswest: Hmm, we also need that union type as the return type for the IntentHandler:
```typescript
export type IntentHandler = (context: Context) => Promise<Context | PrivateChannel> | void;
```

Perhaps we should introduce an explicate union type (as we do for [`TargetApp`](https://fdc3.finos.org/docs/api/ref/Types#targetapp)). While that doesn't change anything in Typescript, it could be a class in other languages that can hold either response type?

kriswest: Updated to getResult() (which will return a union type in a later PR when we add support for returning a channel)

docs/api/spec.md
```diff
@@ -119,27 +121,28 @@ Raising an Intent will return a Promise-type object that will resolve/reject bas
...
+   */
+  readonly version?: string;
+  /**
+   * Retrieves a promise that will resolve to data returned by the
+   * application that resolves the raised intent. The promise will 
+   * reject if an error is thrown by the intent handler or the promise
+   * returned by the intent handler is reject. If the intent handler 
+   * does not return a promise this function will return null.
+   */
+  getResult(): Promise<Context> | null;
```
kriswest, line 143:
```suggestion
  /**
   * Retrieves a promise that will resolve to data returned by the
   * application that resolves the raised intent. If an error is thrown by the handler function, 
   * the promise returned is rejected, or no promise is returned then the Desktop Agent 
   * MUST reject the promise returned by the `getResult()` function of the `IntentResolution` 
   * with a string from the `DataError` enumeration.
   */
  getResult(): Promise<Context>;
```

docs/api/ref/Metadata.md
```diff
@@ -126,21 +126,40 @@ The Interface used to describe an Intent within the platform.
...
+   */
+  readonly version?: string;
+  /**
+   * Retrieves a promise that will resolve to data returned by the
+   * application that resolves the raised intent. The promise will 
+   * reject if an error is thrown by the intent handler or the promise
+   * returned by the intent handler is reject. If the intent handler 
+   * does not return a promise this function will return null.
+   */
+  getResult(): Promise<Context> | null;
```
kriswest, line 144:
```suggestion
  /**
   * Retrieves a promise that will resolve to data returned by the
   * application that resolves the raised intent. If an error is thrown 
   * by the handler function,  the promise returned is rejected, or 
   * no promise is returned then the Desktop Agent MUST reject 
   * the promise returned by the `getResult()` function of the 
   * `IntentResolution` with a string from the `DataError` enumeration.
   */
  getResult(): Promise<Context> | null;
```

docs/api/ref/Metadata.md
```diff
@@ -126,21 +126,41 @@ The Interface used to describe an Intent within the platform.
...
+  readonly version?: string;
+  /**
+   * Retrieves a promise that will resolve to data returned by the
+   * application that resolves the raised intent. If an error is thrown 
+   * by the handler function,  the promise returned is rejected, or 
+   * no promise is returned then the Desktop Agent MUST reject 
+   * the promise returned by the `getResult()` function of the 
+   * `IntentResolution` with a string from the `DataError` enumeration.
+   */
+  getResult(): Promise<Context> | null;
```
kriswest, line 145:
```suggestion
  getResult(): Promise<Context>;
```

docs/api/ref/DesktopAgent.md
```diff
@@ -71,27 +71,39 @@ const contactListener = fdc3.addContextListener('fdc3.contact', contact => { ...
...
-  // start chat has been requested by another application
+    // start chat has been requested by another application
+    return;
+});
+
+//Handle a raised intent and return Context data via a promise
+fdc3.addIntentListener("CreateOrder", (context) => {
+  return new Promise<Context>((resolve) => {
+    // go create the order
+    resolve({type: "fdc3.order", id: { "orderId": 1234}});
```
mattjamieson, line 98:
```suggestion
    resolve({type: "fdc3.order", id: { "orderId": 1234 }});
```

src/api/Types.ts
```diff
@@ -6,5 +6,21 @@
...
 export type TargetApp = string | AppMetadata;
+/**
+ * Describes a callback that handles a context event.
+ * Used when attaching listeners for context broadcasts.
+ */
 export type ContextHandler = (context: Context) => void;
+/**
+ * Describes a callback that handles a context event and may return a
+ * promise of a Context object to be returned to the application that
+ * raised the intent. Any response other than Promise indicates that the handler does not return data (which will result in the Desktop Agent rejecting the IntentResolution.getData() promise with the appropriate error message).
```
mattjamieson, line 23:
```suggestion
 * raised the intent. Any response other than Promise indicates that
 * the handler does not return data (which will result in the Desktop
 * Agent rejecting the IntentResolution.getData() promise with the
 * appropriate error message).
```

mattjamieson (changes requested):

mattjamieson (approved):

ggeorgievx (approved):

kriswest:
@greyseer256 I'll merge this one after you've had a chance to a take a look, it has the necessary other reviews.

docs/api/spec.md
```diff
@@ -99,12 +99,14 @@ Intents provide a way for an app to request functionality from another app and d
 - **Remote API**: An app wants to remote an entire API that it owns to another App.  In this case, the API for the App cannot be standardized.  However, the FDC3 API can address how an App connects to another App in order to get access to a proprietary API.
 
 #### Intents and Context
-When raising an intent a specific context may be provided. The type of the provided context may determine which applications can resolve the intent.
+When raising an intent a specific context may be provided as input. The type of the provided context may determine which applications can resolve the Intent.
```
greyseer256, line 102: 'intent' has mixed casing in this sentence

greyseer256 (approved):
> If an error is thrown by the handler function, the promise returned is rejected, or a promise is not returned then the Desktop Agent MUST reject the promise returned by the `getResult()` function of the `IntentResolution`.

This sentence occurs several times, I'm still having some difficulty parsing the logic. I'm taking it to mean this:

If an error is thrown by the handler function, the promise returned is rejected, or a promise is not returned, in which case the Desktop Agent MUST reject the promise returned by the `getResult()` function of the `IntentResolution`.

is that correct?

docs/api/spec.md
kriswest, line 102:
```suggestion
When raising an intent a specific context may be provided as input. The type of the provided context may determine which applications can resolve the intent.
```

docs/api/ref/DesktopAgent.md
```diff
@@ -483,7 +504,7 @@ Alternatively, the specific app or app instance to target can also be provided,
 
 Using `raiseIntentForContext` is similar to calling `findIntentsByContext`, and then raising an intent against one of the returned apps, except in this case the desktop agent has the opportunity to provide the user with a richer selection interface where they can choose both the intent and target app.
 
-Returns an `IntentResolution` object with a handle to the app that responded to the selected intent.
+Returns an `IntentResolution` object with details of the app that was selected to respond to the intent. If the application that resolves the intent returns a promise of Context data, this may be retrieved via the `getResult()` function of the IntentResolution object. If an error is thrown by the handler function, the promise returned is rejected, or a promse is not returned then the Desktop Agent MUST reject the promise returned by the `getResult()` function of the `IntentResolution` with a string from the `DataError` enumeration.
```
kriswest, line 507:
```suggestion
Returns an `IntentResolution` object with details of the app that was selected to respond to the intent. If the application that resolves the intent returns a promise of Context data, this may be retrieved via the `getResult()` function of the IntentResolution object. If an error occurs (i.e. an error is thrown by the handler function, the promise it returns is rejected, or a promse is not returned by the handler function) then the Desktop Agent MUST reject the promise returned by the `getResult()` function of the `IntentResolution` with a string from the `DataError` enumeration. 
```

docs/api/ref/DesktopAgent.md
```diff
@@ -443,15 +455,15 @@ Alternatively, the specific app or app instance to target can also be provided.
 
 If you wish to raise an intent without a context, use the `fdc3.nothing` context type. This type exists so that apps can explicitly declare support for raising an intent without context.
 
-Returns an `IntentResolution` object with details of the app instance that was selected (or started) to respond to the intent.
+Returns an `IntentResolution` object with details of the app that was selected to respond to the intent. If the application that resolves the intent returns a promise of Context data, this may be retrieved via the `getResult()` function of the IntentResolution object. If an error is thrown by the handler function, the promise returned is rejected, or a promse is not returned then the Desktop Agent MUST reject the promise returned by the `getResult()` function of the `IntentResolution` with a string from the `DataError` enumeration.
```
kriswest, line 458:
```suggestion
Returns an `IntentResolution` object with details of the app that was selected to respond to the intent. If the application that resolves the intent returns a promise of Context data, this may be retrieved via the `getResult()` function of the IntentResolution object. If an error occurs (i.e. an error is thrown by the handler function, the promise it returns is rejected, or a promse is not returned by the handler function) then the Desktop Agent MUST reject the promise returned by the `getResult()` function of the `IntentResolution` with a string from the `DataError` enumeration. 
```

docs/api/ref/Metadata.md
```diff
@@ -161,24 +161,28 @@ interface IntentResolution {
...
-
+  /**
+   * Retrieves a promise that will resolve to data returned by the
+   * application that resolves the raised intent. If an error is thrown 
+   * by the handler function,  the promise returned is rejected, or 
+   * no promise is returned then the Desktop Agent MUST reject 
+   * the promise returned by the `getResult()` function of the 
+   * `IntentResolution` with a string from the `DataError` enumeration.
+   */
+  getResult(): Promise<Context>;
```
kriswest, line 176:
```suggestion
  /**
   * Retrieves a promise that will resolve to data returned by the
   * application that resolves the raised intent. If an error occurs 
   * (i.e. an error is thrown by the handler function,  the promise 
   * returned by the handler function is rejected, or no promise is 
   * returned) then the Desktop Agent MUST reject the promise 
   * returned by the `getResult()` function of the `IntentResolution` 
   * with a string from the `DataError` enumeration.
   */
  getResult(): Promise<Context>;
```

docs/api/spec.md
```diff
@@ -142,14 +136,18 @@ interface IntentResolution {
...
    */
   readonly version?: string;
+  /**
+   * Retrieves a promise that will resolve to data returned by the
+   * application that resolves the raised intent. If an error is thrown by the handler function, 
+   * the promise returned is rejected, or no promise is returned then the Desktop Agent 
+   * MUST reject the promise returned by the `getResult()` function of the `IntentResolution` 
+   * with a string from the `DataError` enumeration.
+   */
+  getResult(): Promise<Context>;
```
kriswest, line 150:
```suggestion
  /**
   * Retrieves a promise that will resolve to data returned by the application that 
   * resolves the raised intent. If an error occurs (i.e. an error is thrown by the handler 
   * function, the promise returned is rejected, or no promise is returned) then the Desktop 
   * Agent MUST reject the promise returned by the `getResult()` function of the 
   * `IntentResolution` with a string from the `DataError` enumeration.
   */
  getResult(): Promise<Context>;
```

src/api/DesktopAgent.ts
```diff
@@ -170,26 +170,37 @@ export interface DesktopAgent {
...
    * Alternatively, the specific app or app instance to target can also be provided. A list of valid target applications and instances can be retrieved via `findIntent`.
-   * 
+   *
    * If you wish to raise an Intent without a context, use the `fdc3.nothing` context type. This type exists so that apps can explicitly declare support for raising an intent without context.
    *
-   * Returns an `IntentResolution` object with details of the app instance that was selected (or started) to respond to the intent.
+   * Returns an `IntentResolution` object with details of the app instance that was selected (or started) to respond to the intent. If the application that resolves the intent returns a promise of context data, this may be retrieved via the `getResult()` function of the `IntentResolution` object. If an error is thrown by the handler function, the promise returned is rejected, or no promise is returned then the Desktop Agent MUST reject the promise returned by the `getResult()` function of the `IntentResolution` with a string from the `DataError` enumeration.
```
kriswest, line 178:
```suggestion
   * Returns an `IntentResolution` object with details of the app instance that was selected (or started) to respond to the intent. If the application that resolves the intent returns a promise of context data, this may be retrieved via the `getResult()` function of the `IntentResolution` object. If an error occurs (i.e. an error is thrown by the handler function, the promise returned is rejected, or no promise is returned) then the Desktop Agent MUST reject the promise returned by the `getResult()` function of the `IntentResolution` with a string from the `DataError` enumeration.
```

src/api/DesktopAgent.ts
```diff
@@ -198,11 +209,11 @@ export interface DesktopAgent {
...
+   * Alternatively, the specific app or app instance to target can also be provided, in which case the resolver SHOULD only offer intents supported by the specified application.
    *
    * Using `raiseIntentForContext` is similar to calling `findIntentsByContext`, and then raising an intent against one of the returned apps, except in this case the desktop agent has the opportunity to provide the user with a richer selection interface where they can choose both the intent and target app.
    *
-   * Returns an `IntentResolution` object with a handle to the app that responded to the selected intent.
+   * Returns an `IntentResolution` object with details of the app that was selected to respond to the intent. If the application that resolves the intent returns a promise of Context data, this may be retrieved via the `getResult()` function of the IntentResolution object. If an error is thrown by the handler function, the promise returned is rejected, or no promise is returned then the Desktop Agent MUST reject the promise returned by the `getResult()` function of the `IntentResolution` with a string from the `DataError` enumeration.
```
kriswest, line 216:
```suggestion
   * Returns an `IntentResolution` object with details of the app that was selected to respond to the intent. If the application that resolves the intent returns a promise of Context data, this may be retrieved via the `getResult()` function of the IntentResolution object. If an error occurs (i.e. an error is thrown by the handler function, the promise returned is rejected, or no promise is returned) then the Desktop Agent MUST reject the promise returned by the `getResult()` function of the `IntentResolution` with a string from the `DataError` enumeration.
```

src/api/DesktopAgent.ts
```diff
@@ -216,9 +227,30 @@ export interface DesktopAgent {
   raiseIntentForContext(context: Context, app?: TargetApp): Promise<IntentResolution>;
 
   /**
-   * Adds a listener for incoming Intents from the Agent.
+   * Adds a listener for incoming Intents from the Agent. The handler function may
+   * return void or a promise that should resolve to a context object representing
+   * any data that should be returned to app that raised the intent. If an error is
+   * thrown by the handler function, the promise returned is rejected, or a promise
+   * is not returned then the Desktop Agent MUST reject the promise returned by the
+   * `getResult()` function of the `IntentResolution`.
```
kriswest, line 235:
```suggestion
   * Adds a listener for incoming Intents from the Agent. The handler function may
   * return void or a promise that should resolve to a context object representing
   * any data that should be returned to app that raised the intent. If an error occurs 
   * (i.e. an error is thrown by the handler function, the promise returned is rejected, or 
   * a promise is not returned) then the Desktop Agent MUST reject the promise returned 
   * by the `getResult()` function of the `IntentResolution`.
```

kriswest:
@greyseer256 I've clarified the sentence using parentheses wherever it appears, hopefully that's better:
> If an error occurs (i.e. an error is thrown by the handler function, the promise returned is rejected, or no promise is returned) then the Desktop Agent MUST reject the promise returned by the `getResult()` function of the `IntentResolution` with a string from the `DataError` enumeration.

and yes, there's a huge amount of duplication on this sentence. Its necessary on two different functions, who's detail is duplicated between the TS sources, docs and spec + same again for the IntentResolution object. One day I hope we'll get to generating the docs from the TS sources which will cut the duplication in half...

src/api/DesktopAgent.ts
```diff
@@ -216,9 +227,30 @@ export interface DesktopAgent {
   raiseIntentForContext(context: Context, app?: TargetApp): Promise<IntentResolution>;
 
   /**
-   * Adds a listener for incoming Intents from the Agent.
+   * Adds a listener for incoming Intents from the Agent. The handler function may
+   * return void or a promise that should resolve to a context object representing
+   * any data that should be returned to app that raised the intent. If an error occurs 
+   * (i.e. an error is thrown by the handler function, the promise returned is rejected, or 
+   * a promise is not returned) then the Desktop Agent MUST reject the promise returned
```
kriswest, line 234:
```suggestion
   * any data that should be returned to app that raised the intent. If an error occurs
   * (i.e. an error is thrown by the handler function, the promise returned is rejected, or
   * a promise is not returned) then the Desktop Agent MUST reject the promise returned
```

greyseer256 (approved):
Thanks @kriswest, the new wording is much clearer to me.

PR #499, opened 2021-11-12
498 Allow intents to be resolved on output type (where they return data)

### LINKED ISSUE #498: Make it possible to resolve intent based on output type as well as input
### Use Case:
Issue #432 (and associated PR #495) make it possible to return data from a raised intent. However, an application or resolver UI may wish resolve an intent based on the *output* Context data type, in addition to the input type, ensuring they know what data type to expect.

To do so, additional data representing this information must be added to the AppD's Application schema, while the `findIntent` and `raiseIntent` and related functions will need to take extra arguments facilitating the resolution.  

### Workflow Examples
For example, an application may allow the user to raise the ViewAnalysis intent, but require that data is returned encoding or referencing the analysis for inclusion within notes kept within the application. It either resolves the intent via `findIntent` followed by a targetted `raiseIntent`, or calls `raiseIntent` and relies on a resolver UI for resolution, and passes the extra detail on what type it expects to be returned. Only applications that can return the specified type are offered.

kriswest:
Adds an element to the AppD Application schema that can specify the Context type that intent can return and extends the `findIntent` API calls to be able to use it for resolution.

- resolves #498 
- Follows on from #495  (which should be merged first)

Support for this was not added to the `raiseIntent` APIs as: 
- a second optional string argument would be a problem (that it isn't on `findIntent`)
- this form of resolution is likely to be performed programmatically, rather than through a resolver UI


docs/api/ref/DesktopAgent.md
```diff
@@ -389,7 +436,7 @@ Alternatively, the specific app to target can also be provided. A list of valid
 
 If you wish to raise an Intent without a context, use the `fdc3.nothing` context type. This type exists so that apps can explicitly declare support for raising an intent without context.
 
-Returns an `IntentResolution` object with details of the app that was selected to respond to the intent.
+Returns an `IntentResolution` object with details of the app that was selected to respond to the intent. If the application that resolves the intent returns a promise of Context data, this may be retrieved via the `getResult()` function of the IntentResolution object. If an error is thrown by the handler function, the promise returned is rejected, or a promse is not returned then the Desktop Agent MUST reject the promise returned by the `getResult()` function of the `IntentResolution` with a string from the `DataError` enumeration.
```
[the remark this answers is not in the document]
thorsent: It's a mouthful so maybe more iterations.

I think "Implementation Notes" can help bridge the gap from very precise spec oriented language to something that is more accessible. It's an idea, take or leave.

docs/api/ref/DesktopAgent.md
```diff
@@ -432,7 +486,7 @@ Alternatively, the specific app to target can also be provided, in which case an
 
 Using `raiseIntentForContext` is similar to calling `findIntentsByContext`, and then raising an intent against one of the returned apps, except in this case the desktop agent has the opportunity to provide the user with a richer selection interface where they can choose both the intent and target app.
 
-Returns an `IntentResolution` object with a handle to the app that responded to the selected intent.
+Returns an `IntentResolution` object with details of the app that was selected to respond to the intent. If the application that resolves the intent returns a promise of Context data, this may be retrieved via the `getResult()` function of the IntentResolution object. If an error is thrown by the handler function, the promise returned is rejected, or a promse is not returned then the Desktop Agent MUST reject the promise returned by the `getResult()` function of the `IntentResolution` with a string from the `DataError` enumeration.
```
thorsent, line 489: Perhaps `See raiseIntent`? So we can keep the docs DRY?

docs/api/ref/Errors.md
```diff
@@ -36,6 +36,22 @@ Contains constants representing the errors that can be encountered when calling
 * [`DesktopAgent.findIntent`](DesktopAgent#findintent)
 * [`DesktopAgent.findIntentsByContext`](DesktopAgent#findintentsbycontext)
 
+## `DataError`
+
+```typescript
+enum DataError {
+  NoDataReturned = 'NoDataReturned',
+  IntentHandlerRejected = 'IntentHandlerRejected',
```
thorsent, line 44: We may need some more discussion around error handling. For instance, a request for a quote would likely return some sort of well-defined "Quote" context, but what if the end user is not authorized to receive that market data? We wouldn't want to muddy the Quote context with arbitrary error messages, but then the DataError enum type provides no information about why the intent was rejected, which in this case would result in quite a bit of potential confusion for the end user.

We may want to either (a) consider allowing at least a simple error string to be rejected/thrown, or (b) allow for the possibility of more than one context being returned (e.g. an Error context. Receiving apps may then disambiguate by examining the type member).

docs/api/spec.md
```diff
@@ -99,11 +99,15 @@ Intents provide a way for an app to request functionality from another app and d
...
 
-To raise an Intent without a context, use the `fdc3.nothing` context type. This type exists so that applications can explicitly declare that they support raising an intent without a context (when registering an Intent listener or in an App Directory).
+To raise an Intent without a context, use the [`fdc3.nothing`](../context/ref/Nothing) context type. This type exists so that applications can explicitly declare that they support raising an intent without a context (when registering an Intent listener or in an App Directory).
+
+An optional context object may also be returned as output by an application hanling an intent. For example, an application handling a `CreateOrder` intent might return a context representing the order and including an ID, allowing the application that raised the intent to make further calls using that ID.
```
thorsent, line 108:
```suggestion
An optional context object may also be returned as output by an application handling an intent. For example, an application handling a `CreateOrder` intent might return a context representing the order and including an ID, allowing the application that raised the intent to make further calls using that ID.
```

docs/api/spec.md
```diff
@@ -165,8 +170,19 @@ catch (er){
...
+} catch(error) {
+    console.error(`${resolution.source} returned a data error: ${error}`);
+}
+```
+
 #### Resolvers
-Intents functionality is dependent on resolver functionality to map the intent to a specific App.  This will often require end-user input.  Resolution can either be performed by the Desktop Agent (for example, by displaying a resolver UI allowing the user to pick the desired App for the intent) or by the app calling App handling the resolution itself (by using the `findIntents` API) and then invoking the Intent on a specific target application, e.g.:
+Intents functionality is dependent on resolver functionality to map the intent to a specific App.  This will often require end-user input.  Resolution can either be performed by the Desktop Agent (for example, by displaying a resolver UI allowing the user to pick the desired App for the intent) or by the app calling App handling the resolution itself (by using the  [`findIntent`](ref/DesktopAgent#findintent) or [`findIntentByContext`](ref/DesktopAgent#findintentbycontext) API calls) and then invoking the Intent on a specific target application, e.g.:
```
thorsent, line 185:
```suggestion
Successful delivery of an intent depends first upon the Desktop Agent's ability to "resolve the intent" (i.e. map the intent to a specific App instance). Desktop Agents may resolve intents by any methodology. A common methodology is to display a UI that allows the user to pick the desired App for a given intent. Alternatively, the intent issuing app may proactively handle resolution by calling [`findIntent`](ref/DesktopAgent#findintent) or [`findIntentByContext`](ref/DesktopAgent#findintentbycontext) and then raising the Intent with a specific target application, e.g.:
```

src/api/DesktopAgent.ts
```diff
@@ -122,54 +154,87 @@ export interface DesktopAgent {
...
+   * Alternatively, the specific app to target can also be provided, in which case the resolver SHOULD only offer intents supported by the specified application.
    *
    * Using `raiseIntentForContext` is similar to calling `findIntentsByContext`, and then raising an intent against one of the returned apps, except in this case the desktop agent has the opportunity to provide the user with a richer selection interface where they can choose both the intent and target app.
    *
-   * Returns an `IntentResolution` object with a handle to the app that responded to the selected intent.
+   * Returns an `IntentResolution` object with details of the app that was selected to respond to the intent. If the application that resolves the intent returns a promise of Context data, this may be retrieved via the `getResult()` function of the IntentResolution object. If an error is thrown by the handler function, the promise returned is rejected, or no promise is returned then the Desktop Agent MUST reject the promise returned by the `getResult()` function of the `IntentResolution` with a string from the `DataError` enumeration.
```
thorsent, line 199: DRY docs again.

thorsent (approved):
Thanks for all the work on this. I suggested a few rewordings. It's probably worth several passes to get as close to perfect as possible.

docs/api/ref/DesktopAgent.md
```diff
@@ -150,45 +163,79 @@ const appIntent = await fdc3.findIntent("StartChat");
 await fdc3.raiseIntent(appIntent.intent.name, context, appIntent.apps[0].name);
 ```
 
+An optional input context object and/or output context type may be specified, which the resolver MUST use to filter the returned applications such that each supports the specified input and output types.
+```js
+const appIntent = await fdc3.findIntent("StartChat", contact);
```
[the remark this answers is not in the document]
kriswest: Nope, find intent supports an optional second argument: a context example that will be used to narrow down the apps offered to resolve the intent.

docs/api/ref/DesktopAgent.md
```diff
@@ -26,11 +26,11 @@ interface DesktopAgent {
   addContextListener(handler: ContextHandler): Listener;
 
   // intents
-  findIntent(intent: string, context?: Context): Promise<AppIntent>;
-  findIntentsByContext(context: Context): Promise<Array<AppIntent>>;
+  findIntent(intent: string, context?: Context, outputContextType?: string): Promise<AppIntent>;
```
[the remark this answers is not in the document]
kriswest: Good point - will update throughout

docs/api/ref/DesktopAgent.md
```diff
@@ -71,27 +71,39 @@ const contactListener = fdc3.addContextListener('fdc3.contact', contact => { ...
...
 
 
 ### `addIntentListener`
 
 ```ts
-addIntentListener(intent: string, handler: ContextHandler): Listener;
+addIntentListener(intent: string, handler: IntentHandler): Listener;
 ```
- Adds a listener for incoming Intents from the Agent.
+ Adds a listener for incoming Intents from the Agent. The handler function may return void or a promise that should resolve to a context object representing any data that should be returned to app that raised the intent. If an error is thrown by the handler function, the promise returned is rejected, or a promise is not returned then the Desktop Agent MUST reject the promise returned by the `getResult()` function of the `IntentResolution`.
```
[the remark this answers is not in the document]
kriswest: Updated

docs/api/ref/DesktopAgent.md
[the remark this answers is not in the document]
kriswest: Updated

docs/api/ref/DesktopAgent.md
kriswest: fair - done.

docs/api/ref/Errors.md
kriswest: At present, an app resolving an intent has no means to identify the app raising the intent and couldn't assess authorization. Hence, any such mechanism would have to be a function of the Desktop Agent. However, I take the point on returning some info about why the promise was rejected (for feedback to the end-user - 'IntentHandlerRejected' is not all that informative and we are deliberately losing any info on why the rejection occurred).

I don't particularly like the idea of using an 'ErrorContext' type (I think that overloads the purpose of context). However, we could use the javascript Error type (or similar that we define), setting the message to these values and setting the optional `cause` property to the error/rejection message returned by the intent handler...  However, its a departure from all other rejects, which reject with just a string.

docs/api/spec.md
kriswest: done

docs/api/spec.md
kriswest: update

website/static/schemas/next/app-directory.yaml
```diff
@@ -374,8 +380,12 @@ components:
           items:
             type: string
           description: >-
-            A comma sepaarted list of the types of contexts the intent offered by the application can process. 
+            A comma separated list of the types of contexts the intent offered by the application can process. 
             where the first part of the context type is the namespace e.g."fdc3.contact, org.symphony.contact"
+        outputContext:
```
kriswest, line 385:
```suggestion
        resultContext:
```

docs/api/ref/DesktopAgent.md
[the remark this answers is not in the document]
kriswest: oh wait... will change to `context`

kriswest:
Closing in favour of #508 which contains all the same commits

PR #509, opened 2021-11-22
450 allow intents to target app instances

### LINKED ISSUE #450: Allow intents to be targeted at specific instances of apps
### Use Case:
[`fdc3.raiseIntent`](https://fdc3.finos.org/docs/api/ref/DesktopAgent#raiseintent) and [`fdc3.raiseIntentForContext`](https://fdc3.finos.org/docs/api/ref/DesktopAgent#raiseintentforcontext) support an optional `app?` argument that allow you to bypass an intent resolver UI when raising an intent and instead select a specific app type, which is further enabled via the [`fdc3.findIntent`](https://fdc3.finos.org/docs/api/ref/DesktopAgent#findintent) and [`fdc3.findIntentForContext`](https://fdc3.finos.org/docs/api/ref/DesktopAgent#findintentforcontext) functions which allow programmatic access to the resolver data. 

However, in a typical FDC3-enabled desktop there may be one or more instances of an application already open, which the user will often wish to deliver their intent and context to. This might be to any open instance or it might be to a specific instance of the application that has already interacted with the request in some way. There is, at present, no programmatic way to achieve this (although some desktop agents have implemented this behaviour via their resolver UIs) and no way to retrieve details of open application instances through the FDC3 APIs. 

For example: 
> A user viewing details of a trade may wish to raise an intent to `ViewOrder` and target an open instance of a particular order management application.
> A user viewing details of an order in an OMS may wish to open the customer's details in a CRM application (e.g. by raising the intent `ViewContact`). The CRM application receiving the intent may wish to use details of the application and instance that raised the intent to allow the user to `ViewOrders` in the instance of the OMS that contacted it.
> A user may wish to plot a pricing chart by raising a `ViewChart` intent, then plot comparisons with other stocks by repeatedly raising `ViewChart` (or another suitable intent, e.g. `ViewComparison`) and targeting the same app instance as resolved the initial intent.

### Additional information
There are a number of approaches that could be used to support AppInstances, however they will have to be made carefully in order to preserve backwards compatibility. Consideration should also be given to whether support for targeting specific instances of apps is required for FDC3 spec compliance or not.

kriswest:
resolves #450 
Adds an optional `instanceId` field to `AppMetadata`, allowing it to represent specific instances of applications. 

Also updates `open()` to return `AppMetadata` (identifying the instance opened). `raiseIntent()` already returns details of the app resolving an intent, however, this should now always be an `AppMetadata`, with the optional `AppMetadata.instanceId` field set, to represent the app instance that received the intent (and may have been started to do so).


thorsent (approved):
Great

src/api/AppMetadata.ts
```diff
@@ -6,7 +6,13 @@
 import { Icon } from './Icon';
 
 /**
- * App definition as provided by the application directory
+ * Describes an application, or instance of an application, using metadata that is usually  provided by an FDC3 App Directory that the desktop agent connects to.
+ *
+ * Will always includes at least a `name` property, which can be used with [`open`](DesktopAgent#open) and [`raiseIntent`](DesktopAgent#raiseIntent). If the `instanceId` field is set then the `AppMetadata` object represents a specific instance of the application that may be addressed using that Id.
```
mattjamieson, line 11:
```suggestion
 * Will always include at least a `name` property, which can be used with [`open`](DesktopAgent#open) and [`raiseIntent`](DesktopAgent#raiseIntent). If the `instanceId` field is set then the `AppMetadata` object represents a specific instance of the application that may be addressed using that Id.
```

src/api/DesktopAgent.ts
```diff
@@ -25,28 +26,34 @@ export interface DesktopAgent {
   /**
    * Launches an app by target, which can be optionally a string like a name, or an AppMetadata object.
    *
-   * If a Context object is passed in, this object will be provided to the opened application via a contextListener.
-   * The Context argument is functionally equivalent to opening the target app with no context and broadcasting the context directly to it.
+   * The `open` method differs in use from `raiseIntent`.  Generally, it should be used when the target application is known but there is no specific intent.  For example, if an application is querying the App Directory, `open` would be used to open an app returned in the search results.
+   *
+   * **Note**,if the intent, context and target app name are all known, it is recommended to instead use `raiseIntent` with the `target` argument.
```
mattjamieson, line 31:
```suggestion
   * **Note**, if the intent, context and target app name are all known, it is recommended to instead use `raiseIntent` with the `target` argument.
```

mattjamieson (changes requested):

mattjamieson (approved):

rikoe (approved):

ggeorgievx (approved):

PR #604, opened 2022-02-23
Updating Reference to the current IP policy

mindthegab:
The OLD (obsolete, pre LF merger) IP policy is referenced here, updating to the latest and greatest


kriswest (approved):
I've found and added a bunch of other references to the IP policy to this PR. After the additions, LGTM.

mindthegab (commented):
Thanks for catching all the other entries - looks great!

mattjamieson (approved):
LGTM!

PR #615, opened 2022-03-07
FDC3 Compliance Check - First Iteration

dbutt-scottlogic:
A simple first iteration of compliance tests that verifies the existence of desktop API methods.
- Checks for FDC3/FDC3Ready event. Displays error if not found in 5 seconds.
- Launches a mocha test suite that runs automatically
- For each API method, calls the method. If no unexpected error, test succeeds.
- See README for instructions


toolbox/fdc3-compliance/src/FDC3Conformance.ts
```diff
@@ -0,0 +1,22 @@
...
+
+if ((window as any).fdc3) {
+  initialise();
+} else {
+  window.addEventListener("fdc3Ready", initialise);
+
+  setTimeout(() => {
+    if (!fdc3Available) {
+      const hiddenMessage = document.getElementById("no-fdc3");
+      hiddenMessage?.removeAttribute("hidden");
```
ggeorgievx, line 19: Shouldn't we also remove the listener from line 14?

toolbox/fdc3-compliance/src/test/fdc3.addContextListener.ts
```diff
@@ -0,0 +1,7 @@
+describe("fdc3.addContextListener", async () => {
+  it("Method is callable", async () => {
+    await window.fdc3.addContextListener("fdc3.contact", (info: any) => {
```
ggeorgievx, line 3: Shouldn't we also add clean up logic (e.g. inside of an `afterEach()`) to prevent memory leaks?

toolbox/fdc3-compliance/src/test/fdc3.addIntentListener.ts
```diff
@@ -0,0 +1,10 @@
+describe("fdc3.addIntentListener", async () => {
+  it("Method is callable", async () => {
+    await window.fdc3.addIntentListener(
```
ggeorgievx, line 3: Shouldn't we also add clean up logic (e.g. inside of an `afterEach()`) to prevent memory leaks?

toolbox/fdc3-compliance/src/test/fdc3.broadcast.ts
```diff
@@ -0,0 +1,8 @@
+describe("fdc3.broadcast", async () => {
+  it("Method is callable", async () => {
+    window.fdc3.broadcast({
```
ggeorgievx, line 3: Shouldn't we also await the method?

toolbox/fdc3-compliance/src/test/fdc3.addContextListener.ts
```diff
@@ -0,0 +1,7 @@
+describe("fdc3.addContextListener", async () => {
```
ggeorgievx, line 1: Do we need the top level async (this comment applies to all test files).

toolbox/fdc3-compliance/src/test/fdc3.getInfo.ts
```diff
@@ -0,0 +1,5 @@
+describe("fdc3.getInfo", async () => {
+  it("Method is callable", async () => {
```
ggeorgievx, line 2: Do we need the async here?

toolbox/fdc3-compliance/src/test/fdc3.joinChannel.ts
```diff
@@ -0,0 +1,5 @@
+describe("fdc3.joinChannel", async () => {
+  it("Method is callable", async () => {
+    await window.fdc3.joinChannel("FDC3Conformance"); // This will hang under the Chrome Extension Desktop Agent
```
ggeorgievx, line 3: Shouldn't we create the channel first? It isn't a good practice to rely on state from other tests.

toolbox/fdc3-compliance/src/test/fdc3.open.ts
```diff
@@ -0,0 +1,13 @@
+import { OpenError } from "@finos/fdc3";
+
+describe("fdc3.open", async () => {
+  it("Method is callable", async () => {
+    try {
+      window.fdc3.open("fdc3-workbench");
```
ggeorgievx, line 6: Missing await.

ggeorgievx (changes requested):

toolbox/fdc3-compliance/src/test/fdc3.broadcast.ts
kriswest: Not in 1.2, broadcast becomes async in 2.0

toolbox/fdc3-compliance/src/test/fdc3.joinChannel.ts
kriswest: Do NOT join app channels. This being clarified in the 2.0 docs along with the rename of system channels to app channels. While it's technically possible it's not intended... As such questions have been asked about whether it should be specifically prevented.

You should always be doing getOrCreateChannel then adding a listener to the returned object for a named app channel.

ggeorgievx: Thank you for pointing this out, Kris! This raises a very good question - how can we test `joinChannel()` when we don't have control over the system channels?

kriswest: oof the 1.2 spec doesn't actually require there to be any system channels in the list apart from `global` (which is deprecated - in 1.2 apps MAY be connected to it by default, but don't have to be - its gone in 2.0). 

However, I've not yet seen a desktop agent without system channels and as global is technically required there should always be at least one channel returned by `fdc3.getSystemChannels()` for the app to join. Hence, do that.

kriswest (changes requested):
Seems a reasonable start on the conformance project. At present, the tests only check if functions are callable and not any of their behavior (which presumably we'll see some of in a future update). I've made some comments about checking responses in these initial checks however as I think you could do some confirmation that the app(s) are set-up correctly for the rest of the tests.

A correction is needed in the `fdc3.joinChannel` test as indicated.

Readme could be expanded to cover more of the set-up for the test - including customisation to the manifest that will be required. Could also eventually include 1.2 manifests for the major containers (vendors to supply) - thankfully 2.0 will support multiple manifests and (if possible) some more vendor agnostic fields making them less necessary.

toolbox/fdc3-compliance/src/test/fdc3.addIntentListener.ts
dbutt-scottlogic: We will leave the await in place, given that it isn't causing problems and it will be async in the near future. 
We will also implement the suggested clean-up logic for listeners.

toolbox/fdc3-compliance/src/test/fdc3.addContextListener.ts
dbutt-scottlogic: It turns out that we don't need the top-level async, so will remove it.

toolbox/fdc3-compliance/src/test/fdc3.joinChannel.ts
dbutt-scottlogic: Thanks for the analysis - we will adopt the suggested use of `fdc3.getSystemChannels()` for this version.

toolbox/fdc3-compliance/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+`yarn start`
+
+4. Add the URL http://localhost:8080 to your FDC3-enabled container or desktop agent and ensure it has access to the `window.fdc3` object.
+
+Alternatively, a full set of static files are available in the `build` folder. Load the index.html file into an environment that has the `window.fdc3` object available.
+
+## Application Definition
+
+A basic FDC3 application definition, as defined in the [application directory specification](https://fdc3.finos.org/schemas/1.2/app-directory#tag/Application), is supplied in the file `appDefinition.json`. This may be useful when adding the conformance tests to an application directory.
```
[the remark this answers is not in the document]
dbutt-scottlogic: We can definitely bulk up the README in this area. For this version, the tests run automatically on load, so we will make that clear. 
As far as configuration goes, we are avoiding putting anything vendor-specific in the documentation, but we will make it clear that there may be additional configuration required (the current file works as-is with the existing inline manifest with Glue42).

toolbox/fdc3-compliance/src/test/fdc3.findIntent.ts
```diff
@@ -0,0 +1,13 @@
+import { ResolveError } from "@finos/fdc3";
+
+describe("fdc3.findIntent", async () => {
+  it("Method is callable", async () => {
+    try {
+      await window.fdc3.findIntent("ViewChart");
+    } catch (ex) {
+      if (ex !== ResolveError.NoAppsFound) {
```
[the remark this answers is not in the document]
dbutt-scottlogic: Agreed - those are all things on the TODO list for the next iteration. For this version, anything other than the expected responses will cause a test failure.

toolbox/fdc3-compliance/src/test/fdc3.addContextListener.ts
```diff
@@ -0,0 +1,21 @@
+import { Listener } from "@finos/fdc3";
+
+describe("fdc3.addContextListener", () => {
+  let listener: Listener;
+
+  afterEach(async () => {
+    listener.unsubscribe();
```
ggeorgievx, line 7: We could get rid of the async, add a check for whether the listener's type is !== undefined and finally unset the listener after unsubscribing.

toolbox/fdc3-compliance/src/test/fdc3.addIntentListener.ts
```diff
@@ -0,0 +1,18 @@
+import { Listener } from "@finos/fdc3";
+
+describe("fdc3.addIntentListener", () => {
+  let listener: Listener;
+
+  afterEach(async () => {
+    listener.unsubscribe();
```
ggeorgievx, line 7: Same as with the context listener.

toolbox/fdc3-compliance/src/test/fdc3.getInfo.ts
```diff
@@ -0,0 +1,5 @@
+describe("fdc3.getInfo", () => {
+  it("Method is callable", async () => {
+    window.fdc3.getInfo();
```
ggeorgievx, line 3: We could add an await, as it won't break anything and won't require a revisit for 2.0.

toolbox/fdc3-compliance/src/test/fdc3.joinChannel.ts
```diff
@@ -0,0 +1,15 @@
+describe("fdc3.joinChannel", () => {
+  afterEach(async () => {
+    window.fdc3.leaveCurrentChannel();
```
ggeorgievx, line 3: Missing await.

toolbox/fdc3-compliance/src/test/fdc3.findIntent.ts
```diff
@@ -0,0 +1,13 @@
+import { ResolveError } from "@finos/fdc3";
+
+describe("fdc3.findIntent", () => {
+  it("Method is callable", async () => {
+    try {
+      await window.fdc3.findIntent("ThisIntentDoesNotExist");
```
ggeorgievx, line 6: We expect an error to be thrown so we should throw an error immediately after line 6 to make sure the test fails if no error is thrown by `findIntent()`. This would also require a rework inside of the catch statement.

toolbox/fdc3-compliance/src/test/fdc3.findIntentsByContext.ts
```diff
@@ -0,0 +1,24 @@
...
+      id: {
+        ticker: "ticker",
+        ISIN: "US0378331005",
+        CUSIP: "037833100",
+        FIGI: "BBG000B9XRY4",
+      },
+    };
+
+    try {
+      await window.fdc3.findIntentsByContext(context);
```
ggeorgievx, line 17: Same as find intent.

toolbox/fdc3-compliance/src/test/fdc3.open.ts
```diff
@@ -0,0 +1,13 @@
+import { OpenError } from "@finos/fdc3";
+
+describe("fdc3.open", () => {
+  it("Method is callable", async () => {
+    try {
+      await window.fdc3.open("ThisAppDoesNotExist");
```
ggeorgievx, line 6: Same as find intent.

toolbox/fdc3-compliance/src/test/fdc3.raiseIntent.ts
```diff
@@ -0,0 +1,24 @@
...
+      id: {
+        ticker: "aapl",
+        ISIN: "US0378331005",
+        CUSIP: "037833100",
+        FIGI: "BBG000B9XRY4",
+      },
+    };
+
+    try {
+      await window.fdc3.raiseIntent("ViewChart", context);
```
ggeorgievx, line 17: Same as find intent.

ggeorgievx (changes requested):
Thank you for the revisit, @dbutt-scottlogic ! We are almost there.

ggeorgievx (approved):

kriswest (changes requested):
First off, apologies for the delay (big product release and a vacation got in the way 😬 ). As penance I've tried to add full suggestions you can commit for proposed changes.

Overall it LGTM, however, I spotted one possible issue with error handling. Technically, the FDC3 spec says that `Error` objects should be thrown with the strings from the error enumerations, where the tests are assuming the strings themselves will be thrown (its annoying that javascript allows this IMHO, it perhaps should have upgraded strings to errors). E.g.
>If the resolution fails, the promise will return an `Error` with a string from the [ResolveError](https://fdc3.finos.org/docs/api/ref/Errors#resolveerror) enumeration.
Finsemble appears to be throwing the strings, rather than Errors, which may have steered you wrong on this... The suggestions I've put in would handle it either way - however we should probably have a discussion on this (in the Standards Working Group) as it makes error handling more complex.

Also the test framework is not happy with strings being thrown on failures (Finsemble is throwing the wrong error string for `fdc3.open`), resulting in:
image

toolbox/fdc3-compliance/src/appDefinition.json
```diff
@@ -0,0 +1,31 @@
...
+    "description": "FDC3 Compliance Test",
+    "images": [
+      {
+        "url": "https://fdc3.finos.org/img/fdc3-logo-2019.png"
+      }
+    ],
+    "publisher": "FINOS",
+    "icons": [
+      {
+        "icon": "https://fdc3.finos.org/img/fdc3-logo-2019.png"
```
kriswest, line 18: This logo doesn't work well as an icon, i'd suggest swapping out for http://fdc3.finos.org/toolbox/fdc3-workbench/fdc3-icon-256.png
(which you can copy into this project)

ColinEberhardt:
@kriswest I've added @AidanZealley to the CLA, can you please re-run the checks?

kriswest:
@ColinEberhardt submitted a LF support request as I don't see the checks being re-run, pretty sure I'm doing that correctly

kriswest:
@ColinEberhardt support says:

> I checked EasyCLA and there is no signature on file for https://github.com/AidanZealley ID: 2430106, Aidan needs to go to "Please click here to be authorized" link in the PR comment https://github.com/finos/FDC3/pull/615#issuecomment-1113135080 and complete the signature and acknowledgment process for EasyCLA. Once done EasyCLA should authorize the commits https://github.com/finos/FDC3/commit/c0150382a56a891fc5aad32f1591d6d36e76db2d and https://github.com/finos/FDC3/commit/b1c63e67c0f7c0b02135e4c2dc95fd70542a3b06.

When you've got that done, comment `/easycla` to re-run the check.

kriswest:
Closing as this project has moved to it own repo and will merge back a finalized build instead

Issue #600, opened 2022-02-21
Intent Proposal: View Profile


pauldyson:
### Use Case
A Corporate Access administrator is researching a potential invite list for an upcoming IPO. To do this they look at data held in multiple systems such as their Order Management System, their Bookbuilding System, etc in order to determine which of the organisations who have previously participated in offerings may be likely to participate in the IPO they are prepping. Once they have determined which organisations to target, they will need to figure out which individuals at that organisation to approach. This information will be held in the CRM.

A salesperson is about to initiate a chat with a Buy Side fund manager on IM and just wants to check their details; job title, investment style, before they pick up the conversation. This information will be held in the CRM

#### ViewProfile
ViewProfile would allow a quick navigation to an individual's or a company’s 'basic information'. ViewProfile would be called with the Contact or Organization context.

### Additional Information
ViewProfile with the Contact context is essentially the same as the current ViewContact intent. This could be deprecated over time but, for the time being, they should be considered equivalent.

This issue was created after feedback on a proposal for ViewOrganization to make it more general https://github.com/finos/FDC3/issues/586

PR #619, opened 2022-03-08
600 ViewProfile addition and ViewContact deprecation

pauldyson:
resolves: #600

My first go at this so feedback welcome!

ViewProfile added with various links from sidebars, the Contact and Organisation Context, etc.

ViewContact amended with a deprecation notice and removes from sidebar.

Updated spec.md to summarise all the standard intents and move ViewContact into a Deprecated section.

I haven't removed ViewContact from intents.ts or standard intents.js as it's not clear to me where they're used downstream


kriswest:
> I haven't removed ViewContact from intents.ts or standard intents.js as it's not clear to me where they're used downstream

It shouldn't be removed from there until ViewContact is removed from the standard in a later version. Hence, this is fine.

docs/intents/ref/ViewContact.md
```diff
@@ -4,6 +4,10 @@ sidebar_label: ViewContact
 title: ViewContact
 hide_title: true
 ---
+
+> **Note:** ViewContact has been deprecated in FDC 2.0 in favour of the more general [ViewProfile](ViewProfile)
+
+
 # `ViewContact`
```
kriswest, line 11:
```suggestion
# `ViewContact`
> **Note:** ViewContact has been deprecated in FDC 2.0 in favour of the more general [ViewProfile](ViewProfile)

```

kriswest (approved):
LGTM, added a few minor suggestions

greyseer256 (approved):

mistryvinay (commented):
Reviewing the use case detailed in #600 ,

"A salesperson is about to initiate a chat with a Buy Side fund manager on IM and just wants to check their details; job title, investment style, before they pick up the conversation. "

But ViewProfile is only returning name and emailAddress. Do we need to add the persons jobTitle?

docs/intents/ref/ViewContact.md
```diff
@@ -5,6 +5,8 @@ title: ViewContact
 hide_title: true
 ---
 # `ViewContact`
+> **Note:** ViewContact has been deprecated in FDC 2.0 in favour of the more general [ViewProfile](ViewProfile)
```
kriswest, line 8:
```suggestion
> **Note:** ViewContact has been deprecated in FDC3 2.0 in favour of the more general [ViewProfile](ViewProfile)
```

mistryvinay:
> > But ViewProfile is only returning name and emailAddress. Do we need to add the persons jobTitle?
> 
> @mistryvinay As a 'View' intent (see discussion from the Context & Intents meeting when minutes are posted #627) its not actually expected to return data, but rather to start/use an app to display the data. The context object is actually an input that will be used to find the contact record to display. That said, further fields could be added to the context where relevant for retrieval - e.g. a IM username. However, that could be the subject of a separate PR (as the context can be expanded separately). If this has relevance to Symphony that could be an issue/PR you look to raise @mistryvinay.
> 
> We should expect to add a `GetProfile` intent in future that would actually return a context (either Contact or Organization), at which point it would indeed be good to expand those records to have more data fields in (than we need to search for/identify the contact)...

Thanks for the clarification @kriswest that makes sense, using this context object as an input for retrieving information. 👍

mistryvinay (approved):
Looks good to me.

PR #620, opened 2022-03-08
574 StartChat: ChatInitSettings and ChatRoomRef

bertrand-s:
I made several simplifications to the initial issue (574)
- ChatInit was renamed ChatInitSettings
- ChatInitSettings options are not chat-provider specific anymore
- ChatInitSettings options support a new option (*groupRecipients*) that allows to create one room per recipients or one room with all members - this is why the proposal is to have StartChat return a list of ChatRoomRef
- ChatRoomRef app is now an AppMetadata
- ChatRoomRef id is a simple string


docs/context/ref/ChatInitSettings.md
```diff
@@ -0,0 +1,85 @@
...
+
+https://fdc3.finos.org/schemas/next/chatInitSettings.schema.json
+
+## Details
+
+| Property                       | Type      | Required | Example Value                                                        |
+| ------------------------------ | --------- | -------- | -------------------------------------------------------------------- |
+| `type`                         | string    | Yes      | `'fdc3.chat.initSettings'`                                           |
+| `chatName`                     | string    | No       | `'Instrumet XYZ'`                                                    |
+| `members`                      | Contact[] | No       | `[contact1, contact2]`                                               |
```
ggeorgievx, line 26: Maybe we could use [`ContactList`](https://fdc3.finos.org/docs/context/ref/ContactList) instead of `Contact[]`.

ggeorgievx (approved):

docs/context/ref/ChatInitSettings.md
bertrand-s: Yes - I though of this option, but I realised that ContactList adds extra verbosity (type, sub-structure, etc.) this is why I preferred the Contact[] option

bertrand-s: OK - I will update the PR with Contact and ContactList then

docs/intents/ref/StartChat.md
```diff
@@ -33,13 +37,42 @@ const contact = {
 }
 
 fdc3.raiseIntent('StartChat', contact)
+
+// chat with initialization settigns
```
greyseer256, line 41:
```suggestion
// chat with initialization settings
```

greyseer256 (approved):

mistryvinay (approved):
Looks good to me.

mistryvinay:
@finos/fdc3-maintainers any blockers for not approving?

docs/context/ref/ChatInitSettings.md
```diff
@@ -0,0 +1,86 @@
...
+## Schema
+
+https://fdc3.finos.org/schemas/next/chatInitSettings.schema.json
+
+## Details
+
+| Property                       | Type        | Required | Example Value                                                        |
+| ------------------------------ | ----------- | -------- | -------------------------------------------------------------------- |
+| `type`                         | string      | Yes      | `'fdc3.chat.initSettings'`                                           |
+| `chatName`                     | string      | No       | `'Instrumet XYZ'`                                                    |
```
kriswest, line 25:
```suggestion
| `chatName`                     | string      | No       | `'Instrument XYZ'`                                                    |
```

docs/context/spec.md
```diff
@@ -80,6 +80,13 @@ The identifier "foo" is proprietary, an application that can use it is free to d
 The following are standard FDC3 context types.
  __Note:__ The specification for these types are shared with the [FINOS Financial Objects](https://fo.finos.org) definitions, JSON schemas are hosted with FDC3.
```
kriswest, line 81: Note for @greyseer256: this sentence is no longer correct once we merge in new types - and that group has changed focus/no longer maintains those definitions. Hence, its probably time to drop this

kriswest (approved):
LGTM

kriswest:
Needs a CHANGELOG.mb entry adding!

docs/context/ref/ChatInitSettings.md
```diff
@@ -0,0 +1,86 @@
+---
+id: ChatInitSettings
```
robmoffat, line 2: Suggest calling this ChatSettings?   It seems like these settings would be relevant even after the chat had been created, and could be the basis for further, ongoing intents.  e.g. EmailTranscript, CreateAppointment, RecordInCRM, etc. which could be handled by context listeners.

kriswest:
As discussed in the meetings, it is expected that a ChatRoomRef type will
be added later and returned by intents that create a chat and it's that
type that would be used for those purposes @robmoffat.

I think there's more on this in meeting minutes.

robmoffat:
Yes, that's very symphony-specific: there, rooms have names, chats are between a group of people.   

I mean, there's an argument that there should be a single context object covering both but I expect somewhere you've already had that discussion before...

But, based on the idea that this should be `context` data, I still think we should follow the naming scheme we already have for other bits of context: `Instrument`, `Organisation`, `Contact`.

PR #633, opened 2022-03-23
544 desktop agent bridging proposal

kriswest:
Migrating proposal branch to FINOS repo to ease collaboration


Soul-Master:
@kriswest 
Do you plan to add spec for desktop agent client authorization? 

I think it should be better for end-user to show some confirmation window about desktop agent client connection. It should allow user to reject connect if they don't know/trust desktop agent client.

kriswest:
@Soul-Master good suggestion - I actually like that idea better than the configuration for an authenticated connection (although not mutually exclusive). 

Please do feel free to either add to the discussion #544 or the PR I've just opened to help collect reviews/suggestions of this proposal #634 (this PR was just move the proposal doc from our repo to the FINOS one to enable collab on it -'ll lock comments so we focus on the other PR).

PR #634, opened 2022-03-23
[WIP] 544 desktop agent bridging proposal

### LINKED ISSUE #453: Support the production of Desktop Agent Bridging
## Enhancement Request
Extend the FDC3 API's to improve support for producing bridges between FDC3 implementations aka Desktop Agents.

In this context an FDC3 Bridge would allow an application running on one Desktop Agent, to be able to integrate with FDC3 applications running on a second Desktop Agent for the same user.

The interop between applications running on different Desktop Agents aka Platforms would ideally cover 
- Channels
- Intents
- Application Launching
  
### Use Case:
1. A Client has two Desktop Agent environments running on their desktop and wants to be able to raise an intent from an application running on Desktop Agent 1 (say Glue42) and the intent is implemented on Desktop Agent2 (say OpenFin).

2. A similar use case occurs `round User Selected channels, where a user wants to see applications running on the 'same channel' from the multiple desktop agents, synchronise on a common symbol.
NB This assumes that from a user PoV a red channel on Platform is the same as a red channel on Platform 2. 

3. An existing platform, say a Terminal offers interop between its applications using its own (non FDC3) API's. It wants to allow applications running on other Platforms to interwork with apps running on the Terminal. This could be acheived by delivering a Bridge for whatever Platforms a user may be running. 

### Additional Information
If this use case is accepted we can imagine two implementations, either adding a few extra API methods (for which change requests will be added) or by defining a Desktop Agent to Agent interop bus/api. But our view of the agenet to agent API is that it is too big a change.

kriswest:
Work-In-Progress PR for the desktop agent bridging proposal - raised to enable review and collaboration.

resolve #453


docs/api-bridging/spec.md
```diff
@@ -0,0 +1,1193 @@
...
+            "desktopAgent": "agent-C",
+        }
+    }
+}
+```
+
+Then on agent-A the originating app finally gets back the following response from the bridge:
+
+```JSON
+// DAB -> agent-A
```
timjenkel, line 552: Is this `DAB -> agent-A` or `agent-A -> app` ?

Does the DAB wait for responses from agents B and C then send the combined response to agent A, or does the DAB send the responses from B and C separately and agent A has to wait for all responses then combine?

docs/api-bridging/spec.md
```diff
@@ -0,0 +1,1193 @@
...
+  * check if any of your clients is the target and transmit the call to them and await a response
+  * If you are also a client of another server follow the client steps
+  * otherwise return `OpenError.AppNotFound`
+* If you are a client
+  * transmit the call to the server and await a response
+
+The second case is a little trickier as we don't know which agent may have the app available:
+
+* If the local Desktop Agent has the app, open it and exit.
+* If you are a server
```
timjenkel, line 1078: This will need to be reworded if we are using the bridge instead of client/server approach

docs/api-bridging/spec.md
```diff
@@ -0,0 +1,1410 @@
...
+    * already includes supported FDC3 version.
+  * request for a specific agent name.
+
+```mermaid
+sequenceDiagram
+    participant DA as Desktop Agent A
+    participant DAB as Desktop Agent Bridge
+    participant DB as Desktop Agent B
+    participant DC as Desktop Agent C
+    DAB ->>+ DA: hello
```
Yannick-Symphony, line 134: should we add the initial websocket connection to the flow diagram? currently it looks like it's the DAB connecting to the desktop agent

Vivek-NatWest:
Channel State Synchronization step need a bit more elaboration for below scenario:
1- DA would async connect with bridge in background without blocking users to use interop in its context. How would this impact multiple desktop agents channel state synchronization.
-   behaviour of `channel.getCurrentContext` in case both desktop agents have channel state for same context type but with different value for example channel1 in both have fdc3.instrument but one is for apple and other for msft.

kriswest:
@Vivek-NatWest 

> * behaviour of `channel.getCurrentContext` in case both desktop agents have channel state for same context type but with different value for example channel1 in both have fdc3.instrument but one is for apple and other for msft.

I believe this is covered in the last two paragraphs of Step 6 of the connection protocol (connectedAgentsUpdate message):

> Handling by the Desktop Agent of these synchronization messages from the DAB should be atomic to prevent message overlap with `fdc3.broadcast`, `channel.broadcast`, `fdc3.addContextListener` or `channel.getCurrentContext`. I.e. the `connectedAgentsUpdate` message must be processed immediately on receipt by Desktop Agents and updates applied before any other messages are sent or responses processed.
>
> Similarly, the DAB must process `handshake` messages and issue `connectedAgentsUpdate` messages to all participants (steps 3-6) atomically, allowing no overlap with the processing of other messages from connected agents.

(n.b. I slightly improved some clumsy copy in the first paragraph before replying).

If that doesn't resolve this for you let me know and we can discuss a further change to clarify.

Vivek-NatWest:
> @Vivek-NatWest
> 
> > * behaviour of `channel.getCurrentContext` in case both desktop agents have channel state for same context type but with different value for example channel1 in both have fdc3.instrument but one is for apple and other for msft.
> 
> I believe this is covered in the last two paragraphs of Step 6 of the connection protocol (connectedAgentsUpdate message):
> 
> > Handling by the Desktop Agent of these synchronization messages from the DAB should be atomic to prevent message overlap with `fdc3.broadcast`, `channel.broadcast`, `fdc3.addContextListener` or `channel.getCurrentContext`. I.e. the `connectedAgentsUpdate` message must be processed immediately on receipt by Desktop Agents and updates applied before any other messages are sent or responses processed.
> > Similarly, the DAB must process `handshake` messages and issue `connectedAgentsUpdate` messages to all participants (steps 3-6) atomically, allowing no overlap with the processing of other messages from connected agents.
> 
> (n.b. I slightly improved some clumsy copy in the first paragraph before replying).
> 
> If that doesn't resolve this for you let me know and we can discuss a further change to clarify.

Thanks @kriswest for replying.

Does this mean `fdc3Ready `event would not be fired by desktop agent until its connected to bridge? What would happen if there is no bridge running? 

If there is no operations allowed till DA connects to bridge, would channels state contain Context JObjects (think not since channel has no history until any context broadcast happens over it?). 

Also, would DAB update its channels state if it receives a channel not previously seen during broadcast? This state would be posted only on new DA connection/ disconnection?

kriswest:
> Does this mean `fdc3Ready `event would not be fired by desktop agent until its connected to bridge? What would happen if there is no bridge running?

No, the bridging specification as currently written is not intended to have DAs configured to require connection. Rather, they should be able to connect to a bridge as they start up or after they are running - which will make disconnect/reconnect events (e.g. because the bridge crashed) easier to handle. Once they are connected, they need to take the bridge into account during FDC3 API calls. 

> If there is no operations allowed till DA connects to bridge, would channels state contain Context JObjects (think not since channel has no history until any context broadcast happens over it?).

I think we have to assume that channels _may_ contain state when a DA connects. This could be down to a number of reasons, for example:
- The bridge crashed and was restarted, the DA has remained running
- State is being stored in a workspace or similar and restored by the DA
- Bridge instances on another machine are added to your cluster and bring state with them (resolving that would be an internal concern of the bridge implementation)

> Also, would DAB update its channels state if it receives a channel not previously seen during broadcast? This state would be posted only on new DA connection/ disconnection?

It should yes. For both app channels and user channels it might be possible for connecting DA to have state on those channels (and channels themselves) that have not yet been seen. It should take on their state, merge it into the current set and then redistribute that out to other agents via the `connectedAgentsUpdate` message. 

I think this also applies to your multi-machine implementation when a node is added to the cluster - you might bring channel state with you that has to be merged, along with additional agents...

I think the proposed sync procedure is workable and (probably) necessary. However, I'm not currently implementing this in a bridge as you are. Hence, if you foresee problems or believe a simplified procedure or simplifying assumptions are warranted then we should discuss that and make changes. If its a major change then probably the best way to kick that off is to raise an issue suggesting a change, which we'll discuss at the next meeting and have participants vote on.

kriswest:
Closing PR as this branch has an accidental reversion of some 2.0 commits. Will reopen on a clean branch

kriswest:
Recreated as PR:
- #860

PR #638, opened 2022-03-25
Conformance app

AidanZealley:


kriswest:
Although this PR is closed, do be aware that it had changes to the top-level package.json and yarn.lock, ideally changes should be confined to /toolbox/fdc3-conformance (was probably just an accidentally yarn command in the wrong folder)

PR #640, opened 2022-03-25
639 remove unachievable compliance requirement

### LINKED ISSUE #639: Remove or adjust Platform Provider compliance requirement to implement a workflow
### Area of Issue
[x] API
[x] Other

FDC3's Compliance page contains a compliance requirement for Platform Providers that is problematic:
> At least one workflow of one of the use cases marked as Accepted by the Use Cases working group SHOULD be satisfied by the implementation.

Platform Providers are expected to enable workflows that implement use-cases (by providing interop functionality as specified in the API specification), but do not provide the apps that handle the individual steps of the workflow. Hence, this requirement is impossible to satisfy without also being an Application Provider.

Either remove the requirement or replace it with a more accurate and achievable requirement. 

###Proposed solution
I would recommend removal as the design of the API standard should be such that it enables (or at least aspires to enable) the implementation of *all* the workflows/use-cases. The requirement should also not be moved to the Application Provider section as a single application is unlikely to be able to fulfill an entire workflow and there is no guarantee that all possible use-cases/workflows were recorded.

kriswest:
resolves #639 by removing the unachievable compliance requirement

### DIFF
--- a/docs/fdc3-compliance.md
+++ b/docs/fdc3-compliance.md
@@ -34,7 +34,6 @@ For platform providers FDC3 compliance requires:
     * Support for intent and context resolution using a resolver UI
     * Support for retrieving information about the version of the FDC3 specification supported by a Desktop Agent implementation and the name of the implementation provider
 * In the case of web applications, a Desktop Agent MUST provide the FDC3 API via a global accessible as `window.fdc3`.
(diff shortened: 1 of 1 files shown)


kriswest:
See issue for description - not brought up at SWG yet, but hopefully not contentious and easy to approve at the April meeting

greyseer256 (approved):

mattjamieson (approved):
:)

PR #642, opened 2022-03-28
fixing a number of broken 'see also' links in the API docs

kriswest:
There are a number of bad see also links in teh 1.2 and 2.0 docs, which this PR corrects. 

Note: I didn't check further back, but I think most stem from the consolidation of 'Metadata' and 'Types' under articles of that name hence 1.1 is less likely to be affected.


greyseer256 (approved):

rikoe (approved):
lgtm

PR #660, opened 2022-04-06
Update Glue42 Logo for FDC3 Website

Julia-Ritter:


kriswest (changes requested):
Please also remove the existing image(s) at: 
- _website/static/img/users/GLUE42.png_
- _website/static/img/users/GLUE42.webp_

Issue #491, opened 2021-11-03
Reorganizing the Specification and Incorporating Reference Material


kriswest:
## Formalizing the FDC3 Standard documentation

This issue has been raised as part of work, sponsored by FINOS, to improve and formalize the documentation of the FDC3 standard.

##  Issue description

Historically, only the 4 'specification' documents have been considered to be the official parts of the FDC3 standard (as defined here https://fdc3.finos.org/docs/fdc3-intro#specifications), that are voted on by FDC3 participants, with everything else considered to be supporting documentation only. However, the 'specification' documents do not provide a complete definition of FDC3. 

For example, the specification for the APIs (https://fdc3.finos.org/docs/api/spec) does not include a full set of definitions for the API calls, their signatures and expected behaviour. However, this information is provided in the API reference documents (e.g. https://fdc3.finos.org/docs/api/ref/DesktopAgent) and Typescript source files (these are at least weakly referred to in the API specification).  

Each section also includes an overview document, which provides a brief introduction to each part of the standard, including information that is generally repeated in more detail in the current specification documents. However, the content provides a useful introduction, which could be moved forward to the introduction section.

Hence, it is recommended that: 
- The current 'specification' documents are renamed to remove 'specification' from their names, to ensure that other documents also form an official part of the standard
- The navigation of the standard website is updated to: 
  - Clearly differentiate introductory or 'getting started' content from the formal parts of the 'Standard' 
  - Ensure other documents are clearly part of the standard
- The 'overview' documents are moved to the 'getting started' section, which sits outside the 'Standard' documentation.
  - Duplication between the overview documents and 'Standard' documentation is expected.
- The reference documents are formally included as part of the Standard
  - Any duplication between the reference documents and 'specification' documents should be removed (particularly in the Context and Intents documents)
- Governance documentation is provided, clearly defining the formal part of the standard

Adopting these recommendations is closely related to implementation of the other recommendations in issues: 
- #482 
- #483 
- #484 

### Navigation and Structure

**Proposal(s)**
Refactor the navigation structure as given below:

- **Getting Started**
  - Introduction <-- Introduction to the four parts of the specification + use-cases intro
  - Why FDC3?
  - FDC3 Charter
- **Use-cases**
  - Equity sellside trader
  - Buy side Portfolio manager
  - …	
- **Implementations** _<-- new section proposed by issue #466_ 
  - Platform Providers
  - Application Providers
  - Tools and Training
  - Custom Types and Intents
[...]

PR #666, opened 2022-04-19
553 Support v1 and v2 URLs in AppD spec

### LINKED ISSUE #553: Add backwards compatibility support for the App Directory
PR #437 (issue #314) introduced a breaking change to the App Directory specification, namely changes to the `Application` schema. This issue suggests adding backwards compatibility to the specification by using the versioned URL end-points:
- Introduce `v2` end-points for all newly added changes
- Add back the old Application schema and sort out any component naming issues
- Mark the `v1` end-points as optional/deprecated

kriswest:
resolves #553 

Moves latest work on appD to /v2/ urls and restores /v1/ urls to 1.2 standard version of spec. 
Will enable an appD implementation to serve both formats (if desired) to simplify migration.

### DIFF
--- a/src/app-directory/README.md
+++ b/src/app-directory/README.md
@@ -12,8 +12,8 @@ The purpose of the Application Directory specification within FDC3 is to:
 
 ## Specification
 
-* [View the OpenAPI Specification in Swagger](https://editor.swagger.io/?url=https://fdc3.finos.org/schemas/1.1/app-directory.yaml)
-* [More about the Specification on the FDC3 Website](https://fdc3.finos.org/docs/1.1/app-directory/overview)
+* [View the OpenAPI Specification in Swagger](https://editor.swagger.io/?url=https://fdc3.finos.org/schemas/1.2/app-directory.yaml)
+* [More about the Specification on the FDC3 Website](https://fdc3.finos.org/docs/1.2/app-directory/overview)
 
 
 ## Directory structure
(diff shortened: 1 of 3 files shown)


kriswest:
@ggeorgievx FYI I'm targetting this PR at a branch so that we can collect up a number of changes to present to the SWG as a consolidated set (rather that 20 different PRs). Other rounds of changes from discussion groups have devolved into a nightmare of conflict resolutions (each merge created conflicts in all the remaining PRs) which I think we need to avoid this time ;-) particularly as I'm about to start on the spec re-org PR (#491) which will conflict with _everything_.

kriswest:
@ggeorgievx I'm going to go ahead and merge this into the consolidated branch as I need to base other PRs on it. Once done I'll raise a new WIP PR for the consolidated branch to enable review and tag you in that

ggeorgievx (commented):
Everything looks great @kriswest 🙌

I will look into the version (1.2 vs next) issue on the AppD part of the website.

kriswest:
Good luck - I thought it might be todo with the fact the appD page wasn't an HTML fragment (see website/pages/schemas/next/app-directory.html) as expected by the [wrapPagesHtml](https://v1.docusaurus.io/docs/en/site-config#wrappageshtml-boolean) option (was producing HTML with two html/head/body tags) but that wasn't it. Hence, I suspect its just a bug in docusaurus... it can be manually fixed in the generated HTML post-deployment (the version number is generated into the HTML not set by a script of anything similar)

Issue #657, opened 2022-04-04
App Directory Discussion group - Monday 4th April 2022


kriswest:
## Group overview
Discussion group to develop proposals for the improvement of the App Directory as part of the FDC3 standard. The group will examine the role/value/necessity of the AppD as part of the FDC3 specification and a compliance requirement for use of the FDC3 API, in addition to working on proposals to improve the description and functionality of the App Directory. 

### Relevant tags
Current open issues and PRs that relate to the above concepts with the label: 
[![image](https://user-images.githubusercontent.com/1701764/151783600-e65898db-0145-45bc-b500-2e3cf7b1010a.png)](https://github.com/finos/FDC3/labels/app-directory)

A [github discussion](https://github.com/finos/FDC3/discussions/618) is also available for this group (may be used to arrange meetings, agenda ask questions etc.)

## Meeting Date
Monday 4th April 2022 - 9am EST / 2pm BST

### WebEx info
- Meeting Link: [[insert link](https://finos.webex.com/finos/j.php?MTID=m2316d45da678d0203fd86f9342ff63ab)](https://finos.webex.com/finos/j.php?MTID=m2316d45da678d0203fd86f9342ff63ab )
- Meeting number: 2551 735 2618

*More ways to join*
- Join by video system:
   - Dial 25517352618@finos.webex.com
   - You can also dial 173.243.2.68 and enter your meeting number
- Join by phone
   - +1-415-655-0003 US Toll
   - +44-20319-88141 UK Toll
- Access code: 2551 735 2618

## Meeting notices
- FINOS **Project leads** are responsible for observing the FINOS guidelines for [running project meetings](https://github.com/finos/community/blob/master/governance/Meeting-Procedures.md#run-the-meeting). Project maintainers can find additional resources in the [FINOS Maintainers Cheatsheet](https://odp.finos.org/docs/finos-maintainers-cheatsheet/).

- **All participants** in FINOS project meetings are subject to the [LF Antitrust Policy](https://www.linuxfoundation.org/antitrust-policy/), the [FINOS Community Code of Conduct](https://github.com/finos/community/blob/master/governance/Code-of-Conduct.md) and all other [FINOS policies](https://github.com/finos/community/tree/master/governance#policies).

[...]

PR #670, opened 2022-04-21
622 AppD lang and translations

### LINKED ISSUE #622: Add language metadata to AppD for accessibility
### Use Case:
* An institution has an app that supports multiple languages. They want to include multiple manifest entries, where they have a different version of the app for each language the app supports.
* Consider an implementation of a desktop agent that contains multiple applications across multiple languages. If an end-user is blind and using a screen reader, the screen reader will be unable to independently determine which applications have which languages. The desktop agent will need to provide that language information, which it must determine for a given application.

For example: 
> A bank uses a desktop agent to provide a financial application. The application has an English and Chinese version. The bank wants their English-language employees to be able to use the English version of all of the applications, and their Chinese-language employees to be able to use the Chinese version of the applications. The bank could provide those language identifiers in their manifest, and then the desktop agent could surface only certain languages that end-users require.

> A bank uses a desktop agent to provide financial applications to users in Canada. They provide copies of their applications in French and in English, and may even have language-specific applications. Let's say one of the users at this bank is blind and uses a screen reader. They are a native English speaker, so their operating system, desktop application, and screen reader are all running in English. Whenever the screen reader encounters the title or contents of the _french_ applications, it pronounces the french words as if they are English, resulting in cognitively-taxing confusion. Further more, let's say this user is also fluent in french. With no metadata provided for the french versus english applications, when the user encounters any of the french apps, the screen reader will pronounce french words using english pronunciation rules. In other words, they become gibberish.

### Workflow Description
If the AppD provided metadata for the language of the app (assuming the title of the app and the contents of the app are all in the same language), the desktop agent could surface that language appropriately.

### Additional Information
HTML contains the global attribute, `lang`. This attribute can be used on any HTML element, and would likely be leveraged by any HTML-based desktop agent. The lang attribute typically uses BCP 47, which includes a 2-digit language code (like "en" or "fr") and optionally a 2-digit country code (like "en-US" or "en-GB").
[HTML lang attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/lang)

The [W3C's Web Application manifest also supports a `lang` attribute](https://w3.org/TR/appmanifest/#lang-member), with [alternative manifest URLs or geolocation and targeting being used to deliver the localized version of the manifest](https://www.w3.org/TR/appmanifest/#internationalization).

kriswest:
resolves #622 
Adds `lang` and `localizedVersions` fields to the appD application record schema to support accessiblity (screenreaders need to know the primary language of the record content) and localisation (of both appD record's content and the applications themselves through alternative launch url/config/hostManifests)

Change agreed at meeting:
- #657

### DIFF
--- a/src/app-directory/specification/appd.yaml
+++ b/src/app-directory/specification/appd.yaml
@@ -455,6 +455,10 @@ components:
         tooltip:
           type: string
           description: Optional tooltip description e.g. for a launcher
+        lang:
+          type: string
+          pattern: '[a-z]{2}(-[a-zA-Z0-9]{2,8}){0,1}'
+          description:  a language tag that specifies the primary language of both application and it's AppD entry, as defined by IETF RFC 5646.
         description:
           type: string
           description: >-
--- a/src/app-directory/specification/appd.yaml
+++ b/src/app-directory/specification/appd.yaml
@@ -404,18 +404,7 @@ components:
           format: int32
         message:
           type: string
-    Application:
-      description: >
-        Defines an application retrieved from an FDC3 App Directory, which can
-        then be launched.
-
-        Launching typically means running for a user on a desktop.
-        The details around 'launching' including who or what might do it, and how the launch action is initiated are
-        discussed elsewhere in the FDC3 App Directory spec.
-      required:            # details are not required as the host type applications use the hostsManifests mapping instead
-        - appId
-        - name
-        - type
+    BaseApplication:
       properties:
         appId:
           type: string
@@ -457,7 +446,7 @@ components:
           description: Optional tooltip description e.g. for a launcher
         lang:
           type: string
-          pattern: '[a-z]{2}(-[a-zA-Z0-9]{2,8}){0,1}'
+          pattern: '^[a-z]{2}(-[a-zA-Z0-9]{2,8}){0,1}$'
           description:  a language tag that specifies the primary language of both application and it's AppD entry, as defined by IETF RFC 5646.
         description:
           type: string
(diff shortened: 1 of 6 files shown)


kriswest:
@greyseer256  If you're in a reviewing mood, could you check https://github.com/finos/FDC3/pull/670 for me - its another one where the appD schema structure has to change (doesn't change the record structure much, but the YAML had to change) so I need to base other PRs off of it (or deal with a lot of conflicts)

src/app-directory/specification/appd.yaml
```diff
@@ -455,6 +444,10 @@ components:
         tooltip:
           type: string
           description: Optional tooltip description e.g. for a launcher
+        lang:
+          type: string
+          pattern: '^[a-z]{2}(-[a-zA-Z0-9]{2,8}){0,1}$'
+          description:  a language tag that specifies the primary language of both application and it's AppD entry, as defined by IETF RFC 5646.
```
greyseer256, line 450:
```suggestion
          description:  a language tag that specifies the primary language of both application and its AppD entry, as defined by IETF RFC 5646.
```

website/static/schemas/next/app-directory.yaml
```diff
@@ -455,6 +444,10 @@ components:
         tooltip:
           type: string
           description: Optional tooltip description e.g. for a launcher
+        lang:
+          type: string
+          pattern: '^[a-z]{2}(-[a-zA-Z0-9]{2,8}){0,1}$'
+          description:  a language tag that specifies the primary language of both application and it's AppD entry, as defined by IETF RFC 5646.
```
greyseer256, line 450:
```suggestion
          description:  a language tag that specifies the primary language of both application and its AppD entry, as defined by IETF RFC 5646.
```

greyseer256 (approved):
Minor comments, looks good

Issue #664, opened 2022-04-19
App Directory Discussion group - Tuesday 19th April 2022


kriswest:
## Group overview
Discussion group to develop proposals for the improvement of the App Directory as part of the FDC3 standard. The group will examine the role/value/necessity of the AppD as part of the FDC3 specification and a compliance requirement for use of the FDC3 API, in addition to working on proposals to improve the description and functionality of the App Directory. 

### Relevant tags
Current open issues and PRs that relate to the above concepts with the label: 
[![image](https://user-images.githubusercontent.com/1701764/151783600-e65898db-0145-45bc-b500-2e3cf7b1010a.png)](https://github.com/finos/FDC3/labels/app-directory)

A [github discussion](https://github.com/finos/FDC3/discussions/618) is also available for this group (may be used to arrange meetings, agenda ask questions etc.)

## Meeting Date
Tuesday 19th April 2022 - 9am EDT  / 2pm BST

### WebEx info
- Meeting Link: (https://finos.webex.com/finos/j.php?MTID=m2316d45da678d0203fd86f9342ff63ab) 
- Meeting number: 2551 735 2618

*More ways to join*
- Join by video system:
   - Dial 25517352618@finos.webex.com
   - You can also dial 173.243.2.68 and enter your meeting number
- Join by phone
   - +1-415-655-0003 US Toll
   - +44-20319-88141 UK Toll
- Access code: 2551 735 2618

## Meeting notices
- FINOS **Project leads** are responsible for observing the FINOS guidelines for [running project meetings](https://github.com/finos/community/blob/master/governance/Meeting-Procedures.md#run-the-meeting). Project maintainers can find additional resources in the [FINOS Maintainers Cheatsheet](https://odp.finos.org/docs/finos-maintainers-cheatsheet/).

- **All participants** in FINOS project meetings are subject to the [LF Antitrust Policy](https://www.linuxfoundation.org/antitrust-policy/), the [FINOS Community Code of Conduct](https://github.com/finos/community/blob/master/governance/Code-of-Conduct.md) and all other [FINOS policies](https://github.com/finos/community/tree/master/governance#policies).

[...]

PR #671, opened 2022-04-21
562 AppD vendor agnostic native apps

### LINKED ISSUE #562: Add vendor-agnostic properties to the AppD application definition format for launching native apps
## Enhancement Request
Add vendor-agnostic properties to the AppD application definition format for launching native apps

FDC3 and the AppD are expected to support native and hybrid applications in addition to web apps. However, the appD app definition format currently requires a hostManifest (in an unstandardized, vendor-specific format) in order to support launching of native applications. 

Since the merge of PR #437, the appD's application definition format includes a vendor agnostic `details.url` property that can be used to launch web applications. Consider the addition of further sub-properties of `details` that will support vendor-agnostic launching of native applications. 

### Use Case:
Vendors of native apps, and in-house native app development teams may wish to publish details of their apps for inclusion in app directories. Despite the fact that native apps may not be launchable without installation, there are several use-cases for inclusion in an appD:
- Providing configuration to customers that have installed the applications and wish to use them with FDC3 (i.e. to save the customer figuring an optimised configuration themselves).
- Native apps that do support launch without installation (e.g ClickOnce apps).
- In-house developed native apps that are already deployed.
- Supporting discovery of apps that could be installed to work with FDC3

kriswest:
resolves #562 

Introduces launch details for native application types and adjusts the standardized application types to accommodate. Also adjusts the relationship between `type`, `details` and `hostManifests` by eliminating the `host` application type and allowing `hostManifests` to override `details` if they want/need to.  I believe this makes the (vendor agnostic) `details` field more useful and more likely to be used than if it is simply ommited when a hostManifest is used. This allows `details` to become a required field.

Change was agreed in principle (but not in detail) at meeting:
- #664

### DIFF
--- a/src/app-directory/specification/appd.yaml
+++ b/src/app-directory/specification/appd.yaml
@@ -426,10 +426,11 @@ components:
           $ref: '#/components/schemas/Type'
         details:
           description: >-
-            The type specific details of the application. Currently only the "browser" type is standardized.
-            "host" type applications should use the hostManifest's object for all application details.
+            The type specific launch details of the application. These details are intended to be 
+            vendor-agnostic and _MAY_ be duplicated or overridden by details provided in the hostManifests 
+            object for a specific host.
           oneOf:
-            - $ref: '#/components/schemas/BrowserDetails'
+            - $ref: '#/components/schemas/LaunchDetails'
         version:
           type: string
           description: >-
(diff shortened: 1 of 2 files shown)


kriswest:
@ggeorgievx here's an initial proposal for vendor-agnostic native app launch details and the tweaks I mentioned to the type/details/hostManifests relationship. 

Base of the PR is wrong, but have set it up this way so you can see just these changes. Will change the target when we're agreed on it.

In particular, can you check the citrix launch args against what you use?

To run the thing locally run the following in the website folder:
`yarn prebuild; yarn build; yarn start`

then take a look at: http://localhost:3000/schemas/next/app-directory#tag/Application/paths/~1v2~1apps/post

ggeorgievx (commented):
@kriswest LGTM! Thank you for clarifying the relationship between `hostManifests` and `details`!

Two minor things:
- should we add a comment about DAs that don't support certain types of applications?
- should we leave the **host** application type to allow for _other_ application types?

kriswest:
@ggeorgievx 
> * should we add a comment about DAs that don't support certain types of applications?
Good idea. We could either:
- State that at least one application type MUST be supported, but any subset MAY be supported.
- State that at least the Web type must be supported, while other types MAY be supported.

I'm leaning towards the latter as interop between web and other apps types is what I think most would expect of a 'FDC3 compliant' desktop agent and I know of no implementations that don't support web apps...

> * should we leave the **host** application type to allow for _other_ application types?
A good point - although I might prefer and `other` type for that and state that its details would have to be defined by a hostManifest? The first example that springs to mind is something like a python app, but I suspect you could still handle that as native and start with a `path` and `arguments`...
What your preference here?

ggeorgievx:
> I'm leaning towards the latter as interop between web and other apps types is what I think most would expect of a 'FDC3 compliant' desktop agent and I know of no implementations that don't support web apps...

I agree with you.

> A good point - although I might prefer and other type for that and state that its details would have to be defined by a hostManifest?
...
What your preference here?

Happy to rename `host` to `other`. This would give DAs the flexibility to provide other, custom application types e.g. a `node` (Node.js) application type that uses the internal, Electron Node.js runtime.

kriswest:
Heres what it looks like after that last change (to simplify review):
image

image

image

image

image

My only outstanding question is whether `path` & `arguments` is enough for `native` apps (and `url` for online native) or whether we need to support some sort of appAsset and `alias` combo.

ggeorgievx (commented):
> My only outstanding question is whether path & arguments is enough for native apps (and url for online native) or whether we need to support some sort of appAsset and alias combo.

From our side `path` + `arguments` is sufficient. Thank you @kriswest!

ggeorgievx (approved):
> My only outstanding question is whether path & arguments is enough for native apps (and url for online native) or whether we need to support some sort of appAsset and alias combo.

From our side `path` + `arguments` is sufficient. Thank you @kriswest!

PR #673, opened 2022-04-22
369 App categories

### LINKED ISSUE #369: Add  categories metadata to AppD application definitions
## Enhancement Request
Add ~tags~ categories metadata to AppD application definitions.

### Use Case:
A large AppD will have applications in it for a variety of use-cases. To help users navigate and search the AppD metadata fields should be provided to classify applications, tags (keywords) are a convenient and flexible mechanism for doing so that can be used to implement a number of different search or navigation use-cases. Tags are not restrictive as to the meaning of the tags or number that can be applied to an application, ensuring that they can be used to sub-divide the directory in a number of different ways depending on the preferences of the owner of the AppD.

For example: 
> A user wishes to search for applications for working with Fixed Income orders, they can search the UI provided for their AppD implementation via keywords "Fixed Income orders" and review apps returned by search, or they can use the provided navigation to click on the "Fixed Income" tag and review associated apps, or they can use a provided filtering mechanism and select both the "fixed Income" and "orders" tags to filter the list of applications.

Proposed format:
```JSON
{
  "appId": "string",
  "name": "string",
  "manifest": "string",
  ...
  "categories": [
    "Fixed income"
    "Orders"
    "Anther tag"
  ],
  ...
```

kriswest:
resolves #369 

Adds a categories field to the appD record (modelled on the web application manifest), along with a suggested list of categories.


src/app-directory/specification/appd.yaml
```diff
@@ -447,12 +447,57 @@ components:
...
+            - screen sharing<br>
+            - security<br>
+            - spreadsheet<br>
+            - trade cost analysis<br>
+            - trading system<br>
+            - training<br>
+            - travel<br>
+            - video<br>
+            - visualisation<br>
+            - weather<br>
```
greyseer256, line 497: I know it's a nitpick, but I'm just wondering if there's a better way to achieve newlines in yaml than <br> tags on every line. I'm not very familiar with yaml, but I did find this stackoverflow answer that seems relevant: https://stackoverflow.com/questions/3790454/how-do-i-break-a-string-in-yaml-over-multiple-lines

src/app-directory/specification/appd.yaml
```diff
@@ -936,5 +985,11 @@ components:
               },
               Web App Manifest: https://example.com/fdc3-workbench.json
             }
+            localizedVersions: {
+              fr-FR: {
+                title: FDC3 Table de travail,
+                description: Outil de développement et de test pour les desktop agents et applications FDC3
+              }
+            }
```
greyseer256, line 993: This is supposed to be here right? Just wondering cause it seems like it might have belonged in https://github.com/finos/FDC3/pull/670

greyseer256 (approved):
Content looks good

src/app-directory/specification/appd.yaml
kriswest: ~You can break lines in openapi descriptions written in yaml, but they won't render in the output... There are two options for rendering newlines and that's `\n` or `<br>` (although I should probably have written `<br />`... Personally I don't have much of a preference.~

~In theory the github flavoured markdown should be working so this list should have rendered and not needed the line breaks... but it didn't.~

switching from a `>` to a `|` block and putting a newline before the list caused it to render! much cleaner (in both code and rendered) now.

src/app-directory/specification/appd.yaml
kriswest: yep - these PRs keep changing the structure of the record so they are going into a consolidated branch (where the plan for the change was clear). We'll agree the less clear ones individually, merge them into the consolidated branch and then present the whole thing to the standards working group

Issue #312, opened 2021-01-28
The `AppIntent` interface incorrectly describes an intent


ggeorgievx:
## Minor Issue

The `AppIntent` interface incorrectly describes an intent. The `displayName` of the `AppIntent`'s `intent` is ambiguous.

### Area of Issue

[ ] App Directory
[X] API
[ ] Context Data
[X] Intents
[ ] Use Cases
[ ] Other

### Issue Description:

Let's say we have the following FDC3 application definitions (this is just an example to illustrate the issue - some required properties are missing):

```json
[
    {
        "name": "ChartA",
        "intents": [
            {
                "name": "fdc3.ViewChart",
                "displayName": "Chart A's View Chart Intent",
                "contexts": [
                    "fdc3.instrument"
                ]
            }
        ]
    },
    {
        "name": "ChartB",
        "intents": [
            {
                "name": "fdc3.ViewChart",
                "displayName": "Chart B's View Chart Intent",
                "contexts": [
                    "fdc3.instrument"
                ]
            }
        ]
    }
]
```

After calling `fdc3.findIntent('fdc3.ViewChart')` we would get the following `AppIntent`:

```js
{
    intent: {
        name: 'fdc3.ViewChart',
        displayName: '...'
    },
    apps: [
        {
            name: 'ChartA'
        },
        {
            name: 'ChartB'
        }
    ]
}
```

The issue is that the display name is ambiguous - it could be `Chart A's View Chart Intent` or `Chart B's View Chart Intent`.

Going through the existing open-source FDC3 implementations we noticed that both [Nick Kolba's desktop-agent](https://github.com/finos/fdc3-desktop-agent/blob/eb2021491323c12713b60a621fcfd89c2ef52bf7/src/bg-listeners.ts#L835), as well as [Finsemble's FDC3 implementation](https://github.com/ChartIQ/finsemble-fdc3/blob/f69078551ea5b8e6dcc41744bbd3ec09eb3f8a3c/src/services/FDC3/desktopAgent.ts#L99) would return an `AppIntent` with the intent display name of the first application.

### Additional Context:

A PR for this issue would require a change to the existing `AppIntent` interface.

rikoe:
This is a good catch, we need to fix this, and 2.0 will be a good place to do this, as it may require breaking changes to the `AppIntent` interface.

kriswest:
[...]

PR #701, opened 2022-05-09
681 Update Intent Naming Conventions

mistryvinay:
Update Intent Naming Conventions

### DIFF
--- a/docs/intents/spec.md
+++ b/docs/intents/spec.md
@@ -37,9 +37,8 @@ When creating Intents they should be:
 All standard intent names are reserved. Applications may use their own intents ad hoc.
 However, there is occasionally a need for applications to ensure that their intents avoid collision, for example, where a workflow is highly specific to or internal to an application. The recommended approach is to namespace the intent with the application name. For example, the ‘myChart’ App may expose the ‘ViewChart’ intent and the ‘myChart.Foo’ proprietary intent.
 
-### Intents
-
-The FDC3 2.0 standard expanded the Intents that are available.  The list of available Intents are the following:
+### Intent Name Prefixes
+Early versions of the FDC3 standard included 8 intents, which used one of two different prefixes which help to define the behavior of the expected of an app resolving the intent. FDC3 2.0 expanded this set to include the following:
 
 ### `View___`
   * Expected context: Contact
--- a/docs/intents/spec.md
+++ b/docs/intents/spec.md
@@ -8,14 +8,20 @@ title: Intents Specification (next)
 
 FDC3 [Intents](intents-intro) define a standard set of verbs that, in conjunction with context data acting as nouns, can be used to put together common cross-application workflows on the financial desktop.
 
-### Naming Syntax
+### Naming Conventions
+
+The FDC3 standard included two initial Intents which were `View___` and `Start___`.  As more use cases were identified it was clear further Intents were required.
+
+Naming of Intents must follow the below guidelines:
 * Intent names should be free of non-alphanumeric characters.
 * ‘.’ will be used to namespace the intent (see below).
 * Intent names should be in UpperCamelCase.
 
+> **Note:** The naming guidelines should be adhered to when creating future Intents.  This is to ensure they meet the criteria for addition to the FDC3 standard.
+
 ### Characteristics
 
-Intents shoulde be:
+When creating Intents they should be:
 * Recognizable
     * Generally self-evident what the thing is
 * Repeatable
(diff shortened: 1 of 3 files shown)


docs/intents/spec.md
```diff
@@ -31,29 +37,61 @@ Intents shoulde be:
...
+  * Expected characteristics: todo
+
+### `Delete___`
+  * Expected characteristics: Delete an existing record, but fail if it does not exist
+
+### `Get___`
+  * Expected characteristics: Retrieve an existing record, but fail if it does not exist
+
+### `Share___`
+  * Expected characteristics: Share an existing record, but fail if it does not exist
```
kriswest, line 66: The first two need updating as well CreateOrUpdate, does this work for you:

```suggestion
### `View___`
  * Expected behaviour: Content should be displayed to the user.

### `Start___`
  * Expected behaviour: An interaction, such as a chat room or email thread, should be initiated.

### `Create___`
  * Expected behaviour: A new record or entity should be created. The operation should fail if it already exists.

### `Update___`
  * Expected behaviour: An existing record or entity should be updated. The operation should fail if it does not exist.

### `CreateOrUpdate___`
  * Expected behaviour: A new record or entity should be created, or an existing one updated if it exists.

### `Delete___`
  * Expected behaviour: An existing record or entity should be deleted. The operation should fail if it does not exist.

### `Get___`
  * Expected behaviour: A record or entity should be retrieved and returned as an intent result. The operation should fail if the record does not exist.

### `Share___`
  * Expected behaviour: A record or entity should shared. The operation should fail if it does not exist
```

docs/intents/spec.md
```diff
@@ -31,29 +37,61 @@ Intents shoulde be:
...
+
+
+
+
+
 ## Initial Set of Standard Intents ##
```
kriswest, line 72: I'm not sure we need this anymore and could replace with a comment that standardized intents are defined in the following pages, otherwise just cut it down to a bulleted list with each entry having a link to the relevant page...

kriswest (changes requested):
Thanks Vinay, good start. I've added some suggested changes

A changelog entry is also needed in [CHANGELOG.md](https://github.com/mistryvinay/FDC3/blob/master/CHANGELOG.md)

docs/intents/spec.md
```diff
@@ -8,14 +8,20 @@ title: Intents Specification (next)
 
 FDC3 [Intents](intents-intro) define a standard set of verbs that, in conjunction with context data acting as nouns, can be used to put together common cross-application workflows on the financial desktop.
 
-### Naming Syntax
+### Naming Conventions
+
+The FDC3 standard included two initial Intents which were `View___` and `Start___`.  As more use cases were identified it was clear further Intents were required.
```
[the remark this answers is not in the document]
mistryvinay: Have updated based on these comments.

docs/intents/spec.md
mistryvinay: Hi @kriswest is there a way to  have a directory of the Ref/ folde? As this will allow future intents to appear as and when they are added?

As currently I need to include a hyperlink for each definition.

kriswest: We've got the sidebar nav, but thats all. Hence, we either remove the list and just mention that the individual types are on subsequent pages, or manually add a link for each... 

It would be fine (and probably easier) to do this all in a subsequent PR when we've got the full set of intent and context PRs lined up.

kriswest (approved):
LGTM

robmoffat:
I think you should go further with the namespacing.   I think that each intent should have a proper namespaced ID, and a display name.  

- `fdc3.` is reserved for standardised FDC3 intents.
- anything else should be a domain that you own e.g. com.google.some.intent. (if it were google creating the intent).

I am not sure how this would affect intent resolution.  It's possible that there could be multiple intents with the same English name.   

This means if two apps are implementing the same intent, it's very clear, rather than just an accidental collision.

I am willing to accept that I might be completely off with this.. just wanted to know if it'd been discussed.

kriswest:
@robmoffat Namespacing has been discussed previously.  The standardized intents are not currently namespaced - adding that to them would of course be a breaking change and it was deemed unnecessary. There's also a long-standing issues in the API and appD (e.g. https://github.com/finos/FDC3/issues/312) regarding intent display names (currently configured by each app, which doesn't make sense) meaning having them bare/human readable can actually be useful. It also differentiates them slightly from context type names.

That leaves us with only optional namespacing for proprietary intents which is there purely to differentiate a proprietary intent and avoid collision, where the rest of the time you want people to be reusing the same intents and contexts to promote uptake of common workflows and interchangeability of components/widespread interop of applications.

All that said, I would prefer a situation where display names were sorted out and we could be using true IDs. However I think that's a different issue and we should pick up and deal with this on:
- #312
- an additional issue to add /contexts and /intents endpoints to the appD API or standardized types with display names (so you have somewhere authoritative to get the display names from rather than individual app configs).

Issue #703, opened 2022-05-10
Context Data & Intents Discussion group - 12 May 2022


mistryvinay:
## Group overview
At the FDC3 Use Cases Roundtable London, October 5th 2021 participants agreed that the FDC3 lexicon needs to be expanded, both with additional intents and context types to support Use Cases, but also to include more primitive data types in order to construct complex types. A number of participants also agreed that now is an appropriate time to expand the Lexicon.
​
See #455 for full details of the meeting outcomes.
​
This group is being convened to discuss and arrange work to contribute further Context types and Intents to support Use Cases being implemented by participants.

### Relevant issue tags
Current open issues that relate to the discussion group: 
[![image](https://user-images.githubusercontent.com/1701764/139841356-416d9276-81f5-4fce-b5d8-f856d4bda855.png)](https://github.com/finos/FDC3/labels/Context%20Data%20%26%20Intents)

Issues will also be tagged with the labels:
[![image](https://user-images.githubusercontent.com/1701764/139841631-c6bf8697-fafe-4266-8552-63a648625e8b.png)](https://github.com/finos/FDC3/labels/context-data)
[![image](https://user-images.githubusercontent.com/1701764/139841690-969a9d14-a0aa-44a7-bdfe-8c4a43ad30d9.png)](https://github.com/finos/FDC3/labels/intents)

## Meeting Date
Thursday 12 May 2022 - 9am EST / 2pm GMT

### WebEx info
- Meeting Link: [https://finos.webex.com/finos/j.php?MTID=mbb193ae433ce8d4366adcecd94355734](https://finos.webex.com/finos/j.php?MTID=mbb193ae433ce8d4366adcecd94355734)
- Meeting number: 2556 257 8293

*More ways to join*
- Join by video system:
   - Dial 25562578293@finos.webex.com
   - You can also dial 173.243.2.68 and enter your meeting number
- Join by phone
   - +1-415-655-0003 US Toll
   - +44-20319-88141 UK Toll
- Access code: 2556 257 8293

## Meeting notices
- FINOS **Project leads** are responsible for observing the FINOS guidelines for [running project meetings](https://github.com/finos/community/blob/master/governance/Meeting-Procedures.md#run-the-meeting). Project maintainers can find additional resources in the [FINOS Maintainers Cheatsheet](https://odp.finos.org/docs/finos-maintainers-cheatsheet/).

[...]

PR #704, opened 2022-05-11
645 Adding privitive field type conventions to context spec

### LINKED ISSUE #645: Standardize recommended formats for 'primitive' Context fields
## Enhancement Request

Add recommended formats (SHOULD) for common field types to the Context specification, where those fields don't (or don't often or always) warrant a full embedded context type. Those field types should reference external standards where possible.

Note recommending a format for fields of these types doesn't preclude them *also* having a context type. However, if it is decided that they should *always* be full context types then that should instead be _required_ in the standard and this information distributed amongst each individual type.

### Use-case
Providing recommended formats will reduce or prevent competing conventions from being established in both standardized types and in-house types created by app developers - helping to ensure applications can interoperate through FDC3 more reliably. 

Further, conversion of a format or code is often trivial if the target format is specified and it can be validated by a schema. Whereas interpretation of diverse formats from a single field can be far more difficult as is validation via a schema.

### Examples
For example, the following types of data might have a recommended format:
- Time
  -  [ISO 8601-1:2019](https://www.iso.org/standard/70907.html) (already linked to in [References page](https://fdc3.finos.org/docs/next/references))
    - A timezone indicator should be included
    - E.g. 
      - Time in UTC: `2022-03-30T15:44:44Z`
      - Also time in UTC: `2022-03-30T15:44:44+00:00`
      - Same time in EDT: `2022-03-30T11:44:44-04:00`
    - Parsing in Javascript: `let aDate = new Date("2022-03-30T11:44:44-04:00")`
- Date
  - `YYYY-MM-DD` date format from [ISO 8601-1:2019](https://www.iso.org/standard/70907.html) (already linked to in [References page](https://fdc3.finos.org/docs/next/references))
    - E.g. `2022-03-30`
    - Parsing in Javascript: `let aDate = new Date("2022-03-30")`
- Country code
  - Alpha-2-code from [ISO 3166-1](https://www.iso.org/iso-3166-country-codes.html)  (already linked to in [References page](https://fdc3.finos.org/docs/next/references))
    - Note current fdc3.Country type uses *both* the Alpha-2 and Alpha-3 codes - picking one would simplify requirements
  - E.g. `GB` or `UA`
- Currency code 
  - Alphabetic code from [ISO 4217](https://www.iso.org/iso-4217-currency-codes.html) (needs adding to [References page](https://fdc3.finos.org/docs/next/references))
  - E.g. `GBP` or `UAH`

kriswest:
resolves #645 

Introduces a Field Types Conventions section, which contains the new recommendations + the existing convention on id fields:

image
image


kriswest:
Updated with feedback from:
- #703 

To change country code recommended field name to `COUNTRY_ISOALPHA2` for consistency with `CURRENCY_ISOCODE`. Also updated in `Country` context type

nemery-flextrade:
As discussed ISDA's Non-ISO currency code expansion is available here: https://fpml.org/coding-scheme/non-iso-currency

PR #706, opened 2022-05-12
646 adding a context representing a range of time: fdc3.timerange

### LINKED ISSUE #646: Standardize a context type representing a range of time: TimeRange
## Enhancement Request (edited 6th May 2022)

Create a context type that can be used to represent a period of time.

### Use Case:
Any user interfaces that represent or visualise events or activity over time can be filtered or focused on a particular time period, e.g.:
- A pricing chart
- A trade blotter
- An record of client contact/activity in a CRM

A time period may be closed (i.e. start and end date are both known) or open (i.e. only the start date is known, an end date will be known in the future, e.g. when an order is filled).

Example application in a raised intent: 
> User may want to view pricing/trades/customer activity for a security over a particular time period, the time range might be specified as the context for the `ViewChart` intent OR it be embedded in another context (e.g. a Chart context).
  
Example application as a broadcast context:
> User filters a visualisation (e.g. a pricing chart) to show a particular period, the TimeRange is broadcast, other visualisations (e.g.  a heatmap of activity by instrument, or industry sector etc.) receive it and filter themselves to show data over the same range.

#### TimeRange
Represents a specific period of time.

##### Details

| Property    | Type    | Required | Example Value       |
|-------------|----------|----------|---------------------|
| `type`         | string    | Yes        | `"fdc3.timeRange"` |
| `startTime` | string *  | Yes        | `"2022-03-30T15:44:44+00:00"`       |
| `endTime`  | string *  | No        | `"2022-04-30T23:59:59+00:00"`            |

\* The time should be encoded in the [ISO 8601-1:2019](https://www.iso.org/standard/70907.html) format with a timezone indicator included

##### Example
A closed range:
```js
const timeRange = {
    type: "fdc3.timeRange",
    starttime: "2022-03-30T15:44:44+00:00",
    endtime: "2022-04-30T23:59:59+00:00"
}
```
An open range:
```js
const timeRange = {
    type: "fdc3.timeRange",
    starttime: "2022-03-30T15:44:44+00:00"
```

### Additional Information
The format of the strings representing times should refer to advice in the standard about formats, as proposed in: https://github.com/finos/FDC3/issues/645, with a note/link appearing in this context definition.

kriswest:
resolves #646 

image


robmoffat:
This seems like it would be compatible with new JS like Temporal

kriswest:
Preview link: https://deploy-preview-706--lambent-kulfi-cf51a7.netlify.app/docs/next/context/ref/timerange

PR #709, opened 2022-05-12
652 adding a context representing the price and value of a holding: fdc3.valuation

### LINKED ISSUE #652: Standardize a context type representing a Valuation
## Enhancement Request
Create a context type representing the price and value of a holding. 

### Use Case:
A Valuation may be used as a key component of a Trade or Position context, perhaps also an Order. Although it may be used rarely as a standalone context (although use-cases may exist), it is a multi-field type that is likely to be needed in the definition of multiple other types, yielding an advantage to standardizing it.

#### Valuation
Represents the value of a holding in a specified currency.

##### Details

| Property    | Type    | Required | Example Value       |
|-------------|----------|----------|---------------------|
| `type`         | string    | Yes         | `'fdc3.valuation'`   |
| `value`       | number | Yes         | `'500.0'`                 |
| `price`        | number | No         | `'5.0'`                     |
| `CURRENCY_ISOCODE`  * | string  | Yes       | `'USD'`         |

\* CURRENCY_ISOCODE should conform to 3 character alphabetic codes standardized as part of ISO 4217.

##### Example
```js
const valuation = {
    type: 'fdc3.valuation',
    value: 500.0,
    price: 5.0,
    CURRENCY_ISOCODE: 'USD'
}
```

### Additional Information
An issue has also been raised to standardize recommended formats for 'primitive' fields, which might include ISO currency codes:
- #645 
 
If that issue is NOT accepted the CURRENCY_ISOCODE field should be replaced with a `currency` field that makes use of the Currency context type prosed in issue:
- #651

kriswest:
resolves #652


kriswest:
Preview link: https://deploy-preview-709--lambent-kulfi-cf51a7.netlify.app/docs/next/context/ref/valuation

Issue #692, opened 2022-05-04
Desktop Agent Bridging Discussion group 4th May 2022


kriswest:
## Group overview
Discussion group for developing proposals for producing bridges between FDC3 implementations (aka Desktop Agents), allowing applications running on one Desktop Agent, to integrate with FDC3 applications running on a second Desktop Agent for the same user.

The interop between applications running on different Desktop Agents aka Platforms would ideally cover
- Channels
- Raising and resolving Intents
- Launching Applications

### Relevant issue tags
- Current open issues that relate to the above concepts with the label: 
  [![image](https://user-images.githubusercontent.com/1701764/144856698-e49821ab-9c8e-4daf-99a1-23f4774c1fcd.png)](https://github.com/finos/FDC3/labels/Desktop%20Agent%20Bridging)
- The original issue is now a github discussion where content can be added and this issue discussed: 
#544

## Meeting Date
Wednesday 04 May 2020 - 9am EST / 2pm GMT

### WebEx info
- Meeting Link: [insert link](https://finos.webex.com/finos/j.php?MTID=m55d4aa79e148c948bf53163690e74fee)
- Meeting number: 2558 920 8729

*More ways to join*
- Join by video system:
   - Dial 25589208729@finos.webex.com
   - You can also dial 173.243.2.68 and enter your meeting number
- Join by phone
   - +1-415-655-0003 US Toll
   - +44-20319-88141 UK Toll
- Access code: 2558 920 8729

## Meeting notices
- FINOS **Project leads** are responsible for observing the FINOS guidelines for [running project meetings](https://github.com/finos/community/blob/master/governance/Meeting-Procedures.md#run-the-meeting). Project maintainers can find additional resources in the [FINOS Maintainers Cheatsheet](https://odp.finos.org/docs/finos-maintainers-cheatsheet/).

- **All participants** in FINOS project meetings are subject to the [LF Antitrust Policy](https://www.linuxfoundation.org/antitrust-policy/), the [FINOS Community Code of Conduct](https://github.com/finos/community/blob/master/governance/Code-of-Conduct.md) and all other [FINOS policies](https://github.com/finos/community/tree/master/governance#policies).

[...]

Issue #628, opened 2022-03-17
Desktop Agent Bridging Discussion group - 23rd March 2022


kriswest:
## Group overview
Discussion group for developing proposals for producing bridges between FDC3 implementations (aka Desktop Agents), allowing applications running on one Desktop Agent, to integrate with FDC3 applications running on a second Desktop Agent for the same user.

The interop between applications running on different Desktop Agents aka Platforms would ideally cover
- Channels
- Raising and resolving Intents
- Launching Applications

### Relevant issue tags
- Current open issues that relate to the above concepts with the label: 
  [![image](https://user-images.githubusercontent.com/1701764/144856698-e49821ab-9c8e-4daf-99a1-23f4774c1fcd.png)](https://github.com/finos/FDC3/labels/Desktop%20Agent%20Bridging)
- The original issue is now a github discussion where content can be added and this issue discussed: 
#544

## Meeting Date
Wednesday 23rd Mar 2022 - 9am EDT / 1pm GMT

### WebEx info
- [Meeting Link](https://finos.webex.com/finos/j.php?MTID=m55d4aa79e148c948bf53163690e74fee)
  Meeting number: 2558 920 8729
  Password:7cgQYd2eYT3

*More ways to join*
- Join by video system:
   - Dial 25589208729@finos.webex.com
   - You can also dial 173.243.2.68 and enter your meeting number
- Join by phone
   - +1-415-655-0003 US Toll
   - +44-20319-88141 UK Toll
- Access code: 2558 920 8729

## Meeting notices
- FINOS **Project leads** are responsible for observing the FINOS guidelines for [running project meetings](https://github.com/finos/community/blob/master/governance/Meeting-Procedures.md#run-the-meeting). Project maintainers can find additional resources in the [FINOS Maintainers Cheatsheet](https://odp.finos.org/docs/finos-maintainers-cheatsheet/).

- **All participants** in FINOS project meetings are subject to the [LF Antitrust Policy](https://www.linuxfoundation.org/antitrust-policy/), the [FINOS Community Code of Conduct](https://github.com/finos/community/blob/master/governance/Code-of-Conduct.md) and all other [FINOS policies](https://github.com/finos/community/tree/master/governance#policies).

[...]

Issue #611, opened 2022-03-02
Desktop Agent Bridging Discussion group - 9th March 2022


kriswest:
## Group overview
Discussion group for developing proposals for producing bridges between FDC3 implementations (aka Desktop Agents), allowing applications running on one Desktop Agent, to integrate with FDC3 applications running on a second Desktop Agent for the same user.

The interop between applications running on different Desktop Agents aka Platforms would ideally cover
- Channels
- Raising and resolving Intents
- Launching Applications

### Relevant issue tags
- Current open issues that relate to the above concepts with the label: 
  [![image](https://user-images.githubusercontent.com/1701764/144856698-e49821ab-9c8e-4daf-99a1-23f4774c1fcd.png)](https://github.com/finos/FDC3/labels/Desktop%20Agent%20Bridging)
- The original issue is now a github discussion where content can be added and this issue discussed: 
#544

## Meeting Date
Wednesday 9th Mar 2022 - 9am EST / 2pm GMT

### WebEx info
- [Meeting Link](https://finos.webex.com/finos/j.php?MTID=ma4b1de1b9f691d89e4b2ff380ec9687c)
Meeting number: 2556 495 5451

*More ways to join*
- Join by video system:
   - Dial 25564955451@finos.webex.com
   - You can also dial 173.243.2.68 and enter your meeting number
- Join by phone
   - +1-415-655-0003 US Toll
   - +44-20319-88141 UK Toll
- Access code: 2556 495 5451

## Meeting notices
- FINOS **Project leads** are responsible for observing the FINOS guidelines for [running project meetings](https://github.com/finos/community/blob/master/governance/Meeting-Procedures.md#run-the-meeting). Project maintainers can find additional resources in the [FINOS Maintainers Cheatsheet](https://odp.finos.org/docs/finos-maintainers-cheatsheet/).

- **All participants** in FINOS project meetings are subject to the [LF Antitrust Policy](https://www.linuxfoundation.org/antitrust-policy/), the [FINOS Community Code of Conduct](https://github.com/finos/community/blob/master/governance/Code-of-Conduct.md) and all other [FINOS policies](https://github.com/finos/community/tree/master/governance#policies).

[...]

PR #727, opened 2022-05-20
717 Add a recommended set of user channels to the Standard

### LINKED ISSUE #717: Add a recommended set of user channel definitions to the API
## Enhancement Request

Add a set of recommended user channel definitions, with display metadata, to the FDC3 API Standard. 

### Use Case:

The FDC3 API includes a [`fdc3.getUserChannels()`](https://fdc3.finos.org/docs/next/api/ref/DesktopAgent#getuserchannels) (which replaces `fdc3.getSystemChannels()` in FDC3 2.0) function that should return the desktop agent's set of user channels, however, it is left up to each implementation to come up with that set. As these channels are intended for use by end-users (who select which channel an app is on), this has not been an issue in the past.

However, the Desktop Agent Bridging discussion group has identified use cases for a user to be working with multiple desktop agents (on one or more machines) and for those agents to be able to exchange messages through a 'bridge' allowing the apps running under the different agents to interoperate.  If the bridged desktop agents have different sets of user channels the user experience of context sharing via user channels becomes challenging.

While mapping broadcasts from one channel on a first desktop agent to a different channel on another agent is possible via configuration (either static or user administrated in a UI), the UX is still impacted. 

Hence, a set of user channels is proposed for inclusion in the standard, which will be recommended but not required for compliance purposes (SHOULD). A Desktop Agent MAY then also support configuration of their user channels. 

### References:
- https://github.com/finos/FDC3/discussions/544#discussioncomment-1997387
- #611
- #628 
- #692

### Proposed Channels
```JSON
[
    {
        "id": "A",
        "type": "user",
        "displayMetadata": {
            "name": "Channel A",
            "color": "red",
            "glyph": "A"
        }
    },
    {
        "id": "B",
        "type": "user",
        "displayMetadata": {
            "name": "Channel B",
            "color": "orange",
            "glyph": "B"
        }
    },
    {
        "id": "C",
        "type": "user",
        "displayMetadata": {
            "name": "Channel C",
            "color": "yellow",
            "glyph": "C"
        }
    },
    {
        "id": "D",
        "type": "user",
        "displayMetadata": {
            "name": "Channel D",
            "color": "green",
            "glyph": "D"
        }
    },
    {
        "id": "E",
        "type": "user",
        "displayMetadata": {
            "name": "Channel E",
            "color": "lightblue",
            "glyph": "E"
        }
    },
    {
        "id": "F",
        "type": "user",
        "displayMetadata": {
            "name": "Channel F",
            "color": "blue",
            "glyph": "F"
        }
    },
    {
        "id": "G",
        "type": "user",
        "displayMetadata": {
            "name": "Channel G",
            "color": "purple",
            "glyph": "G"
        }
    },
    {
        "id": "H",
        "type": "user",
        "displayMetadata": {
            "name": "Channel H",
            "color": "brown",
            "glyph": "H"
        }
    }
]
```

Note:
- Colors are defined using [CSS standard colors (extended)](https://www.w3.org/wiki/CSS3/Color/Extended_color_keywords). Individual implementations may adjust the specific color chosen but SHOULD choose a shade close to the default specified in [CSS3](https://www.w3.org/wiki/CSS3/Color/Extended_color_keywords)
- Alphanumeric glyph provided for colourblind users, matched to channel id.

kriswest:
resolves #717 

Adds a set of recommended channel definitions to the API to encourage desktop agents to use a consistent set to aid a future Desktop Agent Bridging addition.


nkolba (commented):
Not sure what a recommended set of user channels will achieve.  A "SHOULD" won't guarantee consistent experience for end users and a "MUST" would be a non-starter.  Also, while this is a brave stab at coming up with a standard set of channels (and I appreciate the move away from colors only), the proposal tempts a synethesiac rabbit hole of which color should be associated with what letter.  I'd be in favor of tabling this and taking up again in the context of the desktop bridging discussion.  A more flexible mechanism that allowed for a federated approach across bridges would make more sense and provide end users with much greater consistency and determinism.

kriswest:
You can find minutes for the 3 previous occasions that this has been discussed in the Desktop Agent bridging group meetings (added as references on the issue) at:
- #611
- #628
- #692

We did discuss other mechanisms for doing this. Providing a mapping system in the bridge was not a popular solution as its very hard to expose to the user what the mappings are. Likewise mapping on each end (which might prove inconsistent and more complex to administer). It will always result in a poor UX where messages on a channel with one name/appearance pop out on one with different details.

>A "SHOULD" won't guarantee consistent experience for end users and a "MUST" would be a non-starter.

True, but it will at least encourage one - which no other solution proposed actually does (mapping channels to each other doesn't do that). If a MUST is a non-starter, then this appears to be the best option don't he table.

> Also, while this is a brave stab at coming up with a standard set of channels (and I appreciate the move away from colors only), the proposal tempts a synethesiac rabbit hole of which color should be associated with what letter.

For colourblind users, a purely colour-based channel set is not usable - I would assume that's why `DisplayMetadata` has a glyph field in the first place. Colours are also a challenge for other visually impaired users unless suitable names or glyphs that can be read are applied. It doesn't matter to me whether we use letters or numbers and I would hope the community is mature enough to avoid getting bogged down in a 'synethesiac rabbit hole' and to forgo accessibility for the sake of aesthetics. Do you prefer numbers @nkolba?

greyseer256 (approved):
I actually think this is a reasonable mapping of colors and letters, since the colors are ordered by decreasing wavelength (visible spectrum red to purple/violet). Of course brown is an outlier, I'd almost rather use black or white or just cap it at 7 and leave additional ones up to the desktop agent implementation.

greyseer256:
Another thought, maybe Arabic numerals are more internationally recognized than Latin alphabet characters?

kriswest:
Both good points @greyseer256, I've switched to Arabic numerals and tweak the back half of the range to: cyan, blue, magenta, purple (in the correct order in the spectrum I think), which (to my eyes at least) is easier to differentiate than before.

Heres a preview:
image


preview code
```
 <div style="position: fixed;top: 10px;left: 10px;"><div style="width: 25px; height: 25px; background-color: red;color: black;text-align: center;line-height: 25px;display: inline-block;margin-right:5px;">1</div><div style="width: 25px; height: 25px; background-color: orange;color: black;text-align: center;line-height: 25px;display: inline-block;margin-right:5px;">2</div><div style="width: 25px; height: 25px; background-color: yellow;color: black;text-align: center;line-height: 25px;display: inline-block;margin-right:5px;">3</div><div style="width: 25px; height: 25px; background-color: green;color: black;text-align: center;line-height: 25px;display: inline-block;margin-right:5px;">4</div><div style="width: 25px; height: 25px; background-color: cyan;color: black;text-align: center;line-height: 25px;display: inline-block;margin-right:5px;">5</div><div style="width: 25px; height: 25px; background-color: blue;color: black;text-align: center;line-height: 25px;display: inline-block;margin-right:5px;">6</div><div style="width: 25px; height: 25px; background-color: magenta;color: black;text-align: center;line-height: 25px;display: inline-block;margin-right:5px;">7</div><div style="width: 25px; height: 25px; background-color: purple;color: black;text-align: center;line-height: 25px;display: inline-block;margin-right:5px;">8</div></div>
```

greyseer256 (approved):
👍

openfin-johans:
As mentioned on https://github.com/finos/FDC3/issues/717 I think we must add some form of "fdc3" part to the id to ensure there's no mismatch with any existing or other channels.

I think we had comments about namespacing before so not sure if `fdc3.channel.1` is the way to go? Should it be `fdc3_channel_1`?

kriswest:
> I think we had comments about namespacing before so not sure if `fdc3.channel.1` is the way to go? Should it be `fdc3_channel_1`?

I'm not sure the namespacing discussion was related to channel names specifically - but I suggested this (with periods) for the sake of consistency with context types and intent names. However, I don't think it matters a whole lot - we can even have spaces (e.g. "fdc3 user channel 1").

I'll update the PR with the version with periods - but give me a shout if there's a strong argument for underscores or something else.

kriswest:
@openfin-johans you can see the revised list of channels here:
https://deploy-preview-727--lambent-kulfi-cf51a7.netlify.app/docs/next/api/spec#recommended-user-channel-set

robmoffat (approved):

Issue #247, opened 2020-09-24
Extend AppD application definition to support describing an app's use of interop (App Channels, User Channels & Intents)


kriswest:
## Enhancement Request
Extend the AppD's [Application definition](https://fdc3.finos.org/schemas/1.2/app-directory#tag/Application) to include details of how an application uses Interop, including: 
- what context types are supported by the application.
- how those context types are used (listen or broadcast on user (system) channels, app channels or passed with an intent).
- what app channel sit broadcasts on or listens to.
- what intents and contexts it can raise.

### Use Case:
When assembling a desktop, Users (or desktop implementors) may wish to search an App directory for applications that 'work with' a selected application. This is currently not possible as AppD application records can only provide detail on intent/context combination that can be handled, and does not provide information on intents/context raised, channels or contexts broadcast/listened to on them. 

Additionally, app developers or Desktop Agent vendors may wish to be able to build UIs that can `fdc3.open` applications supporting a particular context. This information is known to the desktop agent at runtime *after* the application has been started, but is not known to either the Desktop Agent or  application developer in advance except by direct interaction with the developer/publisher of each application.

N.B. additional changes may be needed for the FDC3 API to support the latter use case. However, a first step to doing so is to decide if and how the necessary information is provided to the Desktop Agent.

### Workflow Description
A desktop agent or desktop developer may wish to implement functionality that leverages information on what context types are broadcast and listened to by applications in the desktop. E.g. The user browses an Application catalog UI (backed by an AppD instance) to select applications for their desktop and adds one (e.g. a view of the current orderbook developed in-house) to their launcher app. They then wish to find other applications that will work with it, both in an application catalog and via an 'open in ...' function (`fdc3.open`) or 'perform action' (`fdc3.raiseIntent`, `fdc3.raiseIntentForContext`, `fdc3.findIntent`, `fdc3.findIntentForContext`) function within the application itself.

### Workflow Examples
A user (or desktop administrator that manages a desktop for users) is browsing an app catalog to select applications for use on their desktop:
- The app selected supports context sharing via `fdc3.broadcast`, the user wishes to see a list of other apps in the application catalog that listen for that context via channel linking (`fdc3.addContextListener`)
- The app selected supports context listening via `fdc3.addContextListener`, the user wishes to see a list of other apps in the application catalog that broadcast that context via channel linking (`fdc3.broadcast`)
[...]

PR #730, opened 2022-05-23
247 appd use of interop

kriswest:
resolves https://github.com/finos/FDC3/issues/247

Adds a set of fields to AppD application records to describe an app's use of FDC3 and enable search for apps that 'interoperate' with a selected app.


kriswest:
Closing again as preview didn't generate - one more go!

PR #748, opened 2022-06-08
Update the Glue42 logo

ggeorgievx:


kriswest (approved):
LGTM - fits better than the other one

PR #752, opened 2022-06-14
fix homepage URL

CrlsMrls:
In get-involved page, the homepage URL points to https://fdc3.finos.org/fdc3.finos.org instead of https://fdc3.finos.org/ because the protocol is missing.


kriswest (approved):
LGTM, thanks @CrlsMrls!

PR #754, opened 2022-06-16
745 generate context typescript files for 2.0 (and apply prettier)

### LINKED ISSUE #745: Generate typescript types for contexts added in 2.0
Prior to release the FDC3 source and NPM module need to be updated with new typescript types for standard contexts generated from the schemas. These go in [src/context/ContextType.ts](https://github.com/finos/FDC3/blob/master/src/context/ContextType.ts) and [src/context/ContextTypes.ts](https://github.com/finos/FDC3/blob/master/src/context/ContextTypes.ts).

A script is provided to do this: https://github.com/finos/FDC3/blob/544-Desktop-Agent-Bridging-Proposal/package.json/#L30-L31
`yarn typegen`

kriswest:
resolves #745

### DIFF
--- a/src/context/ContextTypes.ts
+++ b/src/context/ContextTypes.ts
@@ -1,17 +1,23 @@
 // To parse this data:
 //
-//   import { Convert, Context, Contact, ContactList, Instrument, InstrumentList, Country, Organization, Portfolio, Position, Nothing } from "./file";
+//   import { Convert, Context, Chart, ChatInitSettings, Contact, ContactList, Country, Currency, Email, Instrument, InstrumentList, Nothing, Organization, Portfolio, Position, TimeRange, Valuation } from "./file";
 //
 //   const context = Convert.toContext(json);
+//   const chart = Convert.toChart(json);
+//   const chatInitSettings = Convert.toChatInitSettings(json);
 //   const contact = Convert.toContact(json);
 //   const contactList = Convert.toContactList(json);
+//   const country = Convert.toCountry(json);
+//   const currency = Convert.toCurrency(json);
+//   const email = Convert.toEmail(json);
 //   const instrument = Convert.toInstrument(json);
 //   const instrumentList = Convert.toInstrumentList(json);
-//   const country = Convert.toCountry(json);
+//   const nothing = Convert.toNothing(json);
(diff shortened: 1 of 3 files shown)


src/context/ContextTypes.ts
```diff
@@ -40,48 +103,61 @@ export interface ContactID {
...
+  subject?: string;
+  textBody?: string;
   type: string;
   id?: { [key: string]: string };
   name?: string;
 }
 
-export interface Instrument {
-  id: InstrumentID;
+export interface RecipientsObject {
```
rikoe, line 138: I don't like `RecipientsObject` as a name for developers to use. Names should be domain-based, not with development concepts like "object" in them.

This is probably a `QuickType` thing, and can hopefully be tweaked by playing with the command line parameters used in package.json.

If not, I think we should investigate these options:
* Try to control the name QuickType generates through its settings
* Manually change the file after generation
* Automatically change the file after generation

rikoe (changes requested):

src/context/ContextTypes.ts
kriswest: I'm not hugely bothered by this as I doubt anyone will use that generated type directly. Its only exported to facilitate type checking. 

I'm also not seeing an option to fix it in the [TypeScript flow renderer in quicktype](https://github.com/quicktype/quicktype/blob/master/src/quicktype-core/language/TypeScriptFlow.ts) (the javascript name option doesn't do it. You can try it here: https://app.quicktype.io/

We may, however, wish to disable the javascript name option as it is replacing capitalized properties or those with underscores with camelcase... Pretty sure we don't want that.

kriswest: Seems like losing the `*Object` names would require at least a custom language definition - which didn't defer to the code generating the names in the ConvenienceRenderer: https://blog.quicktype.io/customizing-quicktype/

rikoe: Okay, I agree it's not a biggie, we can always add a separate issue, and release an update of the npm package only if we find a way to deal with it. I will approve.

rikoe (approved):

mistryvinay (approved):
Looks good to me. 👍

PR #764, opened 2022-06-23
Changed to "community" rather than just implementations.  

robmoffat:
Hi,

I'd like to change the wording on this page to be about "Community" rather than just implementations.  Couple of reasons:
- Some of the content is about training courses etc.  I have a load more to add in that ilk.  These aren't implementations
- I'm going to do the mail shot on Tuesday.  I'd like to be as inclusive as possible, especially for people who are thinking about building stuff but maybe have nothing concrete to show yet.

### DIFF
--- a/docs/fdc3-intro.md
+++ b/docs/fdc3-intro.md
@@ -31,7 +31,7 @@ From its inception, the standards have been informed by real-world [business use
 
 ## Who is using FDC3?
 
-The Financial Desktop Connectivity and Collaboration Consortium (FDC3) standards are created and used by [leading organizations across the financial industry](/users). For more detail on who's using FDC3, developer tools, training and examples, see the [implementations page](/implementations).
+The Financial Desktop Connectivity and Collaboration Consortium (FDC3) standards are created and used by [leading organizations across the financial industry](/users). For more detail on who's using FDC3, developer tools, training and examples, see the [community page](/community).
 
 ## How is FDC3 governed?
 
--- a/website/data/community.json
+++ b/website/data/community.json
@@ -13,7 +13,7 @@
         "infoLink": "https://cosaic.io/chartiq/",
         "docsLink": "https://documentation.chartiq.com/",
         "type": "application-provider",
-        "badges": [{"text": "Supporting FDC3 1.2"} ],
+        "badges": [{"text": "FDC3 1.2 Supported"} ],
         "description": "<p>ChartIQ is a powerful and flexible HTML5 charting library with millions of users worldwide. Written in JavaScript, it runs entirely within the browser. With state of the art integrations, including  third-party apps, educational components, analysis, and more, you can create efficient workflows and streamline the way to work.</p><p>ChartIQ works on any platform (mobile, web, desktop) and any framework (Angular, React).</p><p>Most companies have at least one financial charting library for each platform or app they target—web, C#, Java, Android, iOS, etc., so developers have multiple code bases to maintain. With ChartIQ, write your code once and use it everywhere.</p>"
     },
     {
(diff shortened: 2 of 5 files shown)


kriswest (approved):
I've never been super keen on 'implementations' but couldn't find a better alternative. This works.

Not sure on the supported -> supporting change, but happy for FINOS to call that.

robmoffat:
The  supported -> supporting change is something Vinay suggested.  

e.g. Symphony are "supporting" FDC3 2.0, although they are still on their journey to implementing it...  I think this works- I'm trying to be inclusive here.  

wrt. badges, there will need to be a further change to this later on when we have the conformance program up-and-running since FINOS are getting some swanky graphics designed for apps / containers to display once they have passed.  We'll have to figure out how to make this look nice.

rikoe (changes requested):
I think "supporting" should be changed back to "supported", and if needed a separate indicator for "support coming soon" should be added, if I understand the intention of this change correctly.

mistryvinay:
> I think "supporting" should be changed back to "supported", and if needed a separate indicator for "support coming soon" should be added, if I understand the intention of this change correctly.

Taking a step back. I agree with @rikoe let's make it clear and simple for developers. It's either "supported" or "support coming soon".

robmoffat:
yeah that makes it really clear. good idea

robmoffat:
Just finished the change suggested by @rikoe  and @mistryvinay

rikoe (approved):

mindthegab:
Sorry if I'm late to the party here, but two comments:

- Shouldn't the primary key be (product, version), meaning which version of the product actually supports a specific version of the standard? While I'm sure this information will be available on vendors websites, I expect the FINOS certification to applied against a specific MAJOR version of a product (assuming vendors use semantic versioning, the support should not be broken on MINOR or MAINTENANCE versions). Thinking about forward compatibility of this page with the conformance framework, I would expect the "Certified" badge issued by FINOS to apply to a specific MAJOR version, not to a "vendor" or "product" in general.

- I wonder if  we need to make a stronger differentiation between "supported" (maybe should be "support advertised" or something like that, aka self advertised by a vendor, which we know is not always the case) and "compliant" (i.e formally tested and verified by an independent party, i.e. FINOS, which maybe should be "FINOS certified").  I guess where I'm going is that, increasingly, it shouldn't matter what a vendor "says" it supports, as we have seen over and over again cases, in the docs and in the press, of vendors hooking on the FDC3 name but really not supporting it fully. For the sake of consumers, I think what should matter in the future is that something it's certified. 

Thoughts?

kriswest:
@mindthegab @rikoe @robmoffat It's certainly not too late to change how we present info on platform providers and apps - particularly as another has actually gone out yet. A breakdown of support info over versions is more difficult to display but is definitely useful... I see actual compliance testing / certified compliance badges as a fairly easy addition to that - its actually the breakdown over version that's harder to add/render.

Such a breakdown can be collapsed: perhaps it says '1.1 supported', '1.2 supported' '2.0 coming soon', you pop open the collapse and can see:
- Finsemble 5.x: **1.1 supported**
- Finsemble 6.x - 8.x **1.2 supported**
- Finsemble 9.x: **2.0 coming soon**

Thoughts? For me this isn't 100% necessary, but could be useful and not terrifyingly hard to add (will need an hour or so of hacking HTML, JS and JSON).

kriswest:
> I don't want to close any of this good discussion down, so maybe we can continue it here: https://github.com/finos/FDC3/pull/714

Works for me. Merging this PR.

> as long as we all understand it's in everyone's best interest to strongly differentiate the concept of "formally compliant / certified" from "support advertised"

I certainly agree, add look forward to adopting the compliance framework when ready (into FDC3 and our own CI system ;-) )

Issue #750, opened 2022-06-09
README update / Introductory materials


robmoffat:
## Minor Issue
For a few other FINOS projects we are moving to a more friendly, beginner-centric form of the README, e.g.

https://github.com/finos/spring-bot
https://github.com/finos/morphir 

I think as we approach OSFF we should take a look at the existing README and try and do similar.  Whereas fdc3.finos.org does a great job of laying out the spec for developers, and a somewhat good job of marketing the project in terms of explaining who uses it, the README should be oriented at curious people trying to understand the basic questions of 
 - What is It?
 - Why is it important?
 - What are the benefits?
 - Getting started

There are loads of pre-existing resources around the internet about FDC3 we can draw on and link to to help answer these questions.

rikoe:
The current README was very specifically created to follow the template mandated by the FINOS Standards project blueprint at the time.

However, I am not aware how much that has changed since (quite a bit it seems).

Happy to change it, but again, we should stick to a standard FINOS template as much as possible - is there one that represents the examples you refer to (e.g. Morphir)?

It is also worth pointing out that the current README serves a dual purpose - it is used on npmjs.com when publishing the [`@finos/fdc3` package](https://www.npmjs.com/package/@finos/fdc3). It can be decoupled of course, but they were combined to reduce the maintenance burden and divergence/discrepancies. Any changes made to the project README will therefore also affect the NPM package home page.

robmoffat:
@rikoe thanks that's really useful background.  

Yes, the Morphir one is the template.  Also I included the spring-bot one for reference which I updated a few months back to match.

I am working on a PR here.  I might have something to show next week

PR #770, opened 2022-07-05
More Conformant FDC3 Readme file

robmoffat:
Here is a draft of something for

- #750 

I am not really happy about what this is going to look like in NPM.  It seems we have a couple of choices:

- Add all the typescript instructions in this README, like it was originally... but this seems weirdly specific and out-of-place given the rest of the nature of the project.
- We could move the FDC3 API/Typescript/package.json, so that it's somewhere else, and then we can have a README for the NPM which is separate from the GitHub one.
- We could follow the advice given [here](https://stackoverflow.com/questions/41297117/how-to-specify-different-readme-files-for-github-and-npm) and have README.md for GitHub, and README for NPM, or move the GitHub readme into `.github` dir.  

Ideally, the NPM README should either reference [this page](https://fdc3.finos.org/docs/next/supported-platforms) or be a copy of some of it.  It appears that having the GitHub badges in the NPM file improves the metadata in NPM so we should also have those in there.

oh, you can review how this looks on GitHub [here](https://github.com/robmoffat/fdc3)

wdyt?


kriswest:
Will be superceded by a PR based on https://github.com/finos/FDC3/tree/412_standards_adoption

PR #773, opened 2022-07-06
Proposal update - connection protocol

tpina:


docs/api-bridging/spec.md
```diff
@@ -105,93 +101,209 @@ However, cross-machine routing is an internal concern of the Desktop Agent Bridg
...
+
+1. The incoming channel state `channelState` for a particular channel should be processed from last (oldest) to first.
+2. If there is no current context of that type, broadcast it to listeners of that specific type only.
+3. If there is a current context of that type, and it does not match the incoming object exactly, broadcast it to listeners of that specific type only.
+4. If the most recent (first in the incoming array) type OR value of that type doesn't match the most recent context broadcast on the channel, broadcast it to un-typed listeners only.
+
+After applying the `connectedAgentsUpdate` message, the newly connected desktop agent and other already connected agents are able to begin communicating through the bridge.
+
```
kriswest, line 295:
```suggestion

#### Step 6. Disconnects

Although not part of the connection protocol, it should be noted that the `connectedAgentsUpdate` message sent in step 5 should also be sent whenever an agent disconnects from the bridge to update other agents. If any agents remain connected, then the `channelState` does not change and can be omitted. However, if the last agent disconnects the bridge SHOULD discard its internal `channelState`, instead of issuing the update.
```

kriswest (approved):
LGTM

Issue #707, opened 2022-05-12
Improve the definition of Instrument context type - add MIC


openfin-johans:
## Enhancement Request

The FDC3 Instrument context type is probably the most widely used and the first one people tend to look at when trying out FDC3 and their implementation or application. The most common identifiers are already standardized but these often requires subscriptions  and many early adopters of FDC3 only use tickers (which are not unique) and therefore include extra information in the context object to help a receiving application to correctly identify the instrument. The Market Identifier Code (MIC) seems to be common although sent using different property names or worse - added to the ticker value itself. This enhancement would simply be to standardize the name used when your app has the capability to handle data specified in the ISO 10383 standard.

https://en.wikipedia.org/wiki/Market_Identifier_Code

### Use Case:
Provide a more accurate representation of a financial instrument using freely available sources and allow for better matching between applications that use disparate data sources.

##### Details

| Property    | Type    | Required | Example Value       |
|-------------|---------|----------|---------------------|
| `type`      | string  | Yes      | `'fdc3.instrument'` |
| `name`      | string  | No       | `'Microsoft'`       |
| `id.ticker` | string  | No       | `'MSFT'`            |
| `id.BBG`    | string  | No       | `'MSFT:US'`         |
| `MIC`    | string  | No       | `'XNAS'`         |

##### Example
```js
const instrument = {
    type: 'fdc3.instrument',
    name: 'International Business Machines',
    id: {
        ticker: 'ibm'
    },
    MIC: 'XNAS'
}
```

### Additional Information
... please add any other information that can provide additional detail for this enhancement request

kriswest:
Seems  fine to me. However, if there are other types of code that identify exchanges we should put this under a `market` or `exchange` field allowing other code types (e.g. Bloomberg Exchange code, EOD etc.) and fields (e.g. country code, name) to be included. 

Hence, how about:
```
const instrument = {
    type: 'fdc3.instrument',
    name: 'International Business Machines',
    id: {
        ticker: 'ibm'
    },
    market: {
       MIC: "XNMS",
       BBG: "UQ",
       name: "NASDAQ Global Market",
       COUNTRY_ISOALPHA2: "US"
    }
}
```

openfin-johans:
[...]

PR #819, opened 2022-09-22
Adding optional market identifiers to the Instrument context

openfin-johans:
Improvement to the Instrument context type as discussed in https://github.com/finos/FDC3/issues/707

### DIFF
--- a/docs/context/ref/Instrument.md
+++ b/docs/context/ref/Instrument.md
@@ -20,9 +20,9 @@ In general, the more identifiers an application can provide, the easier it will 
 - It is valid to include extra properties and metadata as part of the instrument payload, but the minimum requirement
 is for at least one instrument identifier to be provided.
 
-- Try to only use instrument identifiers as intended. E.g. the `ticker` property is meant for tickers as used by an exhange,
-if the identifier you want to share is not a ticker or one of the other standardised ones, rather define
-a property that makes it clear what value it is. This makes it easier for target applications.
+- Try to only use instrument identifiers as intended. E.g. the `ticker` property is meant for tickers as used by an exhange.
+If the identifier you want to share is not a ticker or one of the other standardised fields, define
+a property that makes it clear what value represents. Doing so will make interpretation easier for the developers of target applications.
 
 
 ## Type
--- a/docs/context/ref/Instrument.md
+++ b/docs/context/ref/Instrument.md
@@ -15,12 +15,15 @@ Notes:
 - Not all applications will use the same instrument identifiers, which is why FDC3 allows for multiple to be specified.
 In general, the more identifiers an application can provide, the easier it will be to achieve interoperability.
 
+- The `market` map can be used to further specify the instrument and help achieve interoperability disparate data sources. This is especially useful when using an id that is not globally unique.
+
 - It is valid to include extra properties and metadata as part of the instrument payload, but the minimum requirement
 is for at least one instrument identifier to be provided.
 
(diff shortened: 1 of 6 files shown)


docs/context/ref/Instrument.md
```diff
@@ -15,13 +15,16 @@ Notes:
 - Not all applications will use the same instrument identifiers, which is why FDC3 allows for multiple to be specified.
 In general, the more identifiers an application can provide, the easier it will be to achieve interoperability.
 
+- The `market` map can be used to further specify the instrument and help achieve interoperability disparate data sources. This is especially useful when using an id that is not globally unique.
```
kriswest, line 18:
```suggestion
- The `market` map can be used to further specify the instrument and help achieve interoperability between disparate data sources. This is especially useful when using an `id` field that is not globally unique.
```

Could you add an example (e.g. a `id.ticker` field such as...) or a note on why it might not be globally unique (listed on multiple exchanges)

docs/context/ref/Instrument.md
```diff
@@ -15,13 +15,16 @@ Notes:
...
+
 - It is valid to include extra properties and metadata as part of the instrument payload, but the minimum requirement
 is for at least one instrument identifier to be provided.
 
-- Try to only use instrument identifiers as intended. E.g. the `ticker` property is meant for stock tickers specifically,
-if the identifier you want to share is not a stock ticker, or one of the other standardised ones, rather define
+- Try to only use instrument identifiers as intended. E.g. the `ticker` property is meant for tickers as used by an exhange,
+if the identifier you want to share is not a ticker or one of the other standardised ones, rather define
 a property that makes it clear what value it is. This makes it easier for target applications.
```
kriswest, line 25:
```suggestion
- Try to only use instrument identifiers as intended. E.g. the `ticker` property is meant for tickers as used by an exhange.
If the identifier you want to share is not a ticker or one of the other standardised fields, define
a property that makes it clear what value represents. Doing so will make interpretation easier for the developers of target applications.
```

docs/context/ref/Instrument.md
```diff
@@ -15,13 +15,16 @@ Notes:
...
 
+
```
kriswest, line 27:
```suggestion
```

Markdown lint (single line between sections)

docs/intents/ref/ViewNews.md
```diff
@@ -33,6 +33,9 @@ const instrument = {
     name: 'International Business Machines',
     id: {
         ticker: 'ibm'
```
kriswest, line 35:
```suggestion
        ticker: 'IBM'
```

kriswest (changes requested):
Thanks @openfin-johans this looks pretty good. 

I've added a couple of small suggestions you could commit. The main change required is that this schema file needs updating: https://github.com/finos/FDC3/blob/master/src/context/schemas/instrument.schema.json
Afterwards please run `yarn typegen` which should regenerate the types files from the updated schema. Commit the result to the PR branch - then this should be good to go!

greyseer256 (approved):

kriswest (changes requested):
Just realized this is missing a changelog @openfin-johans. Could you add one (in the unreleased section of https://github.com/finos/FDC3/blob/master/CHANGELOG.md) then this is ready to merge.

openfin-johans:
@kriswest - changelog updated!

kriswest (approved):
LGTM

PR #843, opened 2022-10-28
Clarify docs and comments on Errors and and findIntent error response

### LINKED ISSUE #807: Error cases need better documentation, they are rejected Promises
In a number of instance, API methods are described as returning errors with a given error string.

For example, the description for [open](https://fdc3.finos.org/docs/api/ref/DesktopAgent#open) is as follows:

> If opening errors, it returns an Error with a string from the [OpenError](https://fdc3.finos.org/docs/api/ref/Errors#openerror) enumeration.

The `open` method returns a Promise, as a result, it doesn't return an error, instead, the Promise is rejected with a resultant value.

Notably, a few container implementations don't reject `Error` types, instead they reject with the resultant string value.

### LINKED ISSUE #841: findIntent functions docs need to clarify expected response when no apps are found
### Area of Issue
[x] API

### Issue Description: 
The documentation for the `findIntent` functions ([`findIntent`](https://fdc3.finos.org/docs/api/ref/DesktopAgent#findintent) and [`findIntentByContext`](https://fdc3.finos.org/docs/api/ref/DesktopAgent#findintentsbycontext)) are not very clear on the expected response when no apps are found for an intent or no intents/apps are found for a context.

The `raiseIntent` functions are a little clearer on whats expected:

> If a target app for the intent cannot be found with the criteria provided or the user either closes the resolver UI or otherwise cancels resolution, an Error with a string from the [ResolveError](https://fdc3.finos.org/docs/api/ref/Errors#resolveerror) enumeration is returned.

I.e. An error with the message defined by `ResolveError.NoAppsFound` should be returned (rejected) by the promise.

Whereas for `findIntent` this is less clear:

> If the resolution fails, the promise will return an Error with a string from the [ResolveError](https://fdc3.finos.org/docs/api/ref/Errors#resolveerror) enumeration.

With one participant in a [Desktop Agent Bridging discussion group meeting](https://github.com/finos/FDC3/issues/839) commenting that you could also return an `AppIntent` (or `AppIntent[] for `findIntentByContext`) with an empty `apps` element instead.

While either approach is workable (apart fro the fact you can't get display data for the intent needed to populate the `AppIntent` without at least one app being returned - although we are due to deprecate the problematic `displayName` property anyway), I don't believe this is the correct interpretation of the current standard's docs (rather the response should be consistent with `raiseIntent` functions). Allowing for multiple interpretations will promote variation in behavior amongst implementations and difficulties for conformance testing, which is undesireable.

Hence, I propose we clarify the documentation to make clear that `ResolveError.NoAppsFound` should be returned.

kriswest:
resolves #807
resolves #841

Clarifies how errors should be returned (reject promise with JS Error object, using message from enumeration) throughout docs and source comments in 2.0. Includes clarification that `findIntent` fns should error if no apps are found

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -97,6 +97,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 * Clarified the description of the addContextListener functions from the Desktop Agent and Channel APIs in spec and docs. ([#492](https://github.com/finos/FDC3/pull/492))
 * Clarified that implementing `fdc3.getInfo()` is required for compliance with the FDC3 standard ([#515](https://github.com/finos/FDC3/pull/515))
 * Corrected syntax errors in valuation schema ([#834](https://github.com/finos/FDC3/pull/834))
+* Clarified that API errors are promises rejected with a JavaScript (or language appropriate) Error Object with a message chosen from the error enumerations. ([#843](https://github.com/finos/FDC3/pull/843))
+* Clarified that `findIntent` functions should return the `ResolveError.NoAppsFound` error, rather than an `AppIntent` with an empty `apps` array when no apps are fund during intent resolution. ([#843](https://github.com/finos/FDC3/pull/843))
 
 ## [npm v1.2.0] - 2021-04-19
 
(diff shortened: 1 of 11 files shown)


greyseer256 (approved):

robmoffat (approved):
very thorough!

mattjamieson (approved):

PR #847, opened 2022-11-03
846 Corrected schema definition for appD interop.intents.listensFor element

### LINKED ISSUE #846: Example and Definition Don't Match For App Directory `listensFor`
### Area of Issue
- [x] App Directory

### Issue Description: 

The specification defines `listensFor` like this:

```
       listensFor:
              type: object
              description: | 
                A mapping of Intents names that an app listens for via `fdc3.addIntentListener()` to their 
                configuration. 
                
                Used to support intent resolution by desktop agents. Replaces the `intents` element used in appD records prior to FDC3 2.0. 
              additionalProperties:
                x-additionalPropertiesName: Intent name
                items:
                  $ref: '#/components/schemas/Intent'
```

According to https://github.com/OAI/OpenAPI-Specification/issues/668#issuecomment-218829120 this implies that the map listens for will have as values an _array_ of intents.  When I generate typescript bindings for this structure I get:

```
    listensFor?: { [key: string]: components['schemas']['Intent'][] };
```

(note the trailing `[]`)

However, in the example, it appears not to be an array:
 
```
"interop": {
  "intents": {
    "listensFor": {
      "ViewChart": { 
        "displayName": "View Chart",
        "contexts": []
      }, 
      "myApp.GetPrice": {}
    },
    "raises": {}
```

(Taken from https://fdc3.finos.org/schemas/2.0/app-directory#tag/Application/paths/~1v2~1apps~1%7BappId%7D/get)

Was the example validated against the schema?

### Additional Context:

Also seems to be true for `raises`, and possibly other things I haven't checked.

kriswest:
resolves #846
appD schema contained an error that turned the `interop.intents.listensFor` element values (keyed on intent name) into an array of intents instead of a single intent definition - hence it did not match the example (which is correct according to the issue that introduced this element).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -97,6 +97,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 * Clarified the description of the addContextListener functions from the Desktop Agent and Channel APIs in spec and docs. ([#492](https://github.com/finos/FDC3/pull/492))
 * Clarified that implementing `fdc3.getInfo()` is required for compliance with the FDC3 standard ([#515](https://github.com/finos/FDC3/pull/515))
 * Corrected syntax errors in valuation schema ([#834](https://github.com/finos/FDC3/pull/834))
+* Corrected schema definition for appD `interop.intents.listensFor` element ([#847](https://github.com/finos/FDC3/pull/847))
 
 ## [npm v1.2.0] - 2021-04-19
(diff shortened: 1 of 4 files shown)


greyseer256 (approved):

robmoffat (approved):
I tested this out locally, lgtm

mattjamieson (approved):

Issue #683, opened 2022-04-28
Standard WG Meeting - April 28, 2022


kriswest:
## Date
Thursday 28 April 2022 - 10am EST / 3pm GMT

### WebEx info
- [Meeting Link](https://finos.webex.com/finos/j.php?MTID=ma4b1de1b9f691d89e4b2ff380ec9687c)
Meeting number: 665 568 411

*More ways to join*
- Join by video system:
   - Dial 665568411@finos.webex.com
   - You can also dial 173.243.2.68 and enter your meeting number
- Join by phone
   - +1-415-655-0003 US Toll
   - +44-20319-88141 UK Toll
- Access code: 665 568 411

## Meeting notices
- FINOS **Project leads** are responsible for observing the FINOS guidelines for [running project meetings](https://github.com/finos/community/blob/master/governance/Meeting-Procedures.md#run-the-meeting). Project maintainers can find additional resources in the [FINOS Maintainers Cheatsheet](https://odp.finos.org/docs/finos-maintainers-cheatsheet/).

- **All participants** in FINOS project meetings are subject to the [LF Antitrust Policy](https://www.linuxfoundation.org/antitrust-policy/), the [FINOS Community Code of Conduct](https://github.com/finos/community/blob/master/governance/Code-of-Conduct.md) and all other [FINOS policies](https://github.com/finos/community/tree/master/governance#policies).

- FINOS meetings involve participation by industry competitors, and it is the intention of FINOS and the Linux Foundation to conduct all of its activities in accordance with applicable antitrust and competition laws. It is therefore extremely important that attendees adhere to meeting agendas, and be aware of, and not participate in, any activities that are prohibited under applicable US state, federal or foreign antitrust and competition laws. Please contact legal@finos.org with any questions.

- FINOS project meetings may be recorded for use solely by the FINOS team for administration purposes. In very limited instances, and with explicit approval, recordings may be made more widely available.

## Agenda
- [x] Convene & roll call, review meeting notices (5mins)
- [x] Review action items from previous meeting (5mins)
  - #637
- [x] Updates on timeline and outstanding projects for FDC3 2.0 (10 mins)
  - Timeline
    - proposed deadline for pull requests: May SWG meeting
    - proposed fdc3 pre-draft adoption vote: June SWG meeting 
  - #412 
  - Formal specification project
    - #679
      - Call for implementation (app and platform provider) details
    - #549
  - Contexts & Intents discussion group proposals 
  - AppD discussion group proposals
[...]

Issue #511, opened 2021-11-23
Standardise auto-receipt of current context on join/addContextListener in DesktopAgent and Channel


kriswest:
## Enhancement Request
According to the FDC3 API spec:
> When an app joins a channel it will automatically recieve the current context for that channel.

https://fdc3.finos.org/docs/api/spec#joining-channels

We take that to mean that the current context of the channel will be auto-broadcast to the app when it joins the channel with context listeners already set (last context for an un-typed listener, last context of the specific type for a type listener), or when it adds a context listener.

_n.b. assume a context has already been broadcast to the channel in each case below_
```javascript
fdc3.addContextListener(null, console.log);
fdc3.joinChannel("someSystemChannel");
// last context broadcast to the channel should be logged immediately
fdc3.addContextListener("fdc3.contact", console.log);
// last fdc3.contact context broadcast to the channel should be logged immediately
```

This is not the case for Channels, where you have to manually retrieve the context
```javascript
let channel = await fdc3.getOrCreateChannel("someChannel");
channel.addContextListener(null, console.log);
//nothing is received initially until a new context is broadcast
let context = await channel.getCurrentContext();
```

We've found that this difference can be confusing for developers using the API. It also introduces complexity in implementing the DesktopAgent/Channel API that may be unnecessary (and the requirement is easy to miss in the spec). 

To resolve, we suggest that the current context of a channel should be received automatically when adding a context listener and that this detail should be added to the documentation of the `addContextListener` functions (in both APIs).

kriswest:
Agreed to close at meeting:
- #683 
While inconsistency is bad, auto-broadcast on user channels is popular there are use cases for not receiving the current context on adding a listener to an app channel (and while an app can retrieve current context it can't stop itself doing so if we make this change, preventing implementation of those use cases).

PR #864, opened 2022-11-28
Confirm that App channels do not re-play past contexts when a listener is added

kriswest:
Small addition to one of the app channel tests - we need to confirm that App Channels, listened to via the `Channel` interface do not auto-replay context messages (as System/User channels listened to through the fdc3/Desktop Agent interface do). 

A different PR will add this to the 2.0 test cases.

Background:

The Standards Working Group re-affirmed under issue #511 at meeting #683 that the difference in the behaviour of the `fdc3`/`DesktopAgent` and `Channel` interfaces was intentional and necessary to implement certain use cases with App Channels. It was decided that this would not change in FDC3 2.0, hence it remains the case in 2.0.


robmoffat:
Added here: https://github.com/finos/FDC3/blob/conformance2.0-rm/toolbox/fdc3-conformance/App-Channel-Tests.md

Also added some code to be consistent with other steps. 

Can you review and close this PR if it's good?

kriswest:
>Added here: https://github.com/finos/FDC3/blob/conformance2.0-rm/toolbox/fdc3-conformance/App-Channel-Tests.md
>
> Also added some code to be consistent with other steps.

Looks good to me! Closing this PR

Issue #852, opened 2022-11-11
fdc3.ChatInitSettings is missing id and type is incorrectly defined


bingenito:
## Minor Issue
The ChatInitSettings context type is missing the id property and the type property is mistyped as any instead of string

### Area of Issue
[x] Context Data

kriswest:
Will be fixed by #869

Issue #855, opened 2022-11-11
Question: Should ChatInitSettings options have a exported interface


bingenito:
### Question Area
[x] Context Data

### Question
The documentation for fdc3.chatinitsettings has 5 documented optional boolean properties but the exported interface for ChatInitSettings defines options as any.  Should we add the documented properties to an exported type and change options to be that type vs any?

symphony-jean-michael:
Hi @kriswest 
Thank you for your investigation.
I'v created a new PR: https://github.com/finos/FDC3/pull/875
Could you please review it ?
Thanks

PR #869, opened 2022-11-30
make chatInitSettings inherit from Context

Yannick-Symphony:
fixes https://github.com/finos/FDC3/issues/852

### DIFF
--- a/src/context/schemas/chatInitSettings.schema.json
+++ b/src/context/schemas/chatInitSettings.schema.json
@@ -3,10 +3,8 @@
     "$id": "https://fdc3.finos.org/schemas/next/chatInitSettings.schema.json",
     "type": "object",
     "title": "ChatInitSettings",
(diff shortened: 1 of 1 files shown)


mistryvinay (approved):
Looks good to me.

mistryvinay:
hi @Yannick-Symphony just one more update required for this PR.  Currently there is the use of the reserved word `public` in the schema.  This is causing issues when the schema is being built (#869)

`            "public": "boolean",`

Could we change this to something else, maybe `isPublic`?

symphony-jean-michael:
Hi @mistryvinay 
It seems I already fixed the 'public' reserved word in the other PR https://github.com/finos/FDC3/pull/875 (Fixing bug https://github.com/finos/FDC3/issues/855)
So, there is no need to include it in this PR 😉 
(cc @Yannick-Symphony )

kriswest (approved):
LGTM

Issue #872, opened 2022-12-02
Clarify handling of malformed context objects by Desktop Agents


kriswest:
### Area of Issue
[x] API

### Issue Description: 
The Desktop Agent API definitions in TypeScript specify that its functions accept `Context` objects (at their simplest, JSON/Javascript Objects that have a `type` field). However, the Standard does not specify how they should respond if passed a malformed Context (i.e. one that is not an Object or one that lacks a type field). 

Generally, other error responses are defined already and functions are restricted to returning Javascript (or platform-appropriate) Error objects with messages drawn from a specific [error enumeration](https://fdc3.finos.org/docs/api/ref/Errors), e.g.:

`fdc3.open`:
>If an error occurs while opening the app, the promise MUST be rejected with an Error Object with a message chosen from the [OpenError](https://fdc3.finos.org/docs/api/ref/Errors#openerror) enumeration.

`fdc3.raiseIntent`
>The Desktop Agent MUST reject the issuing app's promise, with a string from the [ResultError](https://fdc3.finos.org/docs/api/ref/Errors#resulterror) enumeration, if: (1) the intent handling function's returned promise rejects, (2) the intent handling function doesn't return a promise, or (3) the returned promise resolves to an invalid type.

Due to this lack of defined behaviour, it is not currently possible to test Desktop Agents for consistent behaviour. Hence, test cases related to this (that currently exist in the conformance framework) also need to be removed.

Future versions of the FDC3 Standard should specify how malformed Context Objects should be handled and (where necessary) define error messages that should be returned.

kriswest:
We should probably define an error to be returned by all of `fdc3.broadcast`, `channel.broadcast`, `fdc3.open`, `fdc3.raiseIntent`, `fdc3.raiseIntentForContext`, `fdc3.findIntent`, `fdc3.findIntentByContext`. Doing so will ensure that the developer/app submitting the malformed context is informed of the error. I don't think we need to do anything on the receiving end (addContextListener etc.) is malformed context is not submittable.

PR #873, opened 2022-12-02
Remove malformed context test case as correct behavior is not defined

kriswest:
The FDC3 Standard does not define the expected behavior on receipt of a malformed context (by `fdc3.open`, or indeed any other function). Hence, it is not possible to adequately define test cases for it. There currently only exists a test case for `fdc3.open` which this PR will remove.

See issue https://github.com/finos/FDC3/issues/872 for a request to refine this FDC3 so that we can test it in a future version.


ggeorgievx (approved):

robmoffat (approved):

nkolba (approved):

openfin-johans (approved):
yank it!

PR #876, opened 2022-12-06
Add missing createPrivateChannel and findInstances fns to Methods.ts

kriswest:
The NPM module is missing an a function for createPrivateChannel. This PR adds the function and rolls the pre-release version number so that the NPM module will release itself


kriswest:
Muchas gracias @mattjamieson !

PR #924, opened 2023-03-09
Ensuring FDC3 API types use string rather than String

### LINKED ISSUE #915: API types use String rather than string
### Area of Issue
[x] API

### Issue Description: 
A number of deprecated functions (`open`, `raiseIntent`, `raiseIntentForContext`) use `String` Objects rather than `string` literals, sometimes mixing the types. These should be standardized on `string` literals.

### Additional Context:
Supercedes #883

kriswest:
resolves #915
A number of Desktop Agent function arguments were incorrectly typed as `String` (object) rather than `string` (literal). This would have caused occasional typing headaches and the need for casts. However, the fix should be non-breaking (other than making casts unnecessary and perhaps triggering lint rules, which is fine as they can be removed after the change).


greyseer256 (approved):

mattjamieson (approved):
LGTM!

PR #926, opened 2023-03-09
Deprecating IntentMetadata.displayname

### LINKED ISSUE #312: The `AppIntent` interface incorrectly describes an intent

kriswest:
resolves #312

Marking IntentMetadata.displayName as deprecated as it can only come from app definitions in appD, which means the DA needs to pick one and all could be different. SWG agreed to go with the intent name instead as thats always the same AND is required to be recognizable


website/static/schemas/next/app-directory.yaml
```diff
@@ -607,7 +607,8 @@ components:
       properties:
         displayName:
           type: string
-          description: An optional display name for the intent that may be used in UI instead of the name.
+          description: Optional display name for the intent. Deprecated in favour of the Intent name, which is common amongst all apps that support it, where teh display name may vary.
```
greyseer256, line 610:
```suggestion
          description: Optional display name for the intent. Deprecated in favour of the Intent name, which is common amongst all apps that support it, where the display name may vary.
```

src/app-directory/specification/appd.yaml
```diff
@@ -607,7 +607,8 @@ components:
       properties:
         displayName:
           type: string
-          description: An optional display name for the intent that may be used in UI instead of the name.
+          description: Optional display name for the intent. Deprecated in favour of the Intent name, which is common amongst all apps that support it, where the display name may vary.
```
greyseer256, line 610: Should these usages of `intent` have the same capitalization? Is the capitalized one referencing the Intent concept? Sorry I'm still kind of fuzzy on when capitalization is appropriate for some of these words.

docs/api/ref/Metadata.md
```diff
@@ -302,9 +302,10 @@ interface IntentMetadata {
   /** The unique name of the intent that can be invoked by the raiseIntent call. */
   readonly name: string;
 
-  /** A friendly display name for the intent that should be used to render UI
-   *  elements.
-   */
+  /** Display name for the intent. Deprecated in favour of using the intent name for display as
+   *  display name may vary for each application as it is defined in the app's AppD record. 
+   *  @deprecated
+  */
```
mattjamieson, line 308:
```suggestion
  /** Display name for the intent.
   * @deprecated Use the intent name for display as display name may vary for
   * each application as it is defined in the app's AppD record.
   */
```

src/app-directory/specification/appd.yaml
mattjamieson: I think it should be lowercase...

website/static/schemas/next/app-directory.yaml
mattjamieson, line 610:
```suggestion
          description: Optional display name for the intent. Deprecated in favour of the intent name, which is common amongst all apps that support it, where the display name may vary.
```

src/app-directory/specification/appd.yaml
kriswest: Lets go with lowercase - its that way in the glossary and I think appropriate in all cases other than when we are specifically discussing the concept of Intents.

 ```suggestion
          description: Optional display name for the intent. Deprecated in favour of the intent name, which is common amongst all apps that support it, where the display name may vary as it is defined in the app's AppD record.
```

File: website/static/schemas/next/app-directory.yaml (hunk 1 above)
Comment (kriswest) [reply website/static/schemas/next/app-directory.yaml:610]:
```suggestion
          description: Optional display name for the intent. Deprecated in favour of the intent name, which is common amongst all apps that support it, where the display name may vary as it is defined in the app's AppD record.
```

src/app-directory/specification/appd.yaml
```diff
@@ -607,7 +607,8 @@ components:
       properties:
         displayName:
           type: string
-          description: An optional display name for the intent that may be used in UI instead of the name.
+         description: Optional display name for the intent. Deprecated in favour of the intent name, which is common amongst all apps that support it, where the display name may vary as it is defined in the app's AppD record.
```
kriswest, line 610:
```suggestion
          description: Optional display name for the intent. Deprecated in favour of the intent name, which is common amongst all apps that support it, where the display name may vary as it is defined in the app's AppD record.
```

kriswest:
Changes applied - needs signing off again @mattjamieson @greyseer256

kriswest:
@greyseer256 this'll need your review before its merged

greyseer256 (approved):
👍

PR #929, opened 2023-03-15
Conformance2.0 rm updates

greyseer256:
Consistent capitalization of NOTs to emphasize negation. Also drops one unsubscribe assertion that is covered in the basic tests.


robmoffat:
awesome, thanks @greyseer256!

PR #951, opened 2023-04-12
Quicktype dependency update + release 2.0.0 npm module

kriswest:
Changes:
- Updating to latest (non-vulnerable) Quicktype version
- Switch to use of local files for type generation (with workaround for Quicktype bug)
- Altering Context schema to apply typing to unevaluatedProperties to avoid conflicts with subschemas
- Regenerating contexts types
- Setting npm module version to 2.0.0

Maintainers slack channel chat relating to this PR: https://finos-lf.slack.com/archives/C0296SHSY1J/p1681407013513619

> so... additionalProperties doesn't allow to specify anythign about keys, only value types.
>
> However, I have an answer! It was our schemas and this is a well known problem detailed in the JsonSchema docs: https://json-schema.org/understanding-json-schema/reference/object.html#extending-closed-schemas
> 
> The answer is to use unevaluatedProperties: https://json-schema.org/understanding-json-schema/reference/object.html#unevaluated-properties
>
> This alternative keyword is aware of sub-schemas and only applies to properties not defined anywhere. 
> Hence, applying it to context:
```
"id": {
      "type": "object",
      "unevaluatedProperties": { "type": "string" }
    }
```
>will ensure that any unstandardized field names in id are restricted to having string values.

> quicktype pretty much ignores it in generated typescript, so what I'm getting out is:
```ts
export interface Instrument {
  id: StickyId;
  market?: FluffyMarket;
  type: string;
  name?: string;
  [property: string]: any;
}

export interface StickyId {
  BBG?: string;
  CUSIP?: string;
  FDS_ID?: string;
  FIGI?: string;
  ISIN?: string;
  PERMID?: string;
  RIC?: string;
  SEDOL?: string;
  ticker?: string;
  [property: string]: any;
}

export interface FluffyMarket {
  BBG?: string;
  COUNTRY_ISOALPHA2?: string;
  MIC?: string;
  name?: string;
  [property: string]: any;
}
```
> i.e.:
> - the generated typescript isn't applying the restriction (rather its still doing default additionalProperties: true)... 
> - but the schema will apply it to JSON validation at least and 
> - the typescript types are not totally invalid/will compile
> - the field names make it into the generated type.
>
> I think this is as good as we will achieve, without replacing or fixing quicktype (its source isn't easy to grok...). I also believe this is compatible with the current version of the standard (in fact slightly more correct than current state).
>
> Current quicktype version has (many) issues with -nice-property-names but may improve in future.
>
> I've also figured out how to (work around quicktypes bugs and) compile the types from local files, which I've added to the PR. For now it requires a node script (due to daft quicktype bug with CLI options). Will remove that when its no longer needed. I do NOT wnat to continue to generate from the published URLs as this makes development really really difficult.
>
> if anyone wants to fiddle with this further, check out my branch from the pr and run npm run typegen (and it'll use your local files for the generation).


src/context/ContextTypes.ts
```diff
@@ -22,29 +22,26 @@
...
+export interface InstrumentElement {
+  id: PurpleId;
+  market?: PurpleMarket;
   type: string;
   name?: string;
+  [property: string]: any;
 }
 
-export interface InstrumentID {
+export interface PurpleId {
```
rikoe, line 44: We can't have types like "PurpleId" or "TentacleId" (below).

I had similar issues with QuickType when I first used it - one has to use the settings in such a way so that it accurately names the types after the schema types.

I can't remember exactly what caused this in the past, but there was a way to get around it. It's a pain, I know.

src/context/ContextTypes.ts
```diff
@@ -194,21 +293,14 @@ export interface Valuation {
   type: string;
   valuationTime?: Date;
   value: number;
-  id?: { [key: string]: string };
+  id?: { [key: string]: any };
```
rikoe, line 296: So we are okay with the `any` here? Not a fan, as you know.

src/context/ContextTypes.ts
kriswest: We'll need an answer as to how to achieve that as I can't figure out the arguments needed.

I think that the statement:

> We can't have types like "PurpleId" or "TentacleId" (below).

Isn't actually true however. These type names are only generated for the purposes of referencing within the types, externally you would only use the top-level type named after teh schema, e.g. Instrument, and when type checking it will use/compare the content of the type rather than the the name.

However, I would rathr it was generating InstrumentID, rather than PurpleId.

I would also prefer that it was just genearating Instrument and not also InstrumentElement. Not sure why its done that. However, the type checking it will do based on these is correct

src/context/ContextTypes.ts
kriswest: If we're not we can't use quicktype + Typescript anymore. Now that the indexSignature is inherited by subtypes (specific context types like Instrument), it conflicts with their option properties and the generated types don't compile (the same occurs if the subtype expresses `additionalProperties` instead, as it generates the same index signature). Without the index signature the type can't be extended (due to excess property checking). 

This limitation *only* affects the TypeScrpt types. Specified id types remain typed (so id.BBG has to be a string) but if you made one up its allowed to be `any` by the typescript types generated (so `id.madeUpId = true` ends up valid according to Typescript) - however the JSONSchema (which actually handles the `unevaluatedProperties` - where quicktype ignores it/its not expressible in Typescript easily) would mark that as invalid.

In a hand produced type we could do this:

`id?: { [key: "BBG" | "ticker" | string]: string };`  (filling in all the defined ID types)

However, I'm not sure we can express that in JSONSchema such that it will generate as additionalProperties only affects values not the keys (the key type comes from the JSON spec, I think, where they are always required to be strings). I'll have a poke at the JSONschema docs to double check that's not possible - but don't think it is. Unless I turn up something, I think this proposal is the best we can do.

kriswest: patternProperties could have a regex defining the `id` field names, but also doesn't generate typescript. JSONSchema doesn't have an undefined type so we can't use that to make the index signature:

`id?: { [key: string]: string | undefined };`

which would not conflict with the optional properties. 

Hence, I don't see another way forward other than:
- accept the change in this PR which allows *unstandardized* id field properties to be `any` (standardized ones are still `string`)
- contribute a modification to quicktype causing it to generate `id?: { [key: string]: string | undefined };` when `additionalProperties: {type: string}` is expressed in a schema, and work with a fork until they merge it. I'm not sure this would be correct in all cases and hence whether they would accept it... 
- continue with out of date quicktype with vulnerable dependencies and types that can't be extended.

rikoe (approved):
Approved as discussed on fdc3 maintainers Slack. Will address issues with QuickType generation in future.

mattjamieson (approved):
Agree with @kriswest @rikoe discussion. We should live with this for now to get it out and then rethink the approach.

Issue #921, opened 2023-03-03
Address Content and Accessability issues in FDC3 website


robmoffat:
Address issues identified by @julianna-ciq  in PR comments:
- https://github.com/finos/FDC3/pull/910

julianna-ciq:
All issues identified in #910 have been addressed and merged. This issue can be closed.

PR #954, opened 2023-04-12
docs: (Stylistic) Move paragraph list into html list

julianna-ciq:
Part of the #921 umbrella

Before PR:
image

After PR:
image

I changed the paragraph-based list to using native HTML list controls. This change is stylistic.


kriswest (changes requested):
A worthwhile change, but it also needs to be made in the main docs source here: https://github.com/finos/FDC3/blob/master/docs/api/ref/DesktopAgent.md
or it will regress with the next version.

kriswest (approved):
LGTM

PR #950, opened 2023-04-12
docs: a11y: Add alt text to images on front page

julianna-ciq:
In this PR, I:
- [x] Added empty alt text to icons
- [x] Removed the alt text "Features" from an icon
- [x] Added alt text to the Twitter and Github links in the footer
- [x] Added alt text "FDC3" to the logo image
- [x] Added alt text "FDC3 1.2/2.0 Certified" to those badges

If you want to test the impact of these changes, here's how you can quickly see the difference using Chrome:
1. Go to the webpage (in Chrome)
2. Open Developer Tools (Win: press F12)
3. In DevTools, go to the "Lighthouse" tab
4. Click "Analyze page load" button

Before this PR, you'll see a score in the 80's, with an indicator that there are several images without alt text
image
**Lighthouse score for fdc3.finos.org**

After this PR, you'll see a score in the 90's, without any indicators about alt text
image
**Lighthouse score for dev version of the website, based on this PR**

PR #955, opened 2023-04-13
Remove vestigial twitter badge from FINOS footer

julianna-ciq:
From @kriswest in #950 :
>  I don't believe the maintainers have access to this twitter account and theres next to nothing on it. It needs to go.

So, here's a quick PR to just remove it from the footer while we're talking about it.

Before PR:
image

After PR:
image


kriswest (approved):
LGTM

Issue #943, opened 2023-04-07
Website a11y: Front page: Headings are out of order


julianna-ciq:
## Minor Issue
Please use minor issues for discrepancies or errors found in the spec or supporting documentation, examples, and other materials.  If you wish to propose a new feature or a revision to an existing feature, please use the *Enhancement Request* template.

### Area of Issue
[ ] App Directory
[ ] API
[ ] Context Data
[ ] Intents
[ ] Use Cases
[x] Other

### Issue Description: 
Headings are supposed to go in order and not skip around. Also, it's best practices to only have 1 H1 element.

The current headings order is:
* H2: Open standards for the financial desktop
* H2: API
* H2: Intents
* H2: Context Data
* H2: App Directory
* H1: Use Cases
* H1: Who is Using FDC3?
* H1: FDC3 Conformance
* H3: FDC3 Sail
* H3: Finsemble
* H3: Glue42 Enterprise
* H3: OpenFin

That's several H1's, and the levels are jumping around (2, then 1, then 3).

Alternatively, I'd recommend this organization for headings:
* H1: FDC3
* H2: Use Cases
* H2: Who is Using FDC3?
* H2: FDC3 Conformance
* H3: FDC3 1.2 Certified
* H3: FDC3 2.0 Certified

Note that this wouldn't involve any visual changes. The headings changes would be purely semantic.

### Additional Context:
Organized headings are most helpful for screen reader users. When using a screen reader like JAWS or NVDA, you can press the "h" to jump around to headings, or number keys like "2" to jump to headings of that level. This helps a user build a mental model of the website, as well as framing what is or isn't important. Additionally, if you ever conduct an accessibility audit, skipped headings are one of the first things that will be flagged.

PR #959, opened 2023-04-17
Make grid blocks images/labels clickable

julianna-ciq:
Part of #943 

In this PR, I:
- [x] Removed the H2 tags from the "API", "Intents", "Context Data", and "App Directory" blocks.
- [x] Made the images and labels into one large clickable block
- [x] Remove the links from the descriptions (since the images+labels are now links)

Before:
image

After:
image

This is what a new block looks like with a focus indicator:
image

(The focus indicator is only visible if you use keyboard navigation to tab to the element on the page. If a mouse user simply clicks on the image/label, you'll see an underline under the text)

### DIFF
--- a/website/data/features/api.mdx
+++ b/website/data/features/api.mdx
@@ -1,5 +1,8 @@
-<img alt="" src="/img/feature-api.svg" />
+<a href="/docs/api/spec">
+    <div>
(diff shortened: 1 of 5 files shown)


kriswest (approved):
LGTM, thanks again!

PR #960, opened 2023-04-18
docs: Address all console errors on the front page

julianna-ciq:
If you launch the FDC3 website in dev mode, you get webpack/react warnings like this:
Warning: Each child in a list should have a unique "key" prop

There were 2 cases where a `key` prop being needed, and a couple of cases of `class` being used instead of `className`.

Now, there are no errors/warnings on the home page (in dev mode).


robmoffat (approved):

kriswest (approved):
LGTM

PR #972, opened 2023-04-28
Clarify malformed context handling

### LINKED ISSUE #872: Clarify handling of malformed context objects by Desktop Agents

kriswest:
resolves #872
Adds the relevant error to the ChannelError, OpenError and Resolve enumerations + connects in their docs where appropriate.

### DIFF
--- a/docs/api/ref/Channel.md
+++ b/docs/api/ref/Channel.md
@@ -138,6 +138,8 @@ Channel implementations should ensure that context messages broadcast by an appl
 
 If you are working with complex context types composed of other simpler types (as recommended by the [FDC3 Context Data specification](../../context/spec#assumptions)) then you should broadcast each individual type (starting with the simpler types, followed by the complex type) that you want other apps to be able to respond to. Doing so allows applications to filter the context types they receive by adding listeners for specific context types.
 
+If an application attempts to broadcast an invalid context argument the Promise returned by this function should reject with the [`ChannelError.MalformedContext` error](Errors#channelerror).
+
 #### Example
 
 ```javascript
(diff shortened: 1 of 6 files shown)


greyseer256 (approved):
👍

mattjamieson (approved):
LGTM!

PR #973, opened 2023-04-28
827 Adding appD error route examples

### LINKED ISSUE #827: AppD v2 routes don't define error response samples
## Minor Issue
~AppD v2 routes don't define error responses in their OpenAPI definitions. The endpoint for retrieving a single appD record should at least define a 404 to be returned when the appId is not found.~

AppD v2 routes don't include error response samples, which should be added to the OpenAPI definitions to make the expected error responses clearer.

### Area of Issue
[x] App Directory

kriswest:
resolves #827 
Also corrects a few spelling errors in the appD schema


greyseer256 (approved):
:+1:

mattjamieson (approved):
LGTM!

PR #981, opened 2023-05-11
980 Correcting appd interop.appChannels to use id instead of name

### LINKED ISSUE #980: appD `interop.appChannels` uses channel `name` instead of `id`
### Area of Issue
[x] App Directory

### Issue Description: 
appD's `interop.appChannels` refers to channel by `name` instead of `id`, it shoudn't.

### Additional Context:
When the new `interop` element was added to the appD record it used to `name` in the `appChannels` element rather than `id`. This was an inadvertent error as channels should be referred to by their `id`, rather than the optional `displayMetadata.name`. There is little or no functionality based on this metadata in the FDC3 Desktop Agent, so its recommended that this is corrected ASAP.

kriswest:
resolves #980 

Swaps out `interop.appChannels.name` for `interop.appChannels.id` - use of `name` was inadvertent and confusing (as channels have an `id` and an optional `displayMetadata.name`)


mattjamieson (approved):
LGTM

rikoe (approved):

greyseer256 (approved):

PR #982, opened 2023-05-11
Add deprecation tags to customConfig entries in appD schema

### LINKED ISSUE #900: Question: customConfig vs hostManifests in v2
### Question Area
[x] App Directory
[ ] API
[ ] Context Data
[ ] Intents
[ ] Use Cases
[ ] Other

### Question

With the introduction of `hostManifests` property what is the purpose of the `customConfig` property from now on?

[The API doc of `customConfig`  states](https://fdc3.finos.org/schemas/2.0/app-directory#tag/Application/paths/~1v2~1apps~1%7BappId%7D/get):
> Array of objects (NameValuePair)  An optional set of name value pairs that can be used to deliver custom data from an App Directory to a launcher.

However, this seem to have the same purpose as the hostManifests property because the hostManifests can now too contains launcher specific configuration/metadata entries.

From my point of view I always found the customConfig property a bit odd because it was felt like bag of launcher specific metadata that could easily conflict between each others if 2 launchers for instance rely on a customConfig property of the same name. The `hostManifest` feels like a much more elegant solution at replacing that kind of use case.

kriswest:
resolves #900

Sets the deprecated flag on customConfig elements in the appD spec:
image

prepares for #1006 expected post FDC3 2.1.


mattjamieson (approved):

kriswest:
Revived at #1013 (to be included in FDC3 2.1) as prep for:
- #1006 
which will use a different named element (likely to be `applicationConfig`) to better reflect its usage.

kriswest:
@greyseer256 this deprecation is ready for your review now.

hughtroeger (approved):

PR #985, opened 2023-05-11
updating definition of otherConfig in fdc3.chart to use an array of context objects

### LINKED ISSUE #856: Tighter definition for "otherConfig" in fdc3.chart context type
### Use Case:
The `otherConfig` section of the _fdc3.chart_ context type message allows additional directives to be sent to the resulting chart, such as indicators and annotations.  As it stands now, all target applications that acknowledge `otherConfig` will attempt to digest its contents without actually knowing if it is intended for them.  This can cause unintended consequences when a message tailored for one app is misinterpreted by another app.

### Workflow Description
Rather than looking like this:
```js
otherConfig: {
        indicators: [
            {
                name: "ma",
                parameters: {
                    period: 14,
                    type: "ema"
                }
            },
            {
                name: "volume"
            }
        ]
}
```
the otherConfig section ought to look like this:
```js
otherConfig: {
        configs: [
            {
                type: "[vendor.product.indicator]"
                name: "ma",
                parameters: {
                    period: 14,
                    type: "ema"
                }
            },
            {
                type: "[vendor.product.indicator]"
                name: "volume"
            }
        ]
}
```
In this way, different products can digest only the configurations pertaining to them.

### Workflow Examples
This block will send the same indicator to different vendors who process them differently:
```js
otherConfig: {
	configs: [
		{
			type: "somevendor.someproduct.indicator",
			name: "stddev",
			parameters: {
				period: 10,
				matype: "exponential"
			}
		},
		{
			type: "someothervendor.someotherproduct.formula",
			formula: "standard-deviation",
			fields: {
				lookback: 10,
				type: "ema"
			}
		}
	]
}
```

### Additional Information
Schema change:
```js
"otherConfig": {
	"type": "object"
}
```
changes to:
```js
"otherConfig": {
	"type": "object",
	"additionalProperties": { 
		"type": "array",
		"items": {
			"$ref": "context.schema.json#"
		} 
	}
}
```

kriswest:
resolves #856 

Changes the definition of `otherConfig` in fdc3.chart from an unstandardized object to an array of Context objects. This allows the type of each additional piece of config to be checked and either used or ignored by a charting application, based on its type.

@rbruckheimer


rbruckheimer:
Is there a concern for backwards compatibility here?

kriswest:
@rbruckheimer AFAIK ChartIQ is the main user of this type (although my information on that is not perfect), so you'd know better than I. Do you have clients using it currently? 

I wouldn't propose a change like this for an element that was standardized, however, this field was specifically for unstandardized data. A backwards-compatible change, that also doesn't have a redundant layer would be to use an object and create fields under it which are arrays, but divide them up by type, e.g. indicators, drawings, formulae, other (so that they have a purpose in organizing the config). OR you could keep `otherConfig` as a deprecated field and create a `proprietaryConfig` or `customConfig` field to supercede it. OR just live with that redundancy (`otherConfig.config`). 

Writing it out, seems like either the breaking change or deprecation and creating a new field seem like the best options.

Let me know what you prefer and I can adjust the PR.

rbruckheimer:
> 

Let's go with your proposal and we can change accordingly.

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 * Updated definition of the `Instrument` context type to include optional market identifiers ([#819](https://github.com/finos/FDC3/pull/819))
 * Corrected API functions and object types to always use `string` instead of `String` ([#924](https://github.com/finos/FDC3/pull/924))
+* Updated definition of the `otherCOnfig` element of the `Chart` context type from an Object to an array of Contexts as this allows the `type` of each additional item of config to be examined before it is used ([#985](https://github.com/finos/FDC3/pull/985))
```
kriswest, line 17:
```suggestion
* Updated definition of the `otherConfig` element of the `Chart` context type from an Object to an array of Contexts as this allows the `type` of each additional item of config to be examined before it is used ([#985](https://github.com/finos/FDC3/pull/985))
```

mattjamieson (approved):

kriswest:
@mistryvinay this should be ready for your review and merge into your consolidated update branch for Context and Intents

kriswest:
@mistryvinay actually looks like I've got small git snafu after changing the base, will fix.

kriswest:
fixed and good to go @mistryvinay.
Needs @greyseer256's blessing too.

mistryvinay (approved):
LGTM

kriswest:
@greyseer256 please check this one when you have a second, so we can merge into the consolidated Context & Intents branch for 2.1

greyseer256 (approved):

PR #987, opened 2023-05-12
866 app add listener timeouts

### LINKED ISSUE #866: Are you ready yet? Adding defined (or at least bounded) timeouts to the FDC3 Desktop Agent API
## Enhancement Request
During the implementation of the FDC3 conformance testing framework it became clear there is some variation between desktop agents on the implementation of `fdc3.open` & `fdc3.raiseIntent`/`fdc3.raiseIntentForContext` and the errors that they return. Specifically, the `OpenError.AppTimeout` and `ResolveError.IntentDeliveryFailed` errors, which (as documented in FDC3 2.0) are returned if an app fails to add the necessary context listener or intent listener (via fdc3.addContextListener and fdc3.addIntentListener) within _a timeout_. 

As there is no way in FDC3 for an app to indicate that it is 'ready' other than to add the relevant listeners, the listeners are a key point of synchronization and waiting on them with a timeout is the only way for a Desktop Agent to detect an issue and to return an error. The ability to do so is important as many FDC3 interactions will involve up to 3 firms/development team (AppA, Desktop Agent, AppB) and the error indicates where any issue (currently being debugged) is likely to lie. 

The errors returned by all Desktop Agent functions are very poorly documented in FDC3 1.2 (individual errors are not described or documented) and some tests based on them have been removed from the corresponding version of the conformance testing framework. FDC3 2.0 is more specific in the documentation of the errors - although the reference documentation for the relevant functions could and should be more specific as well to aid both app and platform developers using them as reference. 

However, there is a further issue for conformance testing in that the specific timeouts used are not defined or bounded by the standard. Hence, the conformance tests (currently) adopt a relatively large value (75 seconds) - but also include the caveat that this can be extended if a Desktop Agent implementation uses a longer timeout. 

We propose that: 
- a required maximum timeout is specified, while DA implementation MAY use a shorter timeout if they wish.
- a short description of which errors are returned when is added to the function reference for any Desktop Agent or Channel function that can return them OR a similar reference created in the [Errors page](https://fdc3.finos.org/docs/api/ref/Errors) and referred to.

### Use Case:
Apps in financial services can sometimes take a while to bootstrap themselves (leading to the relatively long timeouts used in some DA implementations). However, there are often ways that this can be handled (such as adding the listeners early and retaining/queuing up messages received to be processed when the app is ready. Were FDC3 to define suitable timeouts (or at least bounds on them) then: 

- developers would be encouraged to use such handling
- behaviour from desktop agents would be more consistent
- conformance tests would not require long (and extensible) timeouts and would run quicker.

kriswest:
resolves #866

Adds details of timeouts after launch within which applications SHOULD (30 seconds) and MUST (90 seconds) add their context or intent listeners via the API, and details of the same timeouts being allowed for by Desktop Agents (MUST 30 seconds, SHOULD 90 seconds).


nkolba:
@kriswest .   Don't agree with the MUST here.  Especially since the language seems to assume there is a single flow in which apps add context and intent listeners (i.e. immediately on startup).

docs/intents/spec.md
nkolba, line 1: What is the reason to restrict use of addIntentListener to with 90 seconds of launch?  This seems overly proscriptive.

src/api/DesktopAgent.ts
nkolba, line 1: 90 second requirement seems overly proscriptive and assuming a specific pattern and flow for apps that may not be the case.  What is the reason for requiring this?

kriswest: The flow is prescribed by `fdc3.open` and `fdc3.raiseIntent` and the passing of context to an app on launch and these requirements specifically linked to that case. 

As per the issue, the lack of any advice or bound on the timeouts means both app developers and desktop agent developers are guessing at values. That in turn makes conformance testing harder as we don't know how long you've got to wait before deciding the DA has failed to notice the failure to add the listener and to subsequently return the relevant error.

The point was to add a range and a recommendation to help guide implementors, nothing more. 

I'm told no one that has conformance tested has used a timeout longer than 70 seconds, so I went higher at 90. I further posted to the FDC3 maintainers to ask if anyone has seen higher or wanted to feedback on it.

docs/intents/spec.md
kriswest: It isn't restricted to that, except where you want to receive intent/context on startup.

I would very much like to continue dynamically adding intent listeners dynamically in the FDC3 workbench, which certainly later than 90 seconds.

src/api/DesktopAgent.ts
nkolba: @kriswest  guidance and recommendations are good.  But what is gained from the MUST?  The conformance tests are a closed loop, so the only reason for a timeout of intent resolution in the conformance tests is if something else is broken.    What's the use case for this being a MUST and are there other use cases that we might end up blocking?

robmoffat: Unless we know what the maximum time is, the conformance test can't finish.  

In the case `2.0-RaiseIntentFailTargetedAppResolve3` where the desktop agent is waiting for a newly-opened app to respond to an intent (and it doesn't) then the desktop agent should return `ResolveError.IntentDeliveryFailed`.

However, if the test calls it (and fails) at 90 seconds, but the desktop agent has a timeout of 95 seconds, then the test will fail erroneously - the DA is performing according to specification but is doing it beyond the wait time of the test.

kriswest: @nkolba as discussed in the maintainers channel the reason for there being a timeout in the conformance test is that Desktop Agents are already required to return a particular error from `ResolveError` or `OpenError` on failure to deliver an intent/context, hence, at some point it has to decide that the DA is not going to return that error and is just hanging.  Without a known timeout or at least some bound on the timeout there's no way to know whether that is the case if it doesn't return. The same applies to an application awaiting an `open` or `raiseIntent` call - without there being a bound on the timeout's used, they might need to implement their own.

An alternative is to just make it a SHOULD. However, that's vague and hand-wavy - an app developer or adopter can't rely on it. Whereas, a stronger requirement should encourage developers to find alternative solutions in cases where their app start-up is longer than the proposed bound.

An alternative way to soften the requirement would be to put the bound on the default timeout and add a MAY allowing the DA to be configured to use a longer timeout for particular apps. I believe you were not keen on adding such to the appD record in standard place, but they could do so within a hostManifest easily enough. Perhaps that is the best solution as it both encourages better behavior by default, but also allows for cases where more than the bounded timeout is needed.

nkolba: @robmoffat @kriswest   Couldn't the conformance tests just set their own timeout?  Is there any downside?  The tests are a known quantity and there is no reason for them to get anywhere even close to 90 seconds unless the test was broken.

kriswest: They do set their own timeout, but what to set it to is unknown. The point of those tests is to confirm the behaviour of the desktop agent (response to `raiseIntent` or `open`) when the app doesn't behave correctly/is broken. 

This isn't just about making it easier to test standard behaviour, rather it's to make what the standard behaviour is (for both apps and Desktop Agents) clearer.

kriswest:
The lack of a defined timeout was independently raised again by a team from a Desktop Agent vendor reviewing the conformance testing framework. In that case, they are counting the timeout from when the intent is raised, rather than from when the app is spawned, hence any time a resolver UI or similar is open is included. 

Counting from when the intent is raised has the advantage that the app raising the intent can know how long it might be open. However, it makes the timeout available to the launched app variable as time is likely to have been taken up with a user interacting with a resolver UI. 

Personally, I think it more useful to advise the app being launched as to how quickly they should be adding their listener - particularly as the app initiating the action (`raiseIntent`/`open`) will receive notice if a timeout occurs via the defined errors, whereas there is no feedback to the opened app if it times out - it can still add its listeners with no error, but the intent/context may just be delivered. However, a subsequent raiseIntent or broadcast could still reach it...

kriswest:
Updated the draft to clarify that the timeouts only apply to intents and contexts that should be receivable on launch via `fdc3.open` and the `raiseIntent` functions. Also added a MAY clause allowing the timeouts to be extended for individual apps.

rikoe:
Please see my comment on the related issue.

I think it is a mistake to push demands about timeout behaviour from the conformance layer into the standard. I think the expected timeouts can live in and be dictated by the conformance tests - if you don't fit into their expected timeouts, you can't get certified.

But because we are dealing with real users and UIs, the edges are too hard in terms of imposing conditions inside the standard itself for this, and that is why it was never done.

kriswest:
@rikoe 

> Please see my comment on the related issue.
I've replied over there: https://github.com/finos/FDC3/issues/866#issuecomment-1561072827

> I think it is a mistake to push demands about timeout behaviour from the conformance layer into the standard. 

Thats not what is intended, rather it was conversations between vendors during meetings looking at the tests that flagged a vague part of the standard that might be more helpful (to multiple stakeholders) if it was more prescriptive.

> But because we are dealing with real users and UIs, the edges are too hard in terms of imposing conditions inside the standard itself for this, and that is why it was never done.

When a Standard ducks out of addressing hard issues, its just pushing those issues down to its users/implementors in my opinion.  Where we're dealing with multiple stakeholders (a couple of app vendors, a user, a DA vendor) that can get quite complex when they vary (e.g. swapping out different apps, different DAs), which is why standardizing behaviour is useful.

This is a relatively small issue, and clearly the upper bound is more controversial (despite the allowance for config to vary/extend it that was added?). A you less uncomfortable with shorter value for at least a minimum bound? E.g. DAs MUST alow at least 10 seconds? Or are you against any form of clarification on how quickly apps should add listeners for use cases where they are handed context/intents on startup?

kriswest:
In the interest of agreeing some form of improvement, I've adjusted the PR to only include: 
- a minimum 15 sec bound on the timeout, 
- no upper bound on the timeout
- a recommendation that apps add their listeners as quickly as possible
- clearer language around the dual use of addCOntextListener for User channel context AND context passed by `open`.

Hopefully this version is acceptable @rikoe and @nkolba - if you'd like further change please suggest wording if at all possible.

nkolba:
@kriswest these changes look better to me, now that the MUST is only on the minimum for a timeout.  Would still prefer no MUST here (an implementor could feasibly have no timeouts and just allow the end user to cancel the operation if the launching app never responded), but this seems OK.

kriswest:
> Would still prefer no MUST here (an implementor could feasibly have no timeouts and just allow the end user to cancel the operation if the launching app never responded), but this seems OK.

This was discussed at the maintainers meeting: The existence of a minimum timeout doesn't preclude having a (near) infinite timeout so it continues until the user somehow cancels the launch operation.  The timeout applies after resolution on app launch, so this is different from the user closing a resolver UI (which should result in `ResolveError.UserCancelled`).

docs/api/spec.md
```diff
@@ -121,6 +121,7 @@ An FDC3 Standard compliant Desktop Agent implementation **MUST**:
...
 - Only require app directories that they connect to to have implemented only the minimum requirements specified in the [App Directory API Part](../app-directory/spec) of this Standard.
 - Provide details of whether they implement optional features of the Desktop Agent API in the `optionalFeatures` property of the [`ImplementationMetadata`](ref/Metadata#implementationmetadata) object returned by the [`fdc3.getInfo()`](ref/DesktopAgent#getinfo) function.
+- Allow, by default, at least a 15 second timeout for an application, launched via [`fdc3.open`](../api/ref/DesktopAgent#open), [`fdc3.raiseIntent`](../api/ref/DesktopAgent#raiseintent) or [`fdc3.raiseIntentForContext`](../api/ref/DesktopAgent#raiseintentforcontext) to add any  context listener (via [`fdc3.addContextListener`](../api/ref/DesktopAgent#addcontextlistener)) or intent listener (via [`fdc3.addIntentListener`](../api/ref/DesktopAgent#addintentlistener)) necessary to deliver context or intent and context to it on launch. This timeout only applies to listeners needed to receive context on launch; further intent and context listeners not required on launch MAY be added later.
```
greyseer256, line 124:
```suggestion
- Allow, by default, at least a 15 second timeout for an application, launched via [`fdc3.open`](../api/ref/DesktopAgent#open), [`fdc3.raiseIntent`](../api/ref/DesktopAgent#raiseintent) or [`fdc3.raiseIntentForContext`](../api/ref/DesktopAgent#raiseintentforcontext) to add any context listener (via [`fdc3.addContextListener`](../api/ref/DesktopAgent#addcontextlistener)) or intent listener (via [`fdc3.addIntentListener`](../api/ref/DesktopAgent#addintentlistener)) necessary to deliver context or intent and context to it on launch. This timeout only applies to listeners needed to receive context on launch; further intent and context listeners not required on launch MAY be added later.
```

docs/api/spec.md
```diff
@@ -352,6 +353,8 @@ Applications need to let the system know the intents they can support.  Typicall
...
 
+Intent handlers SHOULD be registered via [`fdc3.addIntentListener`](ref/DesktopAgent#addintentlistener) within 15 seconds of the application launch (the minimum timeout Desktop Agents are required to provide) in order to be widely compatible with Desktop Agent implementations. Individual Desktop Agent Implementations MAY support longer timeouts or configuration to control or extend timeouts.
```
greyseer256, line 356:
```suggestion
Intent handlers SHOULD be registered via [`fdc3.addIntentListener`](ref/DesktopAgent#addintentlistener) within 15 seconds of the application launch (the minimum timeout Desktop Agents are required to provide) in order to be widely compatible with Desktop Agent implementations. Individual Desktop Agent implementations MAY support longer timeouts or configuration to control or extend timeouts.
```

src/api/DesktopAgent.ts
```diff
@@ -379,7 +383,7 @@ export interface DesktopAgent {
    *
    * If an app is joined to a channel, all `fdc3.broadcast` calls will go to the channel, and all listeners assigned via `fdc3.addContextListener` will listen on the channel.
    *
-   * If the channel already contains context that would be passed to context listeners assed via `fdc3.addContextListener` then those listeners will be called immediately with that context.
+   * If the channel already contains context that would be passed to context listeners added via `fdc3.addContextListener` then those listeners will be called immediately with that context.
```
greyseer256, line 386: good catch

greyseer256 (approved):

mattjamieson (approved):
As discussed and agreed in the maintainers' meeting so LGTM

PR #997, opened 2023-05-22
Don't use deprecated functions in 2.0 version of workbench

kriswest:
@bingenito spotted that the FDC3-workbench was still using deprecated functions (`getSystemChannels` and `joinChannel`) in its 2.0 version. This works if they are implemented, but they don't have to be. Hence, use the modern versions of the functions in 2.0 and only use the deprecated versions in the 1.2 version.


bingenito (approved):
lgtm

PR #1004, opened 2023-05-24
Clarify intended behavior on void intent results

### LINKED ISSUE #1003: Clarify expected behaviour of IntentResolution.getResult() when IntentHandlers resolve to void
### Area of Issue
[x] API

### Issue Description: 
The documentation of the IntentHandler type introduced in 2.0 and related behaviour from the `IntentResolution.getResult()` API call, when a `void` result is returned, is open to interpretation due to a confusing description. Clarify the description to ensure its easy to understand the expected behavior.

### Additional Context:
One of the use-cases for `getResult()` discussed when it was introduced was knowing when the intent handler has completed. This should be possible even when the intent handler is returning void and shouldn't appear to be an error. However, the description of IntentHandler and the `ResultError.NoResultReturned` don't quite match up (You might resolve `getResult()` to `void` OR reject with `ResultError.NoResultReturned`. It was not intended that the promise rejects with an error on expected behavior.

As this was not clear in 2.0 it was suggested that the 2.0 conformance tests allow for either output, while 2.1 fixes the description and standardizes on resolving to `void`.

kriswest:
resolves #1003

Clarifies description of IntentResolution.getResult() when the intent handler returned void (which is not an error)

### DIFF
--- a/src/api/DesktopAgent.ts
+++ b/src/api/DesktopAgent.ts
@@ -291,7 +291,7 @@ export interface DesktopAgent {
   raiseIntentForContext(context: Context, app?: AppIdentifier): Promise<IntentResolution>;
 
   /**
-   * Adds a listener for incoming intents from the Desktop Agent. The handler function may return void or a promise that resolves to a `IntentResult`, which is either a `Context` object, representing any data that should be returned to the app that raised the intent, a `Channel` Object, a `PrivateChannel` over which data responses will be sent, or `void`. The `IntentResult` will be returned to the app that raised the intent via the `IntentResolution` and retrieved from it using the `getResult()` function.
+   * Adds a listener for incoming intents from the Desktop Agent. The handler function may return void or a promise that resolves to an `IntentResult`, which is either a `Context` object, representing any data that should be returned to the app that raised the intent, a `Channel` Object, a `PrivateChannel` over which data responses will be sent, or `void`. The `IntentResult` will be returned to the app that raised the intent via the `IntentResolution` and retrieved from it using the `getResult()` function.
    *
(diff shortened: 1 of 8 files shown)


src/api/DesktopAgent.ts
```diff
@@ -291,11 +291,11 @@ export interface DesktopAgent {
   raiseIntentForContext(context: Context, app?: AppIdentifier): Promise<IntentResolution>;
 
   /**
-   * Adds a listener for incoming Intents from the Agent. The handler function may return void or a promise that should resolve to an `IntentResult`, which is either a `Context` object, representing any data that should be returned, or a `Channel` over which data responses will be sent. The IntentResult will be returned to app that raised the intent via the `IntentResolution` and retrieved from it using the `getResult()` function.
+   * Adds a listener for incoming intents from the Desktop Agent. The handler function may return void or a promise that resolves to a `IntentResult`, which is either a `Context` object, representing any data that should be returned to the app that raised the intent, a `Channel` Object, a `PrivateChannel` over which data responses will be sent, or `void`. The `IntentResult` will be returned to the app that raised the intent via the `IntentResolution` and retrieved from it using the `getResult()` function.
```
greyseer256, line 294:
```suggestion
   * Adds a listener for incoming intents from the Desktop Agent. The handler function may return void or a promise that resolves to an `IntentResult`, which is either a `Context` object, representing any data that should be returned to the app that raised the intent, a `Channel` Object, a `PrivateChannel` over which data responses will be sent, or `void`. The `IntentResult` will be returned to the app that raised the intent via the `IntentResolution` and retrieved from it using the `getResult()` function.
```

greyseer256 (approved):

rikoe (approved):

Issue #989, opened 2023-05-16
Typo in docs/context /spec.md


mukeshmohan:
## Minor Issue

Area of Issue
Typo Error in the content in Spec.md file

Issue Description: 
Under identifier section, Found a Typo in the 2nd paragraph where "identifier" is mentioned as "identifer".

Additional Context:
Created this issue for one of the workshop conducted in our organization for the first time contributors

mukeshmohan:
I am working on this issue.

Issue #832, opened 2022-10-11
Unit Test Policy


robmoffat:
## Minor Issue
We now have the OpenSSF badge on the README.   It's at 99%.  To get to 100%, we need a test policy for the FDC3 unit tests.  It could be something written down like:

> Any change to the FDC3 specification typescript codebase must be supported by accompanying unit tests

or a rule in our build system that says the build fails unless the coverage increases / stays same from the previous build.

kriswest:
@robmoffat uses https://about.codecov.io/ for spring bot's code coverage (free for opensource)

ashok-hariharan:
Wish if I could take this as my first contribution to opensource

robmoffat:
that would be awesome!  I look forward to your PR, @ashok-hariharan!

PR #1012, opened 2023-06-19
Typo Fixes

ashok-hariharan:
Fixing two issues as part of my first contribution to Open Source 
1. Added unit test policy #832  (will raise a separate PR for this issue as it needs more details)
2. Typo in multiple files #989 

Please check this @robmoffat

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -81,7 +81,11 @@ The next section contains more information on the workflow followed for Pull Req
 
 - The Maintainers will apply the proper labels for the issue. This may include an indication of the subject area or type (e.g. `deprecation`).
 
-**4.3.	Reviewing/Discussion.**
+**4.3.	Unit Test Policy**
+
+- Any change to the FDC3 specification typescript codebase must be supoorted by accompanying unit tests.
+
+**4.4.	Reviewing/Discussion.**
 
 - All PRs will be reviewed by at least one of the FDC3 Maintainers and any appointed Editors. PRs are also open to review by FDC3 Participants.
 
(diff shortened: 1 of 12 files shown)


CONTRIBUTING.md
```diff
@@ -81,7 +81,11 @@ The next section contains more information on the workflow followed for Pull Req
 
 - The Maintainers will apply the proper labels for the issue. This may include an indication of the subject area or type (e.g. `deprecation`).
 
-**4.3.	Reviewing/Discussion.**
+**4.3.	Unit Test Policy**
+
+- Any change to the FDC3 specification typescript codebase must be supoorted by accompanying unit tests.
```
kriswest, line 86: Couple of types to fix:

```suggestion
- Any change to the FDC3 specification's typescript codebase must be supported by accompanying unit tests.
```

kriswest, line 86: Some more detail would also be useful here - FDC3 is made up of 4 specifications, one of which relates to the Desktop Agent API. Most of the codebase is just typing to define that API standard. However, there is actual code in method.ts, which is currently all that we apply unit tests to. There is also code in ContextTypes.ts, which is auto-generated from context schemas. This does not currently include tests - and we might not want to be forced to add them when new Context type schemas are added and start autogeneratingnew functions.

There are other apps in the /toolbox directory. These don't currently have unit tests.
We might add further apps there, e.g. a basic app directory implementation - the existing out-of-date implementation was recently removed.

kriswest (changes requested):
Type fixes all look good.

We need a bit more thought on/detail in the unit test policy before merging that. Hence, you might consider splitting this into two PRs so that they can be handled separately.

CONTRIBUTING.md
ashok-hariharan: Removed unit test policy changes. Will keep it as a separate PR  by adding some more additional details to it

kriswest (approved):
LGTM! Will update title to reflect removal of Unit test policy

greyseer256 (approved):

ashok-hariharan:
I don't have write access, requesting @greyseer256 or @kriswest to merge it

PR #1018, opened 2023-06-27
rebranding glue and Finsemble as interop.io (first pass)

kriswest:
First pass at tweaking users/community/conformance data for the interop.io merger.

users.json changes switch the two company logos to a single copy of the new interop.io logo:
image

At present there remain two separate products, published by interop.io. Hence, it seems to make sense to brand both with the interop.io logo and blurb, but keep their other data separate for now. Later we'll remove both and replace with the merged io.Connect product when released and conformance tested.

image

image


kriswest:
@robmoffat, the preview was approved this end.

robmoffat:
I've added the 2.0 conformance stuff - can you check on the front and community pages?

robmoffat (approved):

kriswest:
Conformance data looks good to me - although I'd recommend reversing the sort order on the FDC3 versions so 2.0 is first (generally more relevant and eventually older versions can drop off the bottom). That looks easy on the homepage (where I think it matters more), as you just transpose [these elements](https://github.com/finos/FDC3/blob/3ee1b4b6977cd3dbd7d1b1f639f7f23ea32edab2/website/data/front/conformance.mdx#L7-L13). On the community page it looks like order is down to the order they appear in the array - but there the layout is much more compact and the order matters less.

robmoffat:
changed the order - merge if you're happy

PR #1020, opened 2023-07-05
Copying markdown docs for contexts into schema files

### LINKED ISSUE #1032: Context Part does not explain use of schemas, schemas are missing key documentation
The Context Part of the FDC3 Standard makes use of JSON Schema definitions as the 'source of truth' for context definitions (supported by docs and examples on the reference pages). However, this is not explained anywhere in the Standard's documentation. 

FDC3's use of schemas should be improved by:

- Adding detail to the Context overview on: 
    - How JSON schemas are used
    - What to include in a schema
- Copying documentation from markdown pages into JSON Schemas

Making these changes would enable future changes to generate the docs pages from the schema files (reducing maintenance / making future contributions easier)

### Area of Issue
[x] Context Data

kriswest:
Pr that migrates context docs from markdown files into a schema. The markdown files still exists for the time being - we hope to try an generate markdown pages or directly render documentation pages from these schemas so that we can go full circle (and then delete the hand-maintained markdown docs in favour of generated).

partially resolves #1032


hampshanubs:
@kriswest Great idea.

kriswest:
@robmoffat @mistryvinay @hughtroeger I think I'm done - all context types now have docs for all fields I believe. I've regenerated the TypeScript types, which look better now that all fields have titles - although there are still a few Purples and Fluffys - these occur because quicktype seems to treat top-level types and references differently, generating a duplicate where a context is referenced elsewhere. There may be a workaround possible by referencing all contexts in some other schema just used to do generation (meaning all are references, none are top-level). Will explore when I have time - until then, or until quicktype improves, we'll have to live with them.

These changes will need a review and merge. They should not be changing the standard, just improving the docs.

@robmoffat should be all the feed material needed to look at rendering docs from context schemas.

schemas/context/position.schema.json
```diff
@@ -3,12 +3,42 @@
...
+          "const": "fdc3.position"
+        },
+        "instrument": {
+          "$ref": "instrument.schema.json#",
+          "title": "The financial instrument that this position relates to",
+          "description": ""
+        },
+        "holding": {
+          "type": "number",
+          "title": "Thesize of the holding represented by this position",
```
hughtroeger, line 21:
```suggestion
          "title": "The size of the holding represented by this position",
```

kriswest:
Thanks @hughtroeger, fixes applied for your review.

@mistryvinay  and @finos/fdc3-maintainers I need a review from you to merge this. Most obvious change will be that there are now docs for generated typescript files that will show up in VSCode and other editors (yay) - all are drawn from the Markdown docs that we hope to replace (with rendering from schema files) in future.

kriswest:
@hughtroeger please also be aware of/review the following commit I just added to this PR which finally explains our use of JSON schema to define Context types and bridging messages:
https://github.com/finos/FDC3/pull/1020/commits/1d290271b54b680690f79f8ab3171033951b9a94

docs/context/ref/Action.md
```diff
@@ -6,12 +6,11 @@ hide_title: true
 ---
 # `Action`
```
robmoffat, line 7: this seems to be missing front matter to allow it to appear in the docusaurus nav.

robmoffat (changes requested):

docs/context/ref/Action.md
robmoffat: schema linking otherwise looks good though.

robmoffat: I just looked at the preview

kriswest: That'll be a missing sidebar entry then, will add to PR - thanks for flagging it

kriswest: Should be fixed - spotted a couple of other minor errors in the sidebar and fixed those too.

hughtroeger (approved):

mistryvinay (approved):
LGTM - thanks for putting this together Kris. 🚀

PR #1025, opened 2023-07-18
Pin github actions in package.yml

kriswest:
Pins the github actions to a particular SHA1 to stop semgrep failing


robmoffat (approved):

robmoffat:
@maoo take notice of this - we're probably going to see this a lot

PR #1029, opened 2023-07-19
Apache headers 2

robmoffat:
Added missing Apache 2 License headers to .ts source files
Part of move towards CSL license


src/context/ContextTypes.ts
kriswest, line 1: This file is generated, the header will get wiped out whenever regenerated unless some post-processing is installed.

kriswest: Its generated by this script: https://github.com/finos/FDC3/blob/4efe985d18968c73a2b989b6427733fc000f18de/package.json#L26

Beware changes to this in the bridging PR however.

robmoffat: should that be in the .gitignore then?

kriswest: possibly... its currently checked into the repo and comments on it turn up in reviews, which that would prevent. If ignored it will only exist when the NPM module is built and released.

I suppose the alternative is to a script (run in node) that reads the file, pastes in the notice and writes it back again...

robmoffat: > possibly... its currently checked into the repo and comments on it turn up in reviews, which that would prevent. If ignored it will only exist when the NPM module is built and released.

yeah, that makes sense to me - I generally don't think you should check in files that are generated, otherwise that breaks the "single source of truth" concept.

Happy to add a commit to do that if you want to

maoo (approved):
LGTM!

src/context/ContextTypes.ts
kriswest: What I meant was that maintainer reviews have often caught problems by looking at the generated code - in fact I need to go rectify one involving generated names for 2.1.

However, I agree that it can be confusing as to what the source of truth is + I've learnt a lot more recently about how quicktypes code generation works...

On this one I'd ask the other maintainers (and will drop a quick slack message)

kriswest: https://finos-lf.slack.com/archives/C0296SHSY1J/p1689783212141519

robmoffat: also FWIW we can't really copyright generated code

kriswest: Can we not? Almost any npm module has JS source files generated from TS sources. I'd assume the same principles would apply to JSON-schema -> TS. Hence, I'd assume we can both claim it and license it (and effectively are as the license is applied to the NPM module that it ends up in).

kriswest (changes requested):
Missed adding license to all files in or under:
- [/toolbox/fdc3-workbench/src/components](https://github.com/finos/FDC3/tree/master/toolbox/fdc3-workbench/src/components)
- [/toolbox/fdc3-workbench/src/App.css](https://github.com/finos/FDC3/blob/master/toolbox/fdc3-workbench/src/App.css)
- [/toolbox/fdc3-workbench/src/App.tsx](https://github.com/finos/FDC3/blob/master/toolbox/fdc3-workbench/src/App.tsx) (TSX files but should get same treatment
- [/toolbox/fdc3-workbench/src/index.tsx](https://github.com/finos/FDC3/blob/master/toolbox/fdc3-workbench/src/index.tsx)
- [toolbox/fdc3-explained](https://github.com/finos/FDC3/tree/master/toolbox/fdc3-explained) (JS files but should get same treatment)
- [s2tQuicktypeUtil.js](https://github.com/finos/FDC3/blob/master/s2tQuicktypeUtil.js)
- [schemas/api/t2sQuicktypeUtil.js](https://github.com/finos/FDC3/blob/master/schemas/api/t2sQuicktypeUtil.js)

Otherwise looks good

kriswest (approved):
LGTM

PR #1034, opened 2023-07-24
Correct issue in appd details field's schema, error in example and improve test

### LINKED ISSUE #1033: App Directory schema contains errors
### Area of Issue
[x] App Directory

### Issue Description: 
The app directory schema is currently failing its test. The example schema at https://github.com/InteropIO/FDC3/blob/544-Desktop-Agent_bridging-Proposal-mk2/src/app-directory/specification/examples/application/fdc3-workbench.json contains error (type should be "web" rather than "browser". However, once corrected the test still won't pass as the `details` element can't be validated (due to the use of `oneOf`, where the subschemas overap - this should be `anyOf`), amongst other errors.

Errors are only returned one at a time due to the encoding in YAML, switching to JSON schema would allow all errors to output at once. 

Correct the errors in the App Directory schema and consider switching the encoding to JSON Schema.

kriswest:
resolves #1033 

- the details field of the appD schema is using `oneOf` to combine subschemas
  - However, this requires them to be distinct and non-overlapping, they are not (both web and online native have a `url` property). 
  - `anyOf` should be used in this situation instead as `oneOf` returns validation errors when the element conforms to multiple subschemas.
- commas were removed from email fields in the fdc3-workbench example as these fail validation (just a syntax error in the YAML file)
- The test was improved by adding multiple examples and providing better console output from it.

### DIFF
--- a/src/app-directory/specification/appd.yaml
+++ b/src/app-directory/specification/appd.yaml
@@ -542,7 +542,7 @@ components:
         The type specific launch details of the application. These details are intended to be 
         vendor-agnostic and MAY be duplicated or overridden by details provided in the hostManifests 
         object for a specific host.
-      oneOf:
+      anyOf:
         - $ref: '#/components/schemas/WebAppDetails'
         - $ref: '#/components/schemas/NativeAppDetails'
         - $ref: '#/components/schemas/CitrixAppDetails'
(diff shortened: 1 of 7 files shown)


mattjamieson (approved):
LGTM!

WatsonCIQ:
Fixed my issue! 👍

hughtroeger (approved):

PR #1035, opened 2023-07-24
1033 convert appd schema to json

kriswest:
resolves #1033 
builds on PR #1034 

Converts the AppD schema file from YAML to JSON Schema (where the tooling is a little better and ensuring that we use one schema format across FDC3, where there are already other examples of JSON Schema use for contexts, bridging and bridging asyncAPI).

Updates were applied to to package scripts and the website to use the updated encoding.

PR #1037, opened 2023-07-24
1036 correct appd schema for images and screenshots

### LINKED ISSUE #1036: AppD schema does a poor job of validating icons and screenshots
### Area of Issue
[x] App Directory

### Issue Description: 
Icons in app appD records should have a format like:

```JSON
...,
"icons": [
    "src": "<some URL>,
    "size": "25x45",
    "type": "image/png"
],
...
```
(the last two fields are optional, the `src` field should be required)

However, the schema is poorly written in that it is missing the required property declaration and does not preclude `additionalValues`. Hence, an entry such as:

```JSON
...,
"icons": [
    "url": "<some URL>,
],
...
```

will validate as `src` was not correctly required in the schema and additional properties are allowed.

Refine the schema so that it can better validate appD records.

kriswest:
resolves #1036 
Builds on #1035 which should be merged first to convert appD schema to JSON Schema format.

Requires the src property for each (as originally intended), prevents additional values in these sub-schemas (meaning its more likely to catch inadvertent errors like using a `url` field in place of `src`, or a misspelling of another field).


WatsonCIQ:
The fix worked: Caught the missing error for icons.src for me. 👍

hughtroeger (approved):

mattjamieson (approved):

PR #1042, opened 2023-07-27
Adjusting package.json to roll docusaurus version numbers

kriswest:
The package-lock.json file is gitignored so dependencies are being re-resolved on build. This is causing issues as, for some reason, the them-mermaid resolved to a later version that docusaurus, causing the website to crash on pages with mermaid diagrams.

From the last production deployment:
image

which causes (I believe):
image


mistryvinay (approved):
LGTM

mattjamieson (approved):

PR #1060, opened 2023-09-06
Correct URLs in appD overview docs

kriswest:
(missing /apps element so example URLs are wrong in all past and current versions of the overview - identified by @Yannick-Malins )

The overview and spec don't agree on the URLs - the spec is the source of truth so updating the overview examples to match it.


Yannick-Malins (approved):

mistryvinay (approved):
LGTM

hughtroeger (approved):

PR #1067, opened 2023-09-20
Correcting an error in privateChannel bridging schemas vs. documentation

kriswest:
We spotted a couple of copy/paste errors in the private channel bridging schemas that did not agree with the documentation, and one docs typo which is fixed in this PR.


tpina (approved):

tpina (approved):
Nice!

hughtroeger (approved):

mattjamieson (approved):
LGTM!

PR #1074, opened 2023-09-25
minor docs clarification on bridging channel state sync

kriswest:
The description of how/why channel state is synchronized in the bridging specification could be clearer - by better defining how Channel state works. THis PR adds that minor clarification without changing anything in the specification.


mattjamieson (approved):
LGTM!

hughtroeger (approved):

PR #1080, opened 2023-09-28
Update community.json

JonnyAdaptableTools:
Changing just the AdapTable entry!!!


kriswest:
Looks like you did it while I was reviewing, so all good!

JonnyAdaptableTools:
> Looks like you did it while I was reviewing, so all good!

We like to stay one step ahead :)

PR #1082, opened 2023-09-29
Upgrade quicktype and correct a schema error

kriswest:
While trying to generate Zod schemas for bridging messages I upgraded quicktype (which now does a better job of controlling the const type fields in COntext and bridging message types via single value enums, and in a future version, via const declarations see https://github.com/glideapps/quicktype/pull/2314).

However, I then ran into an issue with one of the bridging schemas, whose message type was missing from the enumeration of all message types. This PR: 
- applies the quicktype update, 
- corrects the missing message type
- regenerates the context and bridging types exposed in the NPM module
- updates the schemas files in 2.1 and next with the correction.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,17 +1,17 @@
 {
   "name": "@finos/fdc3",
-  "version": "2.0.3",
+  "version": "2.1.0-beta.2",
   "lockfileVersion": 3,
   "requires": true,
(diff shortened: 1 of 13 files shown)


mattjamieson (approved):
LGTM!

hughtroeger (approved):

Issue #1061, opened 2023-09-12
Replace tsdx use in build of the FDC3 NPM module


kriswest:
### Area of Issue
[x] API

~Upgrading tsdx to 0.13.3~ Replacing tsdx (used to bundle the FDC3 NPM module) would resolve 17 moderate vulnerabilities in the FDC3 repo and should ease future maintenance as tsdx is now unmaintained ([ref](https://github.com/jaredpalmer/tsdx/issues/1065)).

### npm audit report
```
jsdom  <=16.5.3
Severity: moderate
Insufficient Granularity of Access Control in JSDom - https://github.com/advisories/GHSA-f4c9-cqv8-9v98
Depends on vulnerable versions of request
Depends on vulnerable versions of request-promise-native
Depends on vulnerable versions of tough-cookie
fix available via `npm audit fix --force`
Will install tsdx@0.13.3, which is a breaking change
node_modules/jsdom
  jest-environment-jsdom  10.0.2 - 25.5.0
  Depends on vulnerable versions of jsdom
  node_modules/jest-environment-jsdom
    jest-config  12.1.1-alpha.2935e14d - 25.5.4
    Depends on vulnerable versions of @jest/test-sequencer
    Depends on vulnerable versions of jest-environment-jsdom
    Depends on vulnerable versions of jest-jasmine2
    node_modules/jest-config
      jest-cli  12.1.1-alpha.2935e14d || 12.1.2-alpha.6230044c - 25.5.4
      Depends on vulnerable versions of @jest/core
      Depends on vulnerable versions of jest-config
      node_modules/jest-cli
        jest  12.1.2-alpha.6230044c - 25.5.4
        Depends on vulnerable versions of @jest/core
        Depends on vulnerable versions of jest-cli
        node_modules/jest
          tsdx  >=0.14.0
          Depends on vulnerable versions of jest
          node_modules/tsdx
      jest-runner  21.0.0-alpha.1 - 25.5.4
      Depends on vulnerable versions of jest-config
      Depends on vulnerable versions of jest-jasmine2
      Depends on vulnerable versions of jest-runtime
      node_modules/jest-runner
        @jest/test-sequencer  <=25.5.4
        Depends on vulnerable versions of jest-runner
        Depends on vulnerable versions of jest-runtime
        node_modules/@jest/test-sequencer
      jest-runtime  12.1.1-alpha.2935e14d - 25.5.4
      Depends on vulnerable versions of jest-config
      node_modules/jest-runtime
[...]

PR #1090, opened 2023-10-17
Fixing a build error introduced by quicktype enums

kriswest:
THe latest version of quicktype has started generating single value enums for constant values, which are causing issues in the context type tests (Context types serialize but cause type errors on deserialization). However, its also capable of generating const values - although this option hasn't been exposed correctly through the CLI  (raising a PR to fix). Generating true const value fixes the issue.

I hadn't noticed this test failing when I merged:
- #1082 
This PR is needed to get the build passing again - although we also need a quicktype release to make it repeatable

Also resolving a critical vulnerability (babel) via `npm audit fix`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@finos/fdc3",
-  "version": "2.1.0-beta.2",
+  "version": "2.1.0-beta.3",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@finos/fdc3",
-      "version": "2.1.0-beta.2",
+      "version": "2.1.0-beta.3",
       "license": "Apache-2.0",
       "devDependencies": {
         "husky": "^4.3.0",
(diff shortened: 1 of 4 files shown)


kriswest:
Apologies @hughtroeger @mattjamieson, looks like the last PR I merged broke the test (`npm run test`) this fixes them again.

kriswest:
QuickType PR for exposing the prefer-const-values setting: https://github.com/glideapps/quicktype/pull/2426

To use a custom quicktype change s2tQuickTypeUtil.js to point `quicktypeExec` at the local version:

```js
// Normalise path to local quicktype executable.
const quicktypeExec = "node " + ["..","quicktype","dist","index.js"].join(path.sep);
//const quicktypeExec = ['.', 'node_modules', '.bin', 'quicktype'].join(path.sep);

const command = `${quicktypeExec} --prefer-const-values --no-combine-classes -s schema -o ${outputFile} ${sources}`;
```

Current version:
https://github.com/finos/FDC3/blob/d1273644eb41155013672edf17dc96e320f35826/s2tQuicktypeUtil.js#L43-L46

kriswest:
THis has not fixed the issue as the package scripts are regenerating the context types during the build step. We'll either need to disable that or wait until Quick type merges the option to expose the `--prefer-const-values` setting...

kriswest:
Ok, its passing now - as long as the types are not regenerated. Here's hoping quicktype merge that PR soon

hughtroeger (commented):
Actually, why is there a `package-lock.json` change when the `package.json` doesn't have any changed dependencies? Is that intentional?

kriswest:
> Actually, why is there a `package-lock.json` change when the `package.json` doesn't have any changed dependencies? Is that intentional?

`npm audit fix` did that. Its basically the same set of changes to resolve a critical vulnerability that dependabot wants to make in this PR: https://github.com/finos/FDC3/pull/1089
I'm not sure why it overrides the resolutions in the lock file rather than updating the package.json - but we can do that manually...

kriswest:
@hughtroeger  its because is a transitive dependency, i.e. its coming in through tsdx, husky and jest:

```bash
PS C:\Users\Kris\Documents\code\FDC3> npm ls @babel/code-frame
@finos/fdc3@2.1.0-beta.3 C:\Users\Kris\Documents\code\FDC3
├─┬ husky@4.3.8
│ └─┬ cosmiconfig@7.1.0
│   └─┬ parse-json@5.2.0
│     └── @babel/code-frame@7.22.13 deduped
├─┬ jest-mock-extended@1.0.18
│ └─┬ jest@25.5.4
│   └─┬ @jest/core@25.5.4
│     └─┬ jest-message-util@25.5.0
│       └── @babel/code-frame@7.22.13 deduped
└─┬ tsdx@0.14.1
  ├─┬ @babel/core@7.21.4
  │ ├── @babel/code-frame@7.22.13
  │ └─┬ @babel/template@7.22.15
  │   └── @babel/code-frame@7.22.13 deduped
  ├─┬ @babel/traverse@7.23.2
  │ └── @babel/code-frame@7.22.13 deduped
  ├─┬ babel-eslint@10.1.0
  │ └── @babel/code-frame@7.22.13 deduped
  ├─┬ eslint@6.8.0
  │ └── @babel/code-frame@7.22.13 deduped
  └─┬ rollup-plugin-terser@5.3.1
    └── @babel/code-frame@7.22.13 deduped
``` 

It'll get sorted properly when we replace `tsdx` with something better maintained (see https://github.com/finos/FDC3/issues/1061)

hughtroeger:
Ahh ok that makes sense. Looks good to me then.

mattjamieson (approved):
LGTM!

PR #1093, opened 2023-10-20
Switch to unions from enums for constrained string types in generated source files

kriswest:
After some testing I've found that the enums that QuickType generates for constrained string types are not ideal and cause typing issues between types that should be compatible (const strings don't match to enums that contain them). This leads to using less specific types and hence more errors. 

This PR switches the code generation to generating union types for fields that can have several string values instead of enums. The advantage to an enum was that you could reference the enum value in your code, but it would also make polymorphism difficult as the more specific types (with a const for the value) were not compatible with less specific types (which had an enum of the possible values). 

After the switch I had to update some code where I was using the generated types (specifically wherever I was referencing those enums), but also, after moving to more specific types I could now use, I immediately found some errors that the types were not picking up because of how I had to use them.


mattjamieson (approved):
LGTM, thanks!

hughtroeger (approved):

PR #1094, opened 2023-10-23
Update FINOS IP policy link

maoo:
Linking to FINOS IP policy of current latest commit SHA; this will guarantee that the FINOS IP Policy content is unaltered, also after the publication its latest version.


kriswest:
@maoo makes sense to me (we don't want updates as this is about past version being under the IP policy. While reviewing I noted that the default site version still ahs the old charter, I'll add a small change to this PR to make the nav always use the latest version (as that's what is most relevant).

kriswest:
@maoo scratch that - I'll go back and update the 2.1 (current version) charter doc as you can't link cross-version in a sidebar

maoo:
FYI - The new IP Policy file was updated, see https://github.com/finos/community/pull/289

kriswest:
@maoo you've linked to the old policy, right? As that's what the contributions were made under in the past, where future ones are under the CSL instead, right?

kriswest:
Looks like you have so this LGTM

kriswest (approved):
LGTM

PR #1102, opened 2023-11-10
remove union type for chat message

Yannick-Malins:
force the message part of a chat message to be a types structure, remove the union type that was kept for initial compatibility

update documentation

Issue #1068, opened 2023-09-20
Replace existing Context pages with JSON schema definitions


mistryvinay:
### Use Case:
Generate Docusaurus pages for Context types using JSON schema files.

Proposal already created for markdown docs, which can be found here #1020 , which is included as part of release 2.1

### Work required
Either:
- Script to generate static markdown documentation from JSON schema files
- Build or integrate a JSON Schema viewer solution

and:
- Generate the Docusaurus front matter, programmatically modify the sidebars file to make sure that they're linked and they're in the website or the viewer.

### Additional Information
Discussion around this particular enhancement is being tracked via #940 and #1059

kriswest:
Thanks for raising this @mistryvinay, I made a slight tweak to the description (docs are in 2.1 release and we could either build a viewer or generate static markdown)

kriswest:
- JSON Schema documentation rendering notes from @kriswest 
  - JSON Schema -> Markdown or HTML conversion:
    - <https://tomcollins.github.io/json-schema-static-docs/>
      - Node module
      - Generates static markdown docs
      - Example: 
        - HTML: <https://tomcollins.github.io/json-schema-static-docs/examples/person.html>
        - Generated markdown source: <https://raw.githubusercontent.com/tomcollins/json-schema-static-docs/master/gh-pages/examples/person.md>
    - <https://pypi.org/project/json-schema-for-humans/>
      - Python module
      - Docs: https://coveooss.github.io/json-schema-for-humans/#/
      - Only supports up to JSON Schema Draft-07
      - Can generate JS/HTML or Markdown, number of template options
      - Examples: <https://coveooss.github.io/json-schema-for-humans/#/Examples>
    - <https://github.com/CesiumGS/wetzel>
      - <https://github.com/CesiumGS/wetzel#example>
      - Node module
      - Generates static markdown docs
      - Example: <https://github.com/CesiumGS/wetzel#example>
    - <https://github.com/interagent/prmd>
      - Ruby
      - Generates static markdown docs
    - <https://github.com/BrianWendt/json-schema-md-doc>
      - Node module
      - Generates static markdown docs
      - Example: 
        - Live: <https://brianwendt.github.io/json-schema-md-doc>
    - <https://github.com/adobe/jsonschema2md>
    - <https://www.npmjs.com/package/json-schema-to-markdown>
    - <https://npmjs.com/package/json-schema-to-markdown-table>
  - JSON Schema viewers (have to be embedded on a static page)
[...]

Issue #940, opened 2023-04-05
Should we split context & intent out of the main FDC3 standard?


mistryvinay:
### Question Area
[ ] App Directory
[ ] API
[x] Context Data
[x] Intents
[ ] Use Cases
[ ] Other

### Question
Should we split context & intent out of the main FDC3 standard? This would remove the reliance on a release schedule which is aligned with the main API's.  Improving the frequency of when new context and intents could be made available.
- Where would the context and intents live
- How they should be versioned

mistryvinay:
Also consider releasing NPM module as a 'next' version with the newly generated types.

_Originally posted by @kriswest in https://github.com/finos/FDC3/issues/749#issuecomment-1205190667_

robmoffat:
I think there's a lot of advantage to this.   Personally, I would like to see more of a "repository"-style approach to CD&I, as I think having just "standard" and "proprietary" doesn't quite cut it.  I would like to see:

- _private_ context types, maybe used within a single app or a single vendor's apps.
- _published_ context types, which can be used to interop with a given vendor's data structures,   e.g. "symphony.chat" or "chartiq.chart"
- _standardized_ context types, which are within the "fdc3.*" namespace like "fdc3.contact"

This would allow people to get on with sharing context and interop'ing without having to necessarily wait for standards to happen.

## Versioning

Ideally, if we published these context types into a repository, we could version them all independently.  For example, symphony might have multiple versions of 'symphony.chat'.  We could use the existing `type` field on the context data for this  'symphony.chat@v1', say or add a `version` parameter.

robmoffat:
## Where To Publish Types

Let's say we go with the idea in my earlier comment and allow anyone to publish types:

| - private context types, maybe used within a single app or a single vendor's apps.
| - published context types, which can be used to interop with a given vendor's data structures, e.g. "symphony.chat" or "chartiq.chart"
| - standardized context types, which are within the "fdc3.*" namespace like "fdc3.contact"

For this to work, we need some kind of namespace.  

We could use a java-style namespace technique.  So, for example a Finsemble-published type might be `com.finsemble.fdc3.thing`, since Finsemble own the `finsemble.com` domain name.

This is the simplest suggestion.

- What should we do about Schema discovery?  Do we even care about that?

robmoffat:
Discussion from the last CD&I, and current maintainer's meeting:

- Ignore (for now) difficulties around namespacing / versioning
[...]

PR #1103, opened 2023-11-10
Update workbench for 2.1

### LINKED ISSUE #1087: Update FDC3 workbench for FDC3 2.1
The FDC3 workbench needs to be updated for FDC3 2.1:
- FDC3 Workbench currently reports a version mismatch if you the DA reports version 2.1. As 2.0 and 2.1 are very similar (difference don't affect the workbench) it should accept either version in the 2.0 tab.
- Context templates and Intent list should be updated to include new ones added in 2.1

kriswest:
resolves #1087 

- Renames 2.0 version to 2.0+
- Auto-select 2.0 version if 2.1 DA version is reported
- Adds context templates for types added in 2.1
- Adds an alert under addIntentListener to indicate that DAs may require intents to be added to the apps appD record to be usable
- Adds intents standardized in 2.1

### DIFF
--- a/toolbox/fdc3-workbench/src/fixtures/contexts.ts
+++ b/toolbox/fdc3-workbench/src/fixtures/contexts.ts
@@ -92,7 +92,21 @@ export const contexts: ContextItem[] = [
 					streamId: "j75xqXy25NBOdacUI3FNBH"
 				}
 			},
-			message: "A message to send"
+			message: {
+				type: 'fdc3.message',
+				text: {
+				  'text/plain': 'Hey all, can we discuss the issue together? I attached a screenshot'
+				},
+				entities: {
+				   0: {
+					   type: 'fdc3.fileAttachment',
+						data: {
+						name: 'myImage.png',
+							  dataUri: 'data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAgAAAAIAQMAAAD+wSzIAAAABlBMVEX///+/v7+jQ3Y5AAAADklEQVQI12P4AIX8EAgALgAD/aNpbtEAAAAASUVORK5CYII'
+						}
+					}
+				}
+			  }
 		},
 		schemaUrl: new URL("https://fdc3.finos.org/schemas/next/context/chatMessage.schema.json"),
 	},
--- a/toolbox/fdc3-workbench/src/fixtures/contexts.ts
+++ b/toolbox/fdc3-workbench/src/fixtures/contexts.ts
@@ -8,6 +8,138 @@ import { v4 as uuidv4 } from "uuid";
 export const contexts: ContextItem[] = [
 	{
 		uuid: uuidv4(),
+		id: "Chart example",
+		template: {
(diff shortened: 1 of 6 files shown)


toolbox/fdc3-workbench/src/fixtures/contexts.ts
```diff
@@ -6,6 +6,138 @@ import { ContextItem } from "../store/ContextStore";
...
+				}
+			],
+			range: {
+				type: "fdc3.timeRange",
+				startTime: "2020-09-01T08:00:00.000Z",
+				endTime: "2020-10-31T08:00:00.000Z"
+			},
+			style: "line"
+		},
+		schemaUrl: new URL("https://fdc3.finos.org/schemas/next/context/chart.schema.json"),
```
bingenito, line 35: Should these be next or should it be a specific version to make sure they aren't changed underneath?

bingenito: Given this is existing pattern, it doesn't matter for this change but a perhaps a future discussion.

toolbox/fdc3-workbench/src/fixtures/contexts.ts
```diff
@@ -6,6 +6,138 @@ import { ContextItem } from "../store/ContextStore";
...
+		template:{
+			type: "fdc3.chat.message",
+			chatRoom: {
+				type: 'fdc3.chat.room',
+				providerName: "Symphony",
+				id: {
+					streamId: "j75xqXy25NBOdacUI3FNBH"
+				}
+			},
+			message: "A message to send"
```
bingenito, line 95: @kriswest  This the old schema?  Should we merge #1102 and then update this, or will that PR be updated to update this sample?

kriswest: Updated (after #1102 was merged)

toolbox/fdc3-workbench/src/fixtures/contexts.ts
kriswest: It's worth a future discussion - as the general rules around context don't change often I (personally) don't think the tie to the same standard version and cycle as the API(s) makes sense and hence that it would be better if we can pull in the examples from the standard.

There's an issue representing that question at https://github.com/finos/FDC3/issues/940 and we've started taking steps that might enable us to move to a different system of governance, e.g. reducing the work you have to do to submit a type (down to just a single schema file that contains the schema, docs and examples - see: https://github.com/finos/FDC3/issues/1068). We could eventually build a tool that would help produce the schema from an example pasted into a form, buy using one of several libs that will propose a JSONSchema based on an example JSON.  We could then enable (and require!) the addition of docs etc. and make it really simple and quick to submit a proposal PR, automate publishing or proposed, standard and even proprietary types...

bingenito (approved):

hughtroeger (approved):
Tried it out locally, everything looks good!

PR #1117, opened 2023-11-30
Update NOTICES.md

robmoffat:


kriswest (commented):
Should the licensee name actually be FINOS? 

Whats a system identifier? FDC3 Sail?

robmoffat:
@maoo do you know what this should really say?

kriswest:
@robmoffat @maoo I've gone with the name of the entity (in our case interop.io) as the licensee, as I'm not licensing it personally. However, my contact details are down - I think the contact details used should be those of a CLA manager.

https://github.com/finos/FDC3/pull/1128/files

This seems to me like a step you should take immediately after signing a CLA - and is one you can only complete after doing so. Hence, you could start this PR to check your status, get a CLA signed, get the easyCLA check to pass, merge the PR. Presumably that is the intended process...

PR #1126, opened 2023-12-06
Clarified expected behaviour for addContextListener

openfin-johans:
Added a blurb to clarify which context should be returned when using null as the type.


kriswest:
@openfin-johans can you make the changes in the /docs folder as well or we'll lose them the next time we release a version - you'll then be able to see them in the main/unreleased version of the site.

website/versioned_docs/version-2.0/api/ref/DesktopAgent.md
```diff
@@ -63,7 +63,7 @@ addContextListener(contextType: string | null, handler: ContextHandler): Promise
...
 
-Context broadcasts are only received from apps that are joined to the same User Channel as the listening application, hence, if the application is not currently joined to a User Channel no broadcasts will be received. If this function is called after the app has already joined a channel and the channel already contains context that would be passed to the context listener, then it will be called immediately with that context.
+Context broadcasts are primarily received from apps that are joined to the same User Channel as the listening application, hence, if the application is not currently joined to a User Channel no broadcasts will be received from channels. If this function is called after the app has already joined a channel and the channel already contains context that matches the type of the context listener, then it will be called immediately and the context passed to the handler function. If `null` was passed as the context type for the listener and the channel contains context, then the handler function will be called immediately with the most recent context - regardless of type.
```
kriswest, line 66:
```suggestion
Context broadcasts are primarily received from apps that are joined to the same User Channel as the listening application, hence, if the application is not currently joined to a User Channel no broadcasts will be received from User channels. If this function is called after the app has already joined a channel and the channel already contains context that matches the type of the context listener, then it will be called immediately and the context passed to the handler function. If `null` was passed as the context type for the listener and the channel contains context, then the handler function will be called immediately with the most recent context - regardless of type.
```

kriswest (changes requested):
Please add the same change to /docs/api/ref/DesktopAgent.md to avid it getting lost in future versions

kriswest (approved):
LGTM!

hughtroeger (approved):

PR #1127, opened 2023-12-06
Correcting IntentResult error in FDC3 2.0 and adding a very minor clarification to 2.1/next

### LINKED ISSUE #1107: Correction to the 2.0 IntentResult?
The IntentResult in 2.1 includes the following:

`type IntentResult = Context | Channel | void;`

This was meant to be the case for 2.0 as well but was somehow not added in all the right places during the implementation/documentation of 2.0:

https://fdc3.finos.org/docs/2.0/api/ref/Types#intentresult

The 2.0 Conformance tests actually check for this particular behaviour and even though it's no longer the active version - it make sense to update the docs and the NPM packages to line up with the original intention rather then having it as a change in 2.1.

https://github.com/finos/FDC3-conformance-framework/blob/main/src/test/v2.0/advanced/fdc3.raiseIntent-Result.ts#L23

kriswest:
resolves #1107

Updates FDC3 2.0's docs to match the correction applied in 2.1 and adds a minor clarification in the 2.1 and next specs to carry over into future versions.

An additional branch and PR will be needed to correct the type in a 2.0 NPM module release - although most will move to the 2.1 module (which has the correction and is backwards compatible with 2.0). We can do that release to NPM after this change is merged and approved.

### DIFF
--- a/docs/api/spec.md
+++ b/docs/api/spec.md
@@ -226,7 +226,7 @@ As an alternative to raising a specific intent, you may also raise an unspecifie
 
 ### Intent Results
 
-An optional [`IntentResult`](ref/Types#intentresult) may also be returned as output by an application handling an intent. Results maybe either a single `Context` object, or a `Channel` that may be used to send a stream of responses. The [`PrivateChannel`](ref/PrivateChannel) type is provided to support synchronization of data transmitted over returned channels, by allowing both parties to listen for events denoting subscription and unsubscription from the returned channel. `PrivateChannels` are only retrievable via [raising an intent](ref/DesktopAgent#raiseintent).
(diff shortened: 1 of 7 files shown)


openfin-johans (approved):
Looks all good to me!

hughtroeger (approved):

kriswest:
@bingenito I understand your comment on this - it is a breaking change. It's one we already made in FDC3 2.1 (and without incrementing the major version number), as it wasn't what was intended in 2.0 (due to a mistake on my part). 

When we agreed to change this in 2.1 there was a feeling that this feature wasn't heavily used (yet) and hence the sooner we fix it the better.  Based on your comment, I think that still holds (although less so for those already using the .NET types) - but we very much need to avoid it happening again in the future, before those types are released under a finos/fdc3 namespace. Do you concur?

I suspect having you on the team to help us with reviews might help catch similar things (that matter more in .NET than JS/TS) will help!

@bingenito would implementing this in .NET be a case of changing the `IntentResolution` interface to:

```csharp
public interface IIntentResolution
    {
        ...
        Task<IIntentResult?> GetResult();
    }
```

i.e. adding the `?` as a hint that its possibly null.

That said I do hear there are newer approaches in later c# language versions (and no where near competent in the old ones, let along new ;-) )

bingenito (approved):

PR #1139, opened 2023-12-22
Refactors ContextTypes&Intent to union type instead of enum

### LINKED ISSUE #1138: Refactoring ContextTypes to union type (instead of enum)
Currently the standard FDC3 Context types are defined as an enum:
https://github.com/finos/FDC3/blob/43ce7b2c17b80b9117a339950789867641ea0929/src/context/ContextType.ts#L5-L25

This causes problems when constructing mapped and/or conditional types, e.g. generically typing a function so that for a given ContextType the corresponding Context structure is returned (``fdc3.chart'` -> `Chart`, ``fdc3.action'` -> `Action`, etc.).

The `ContextType` defined here is not helping, as it's NOT a valid union type:
https://github.com/finos/FDC3/blob/43ce7b2c17b80b9117a339950789867641ea0929/src/context/ContextType.ts#L27

Also, in current versions of TypeScript everything enum types offer can be achieved with union types (and much more) without the inherent problems that TypeScript Enums have: https://www.typescriptlang.org/docs/handbook/enums.html#const-enum-pitfalls

My suggestion is to replace the current implementation with a simple union type.
Possibly also clearly differentiate between standard FDC3 context types and custom ones:

```ts
export type StandardContextTypes =
  | 'fdc3.action'
  | 'fdc3.chart'
  | 'fdc3.chat.initSettings'
  | 'fdc3.chat.message'
  | 'fdc3.chat.room'
  | 'fdc3.chat.searchCriteria'
  | 'fdc3.contact'
  | 'fdc3.contactList'
  | 'fdc3.country'
  | 'fdc3.currency'
  | 'fdc3.email'
  | 'fdc3.instrument'
  | 'fdc3.instrumentList'
  | 'fdc3.interaction'
  | 'fdc3.message'
  | 'fdc3.organization'
  | 'fdc3.portfolio'
  | 'fdc3.position'
  | 'fdc3.nothing'
  | 'fdc3.timerange'
  | 'fdc3.transactionResult'
  | 'fdc3.valuation';

export type ContextTypes = StandardContextTypes | string;
```

Currently, I need to define these types in all applications that interact with FDC3, but it would be nice to have them in the main library.

andreifloricel:
closes #1138


src/context/ContextType.ts
```diff
@@ -2,26 +2,35 @@
...
+  | 'fdc3.position'
+  | 'fdc3.nothing'
+  | 'fdc3.timerange'
+  | 'fdc3.transactionResult'
+  | 'fdc3.valuation';
+
+/**
+ * @see https://fdc3.finos.org/docs/context/spec
+ */
+export type ContextType = StandardContextType | string;
```
kriswest, line 36: We're missing the experimental types in this list - we should add them while making this change:

```suggestion
export type ExperimentalContextType = 
  | 'fdc3.order'
  | 'fdc3.orderList'
  | 'fdc3.product'
  | 'fdc3.trade'
  | 'fdc3.tadeList';

/**
 * @see https://fdc3.finos.org/docs/context/spec
 */
export type ContextType = StandardContextType | ExperimentalContextType | string;
```

kriswest (approved):
This change LGTM - although I think we should consider whether to add an array or use an array instead to simplify implementation of a runtime check as to whether a context is a standard type.

andreifloricel:
Sorry, this PR has slipped under my radar over the last few weeks.
I'll update it with suggested improvements in the coming days.

bingenito:
> This change LGTM - although I think we should consider whether to add an array or use an array instead to simplify implementation of a runtime check as to whether a context is a standard type.

@kriswest Do you have a use case in mind where you would need to know if it is a standard type?

kriswest:
@andreifloricel could outline the use case for this type and confirm or deny whether you see a use case for a util function to check if a context is a standard type at runtime? If we can confirm that I can probably document it and add in a util fn to do it (which may need an array or the type names).

Without a use case, one of the other maintainers suggested dropping the enum/union entirely... I sort of see a use case - but I'd much rather confirmation from someone using it for real!

src/intents/IntentsConfiguration.ts
```diff
@@ -0,0 +1,69 @@
...
+    'fdc3.instrumentList',
+    'fdc3.organization',
+    'fdc3.portfolio',
+    'fdc3.nothing',
+  ],
+  ViewOrders: ['fdc3.contact', 'fdc3.instrument', 'fdc3.organization', 'fdc3.nothing'],
+  ViewProfile: ['fdc3.contact', 'fdc3.organization', 'fdc3.nothing'],
+  ViewQuote: ['fdc3.instrument', 'fdc3.nothing'],
+  ViewResearch: ['fdc3.contact', 'fdc3.instrument', 'fdc3.organization', 'fdc3.nothing'],
+};
```
andreifloricel, line 37: I tried to disable Prettier processing for this file via `.prettierignore` or a local ESLint file, to no avail
And TSDX seems to be in maintenance mode...

bingenito:
Isn't this a semver major breaking api change to anyone using the old style constants like below?
```
  window.fdc3.addContextListener(ContextTypes.Instrument, () => {});
  window.fdc3.addIntentListener(Intents.ViewInstrument, () => {});
````

**Comment by bingenito** [comment]
> * added utility functions: `isStandardContextType(contextType: string)`,`isStandardIntent(intent: string)`,`getPossibleContextsForIntent(intent: StandardIntent)`

I don't understand the use case for these.  Do you have a local implementation dependent on these?

**Comment by andreifloricel** [comment]
> Isn't this a semver major breaking api change to anyone using the old style constants like below?
> 
> ```
>   import { ContextTypes, Intents } from "@finos/fdc3";
>   window.fdc3.addContextListener(ContextTypes.Instrument, () => {});
>   window.fdc3.addIntentListener(Intents.ViewInstrument, () => {});
> ```

No, all changes in this PR are backwards compatible:
 - ContextType and Intent types still accept any string value ( as it must support custom Contexts/Intent)
 - also, changes in the DesktopAgent Api don't affect existing implementations, see https://tsplay.dev/wX0KLw
 - the main advantages over the previous implementation are autocompletion and  and significantly better TypeScript flexibility (mapped types, type guards, etc)
  
strict_typing_backwards_compatibility

**Comment by andreifloricel** [comment]
> > * added utility functions: `isStandardContextType(contextType: string)`, `isStandardIntent(intent: string)
> 
> I don't understand the use case for these. Do you have a local implementation dependent on these?

A real use case for us is that we need to dynamically build UI controls for our users (e.g., right-click a grid cell and generate an intent for the context referenced in that specific cell). When we do this, it's very useful to know if that specific intent and/or context is a standard FDC3 one, because in that case we can make a lot of educated guesses. Such as, for example, deriving possible contexts for a standard intention (see the next section of the commentary).
@kriswest what use case did you have in mind?

> > `getPossibleContextsForIntent(intent: StandardIntent)`
> 
> Isn't this opinionated?

Well, it is, but it's FDC3's opinion :)
As I understand the FDC3 specification, each standard intent (**_Verb_**) has an associated list of possible contexts (**_Nouns_**), e.g. https://fdc3.finos.org/docs/intents/ref/ViewChart#possible-contexts

**Comment by bingenito** [comment]
> > Isn't this a semver major breaking api change to anyone using the old style constants like below?
> > ```
> >   import { ContextTypes, Intents } from "@finos/fdc3";
> >   window.fdc3.addContextListener(ContextTypes.Instrument, () => {});
> >   window.fdc3.addIntentListener(Intents.ViewInstrument, () => {});
> > ```
> 
> No, all changes in this PR are backwards compatible:
> 
> * ContextType and Intent types still accept any string value ( as it must support custom Contexts/Intent)
> * also, changes in the DesktopAgent Api don't affect existing implementations, see https://tsplay.dev/wX0KLw

I agree with the change and type guard, but it is a breaking change on the caller side as ContextTypes and Intents is no longer exported.  Put the above in a new file in the project and it is fine.  Switch to this branch and it has errors for missing exports.

**Comment by andreifloricel** [comment]
> I agree with the change and type guard, but it is a breaking change on the caller side as ContextTypes and Intents is no longer exported.

Good spot! I reintroduced the deprecated enum types but marked them as deprecated.

**Comment by kriswest** [comment]
Shall we pick this up briefly at tomorrow's Standards Working Group meeting? Happy to add it to the agenda - I haven't yet caught up on the conversation.

**Comment by andreifloricel** [comment]
Hi @kriswest,

since I can't attend today's Standards Working Group meeting, I'll try to to summarize all the changes in this task:

- marked existing `ContextTypes` and `Intents` enums as deprecated
- introduced new union types `ContextType` and `Intent` (supporting both standard and custom values)
- adjusted all `DesktopAgent` Intent and ContextType references from to the new union types
- added utility functions: 
  - isStandardContextType(contextType: string)
  - isStandardIntent(intent: string)
  - getPossibleContextsForIntent(intent: StandardIntent)

Note:
 - I intentionally kept the public types for Intent&ContextType "clean" and readable and put the internal types and configurations in IntentsConfiguration.ts
 - nevertheless, the public types (Union Types) are the single source of truth: once you add/remove/adjust something there, the compilation __~~fails~~__ _(should fail, but it does NOT, see next point)_ until the IntentsConfiguration.ts is updated accordingly
 - because TypeScript version is quite old and _TSDX_ library is more or less dead and unmaintained, I couldn't type the Intent to Context mapping correctly (using the `satisfies` construct); it's a matter of time before other things break down, I think migration to another tool should be considered (Vite?)

File: src/api/Methods.ts [hunk 3]
```diff
@@ -178,6 +184,30 @@ export function findInstances(app: AppIdentifier): Promise<AppIdentifier[]> {
...
+ */
+export function isStandardIntent(intent: Intent): intent is StandardIntent {
+  return StandardIntentsSet.has(intent as StandardIntent);
+}
+
+/**
+ * Get the possible context types for a given intent.
+ * @param intent
+ */
+export function getPossibleContextsForIntent<I extends StandardIntent>(intent: I): ContextTypeFor<I>[] {
```
bingenito, line 207: @kriswest  Correct me if I'm wrong, but discussion on the call was in favor of the isStandard* being fine but getPossibleContextsForIntent be deferred for future/separate discussion?

kriswest: Correct - the 'possible contexts' content in the intent specifications was intended to be informative but isn't restrictive (I'm told), particularly as proprietary contexts are allowed. The possible contexts don't exist in code or schema anywhere (until this PR).

Apologies, I haven't had time to circle around to fully review this issue and the discussion yet, but I have posted the minutes from the meeting at: 
- #1160

bingenito (approved):
Confirmed backcompat issue I raised is fixed.  Looks good.

hughtroeger (approved):

kriswest:
This PR when in without a changelog entry, which we will need to add before the next version of FDC3 is adopted

PR #1149, opened 2024-02-06
Identifier values should be of type string

kriswest:
Adding a note that `id` field values in context objects should always be of type string (which unfortunately can't be represented in the generated context types.)

> If you use additionalProperties: { type: "string" } optional properties in context types derived from context will not compile (as they are implicitlystring | undefined  and there is no way of expressing undefined in JSONSchema.  While using unevaluatedProperties: { type: "string" } or patternProperties: { ".+": {type: "string" }}  just result in the property value being allowed to be any  (i.e. [property: string]: any) as both allow for defined properties to differ and hence the general constraint [property: string]: string  would conflict
> 
> In short, the documentation is correct and the generated types are as close to correct as they can be with the code generation we use. To do better would require a manual change after code generation.
I think this a minor flaw as each subtype that defines id properties can and should define their type as string (as [instrument does here](https://github.com/finos/FDC3/blob/f4a46c0bf64cb72abf1343ad34ac2e6f251b000c/schemas/context/instrument.schema.json#L19-L27)). Hence, this restriction is only really applied to non-standard id entries.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
+* Added clarification that `id` field values SHOULD always be strings to context schema definition (a restriction that can't easily be represented in the generated types). ([#1149](https://github.com/finos/FDC3/pull/1149))
+
 ### Changed
 
 ### Deprecated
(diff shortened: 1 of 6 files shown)


openfin-johans (approved):
Looks good - ship it!

bingenito (approved):

hughtroeger (approved):

PR #1170, opened 2024-03-06
Remove the IntentResolution.version field as defunct

### LINKED ISSUE #1053: Remove defunct version field from IntentResolution
The IntentResolution type: https://fdc3.finos.org/docs/api/ref/Metadata#intentresolution
has a version field with no purpose - there are no schema versions for intents, nor version fields in any of the metadata fields nor appD records. Hence, this field serves no purpose and should be removed

kriswest:
resolves #1053

Removed the `version` field from `IntentResolution` as there are no version fields for intents in the FDC3 API definitions and hence the field has no purpose.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 
+* Removed the `version` field from `IntentResolution` as there are no version fields for intents in the FDC3 API definitions and hence the field has no purpose. ([#1170](https://github.com/finos/FDC3/pull/1170))
+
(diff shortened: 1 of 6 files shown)


bingenito:
Does this need to be explicitly removed from markdown too?

https://github.com/finos/FDC3/blob/master/docs/api/ref/Metadata.md?plain=1#L324

kriswest:
@bingenito Yes it does. You'd think I'd know how to do this by now 🤦‍♂️

kriswest:
@bingenito fixed the docs reference, and a schema reference to the field that had made it into bridging.

bingenito (approved):

hughtroeger (approved):

PR #1187, opened 2024-04-11
Temporary solution to fix linting (until tsdx is replaced)

kriswest:
linting is fail due to a mismatch between the eslint and typescript versions used in the project (due to tsdx rotting). This will be resolved by:
- #1175 

Until then, this PR adds an override to ensure a newer version of eslint is installed, allowing the commit hook that runs the linter to run.

@maoo @robmoffat can one of you try this out, approve the review and merge? Once done, updating the https://github.com/TheJuanAndOnly99/FDC3/tree/issue-1068 branch with that change should get you unstuck again

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@finos/fdc3",
-  "version": "2.1.0-beta.6",
+  "version": "2.1.0-beta.7",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@finos/fdc3",
-      "version": "2.1.0-beta.6",
+      "version": "2.1.0-beta.7",
       "license": "Apache-2.0",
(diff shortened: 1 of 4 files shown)


maoo (approved):
LGTM - even though - in all fairness - I'm not sure I understand why the issue is occurring and therefore why this change should address it.

Happy to test a commit after the merge tho!

kriswest:
@maoo the issue was caused by a mismatch between the old eslint version tsdx was using and the typescript version. Once code from a newer version of typescript appeared in the project it caused an error in eslint (that has long since been fixed). A very good example of why we need to get the PR to update the project's build system (#1175) merged ASAP!

kriswest:
@maoo  merged -, merge main into your branch in the other PR and you should be able to commit again (with a few warnings).

PR #1193, opened 2024-04-18
Small bug in FDC3 Workbench, doesn't work with hierarchical classes.

robmoffat:


kriswest (approved):
Can you elaborate on this a hair - I think I understand but wanted to confirm. Is your IntentResult returned through the API extending another something else that's providing the `getResult` function (leading to hasOwnProperty not finding it)?

Approving as that code change is not going to hurt anything and would fix that above if that's whats happening

bingenito (approved):

PR #1198, opened 2024-04-26
Only trigger NPM releases on release/* branch merges

kriswest:
This PR remove the trigger for NPM releases on merges to the main branch, meaning that all NPM release going forward will come from release/* branches. This is a more explicit workflow which should be safer and makes it easier to do maintenance releases on past FDC3 versions when errors in them are identified.

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -16,7 +16,9 @@ on:
       - 'docs/**'
       - 'website/**'
   pull_request:
+    types: [opened, reopened, edited, ready_for_review]
     branches:
+      - main
       - release/*
     paths-ignore:
       - '.github/workflows/docusaurus.yml'
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -7,7 +7,6 @@ name: npm-build
 on:
   push:
     branches:
-      - main
       - release/**
     paths-ignore:
       - '.github/workflows/docusaurus.yml'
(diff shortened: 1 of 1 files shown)


bingenito:
You missed the top level file comments about triggering on main branch.

.github/workflows/package.yml
```diff
@@ -7,15 +7,13 @@ name: npm-build
 on:
   push:
     branches:
-      - main
       - release/**
     paths-ignore:
       - '.github/workflows/docusaurus.yml'
       - 'docs/**'
       - 'website/**'
   pull_request:
```
bingenito, line 15: This confused me as I was wondering why we are doing any publishing flow for PRs but further down, the actual publish job checks why we are running the action.  Not for this PR, but we should discuss refactoring this into a workflow that just does build and then include that into this publishing one to make the on triggers easier to understand.

kriswest: I'm no github workflow expert - I think Riko originally set this up. I beleive it runs on PRs and acts as a check that the build isn't broken, and yes the publish is inhibited on pull_request runs.

Now that you say that I think the second removal on line 18 is wrong as  it will stop it running on PRs so we lose that check.

It would indeed make sense to split these workflows up. For now I'll back out the change to the pull_request trigger.

kriswest: Fixed (but tightened up the PR event that it runs on so save FINOS some pennies)

kriswest:
> You missed the top level file comments about triggering on main branch.

Thank you! Added better notes.

bingenito (approved):

PR #1200, opened 2024-04-29
Don't use union types in context definitions

### LINKED ISSUE #1105: FDC3  should avoid union types in contexts and other objects
Both Typescript and JSON Schema allow for a type of polymorphism in types and interfaces that is hard to represent in other languages: allowing the type of a variable to be a union of other, unrelated types. E.g.: in typescript

```ts
type Example = SomeOtherType | YetAnotherType;
```

Similar things are possible in JSON Schema by using the `anyOf` or `oneOf` keywords to type a field.

This can sometimes be handled via a marker interface in another language (e.g. .NET or Java), i.e having a class or interface extend/implement an interface with little or no functionality, just so it can be used as the type for the field, allowing any of the classes that 'implement' it to be set as values. However, this approach is not possible if one of the types in the union is a primitive, meaning it's not a class and can't be modifed, e.g.:

```ts
type Example2 = SomeOtherType | string;
```

Hence, if FDC3 is intended to be implemented in other languages it should avoid, whereever possible, the use of constructs like this cthat cannot easily be represented in other languages - particularly .NET and Java. 

Add guidance/a strong recommendation to the context specification: https://fdc3.finos.org/docs/context/spec#field-type-conventions, https://fdc3.finos.org/docs/supported-platforms and https://fdc3.finos.org/docs/api/spec#desktop-agent-implementation that FDC3 SHOULD NOT use union types or other constructs that are hard to represent in languages other than TypeScript or JSON Schema to ensure ease of implementation in those other languages.

kriswest:
resolves #1105

Adds content to the Context and API overviews about not using unions types (particularly of primitive types) as they not easily implementable in other languages.

### DIFF
--- a/docs/context/spec.md
+++ b/docs/context/spec.md
@@ -104,6 +104,24 @@ It may be as simple as adding an optional `$version` property to types, but it c
 
 This Standard defines a number of conventions for the fields of context types that all context objects SHOULD adhere to in order to reduce or prevent competing conventions from being established in both standardized types and proprietary types created by app developers.
 
+#### Avoid union types / composition of primitive types
+
+Both Typescript and JSON Schema allow for a type of polymorphism in types and interfaces that is hard to represent in other languages: allowing the type of a variable to be a union of other, unrelated types. E.g.: in typescript
+
+```ts
+type Example = SomeOtherType | YetAnotherType;
+```
+
+Similar constructs are allowed in JSON Schema by using the `anyOf` or `oneOf` keywords to specify that a value can take the form defined in one-or-more or one-of-several sub-schemas.
+
+However, other languages can be less less-flexible. In most languages, polymorphism of object types is possible via the implementation and/or extension of an interface (for example all context types are derived from the [Context](ref/Context) schema, which can be modelled as an interface). However, this approach is not possible if one of the types in the union is a primitive, meaning it's not a class and can't be modified to implement an interface, e.g.:
+
+```ts
+type Example2 = SomeOtherType | number;
+```
+
+Hence, to ensure that FDC3 context objects are implementable in other languages context schemas MUST NOT use `anyOf`/`oneOf` compositions of primitive types in JSON schema (and, hence, unions of primitive types in TypeScript) and SHOULD avoid compositions of Object types unless a defined interface (such as [Context](ref/Context)).
+
 #### Identifiers
 
 An `id` field with type `object` is defined in the base [fdc3.context](ref/Context) type, from which all other context objects are derived, and SHOULD be used to encapsulate identifiers. Specific context types may define subfields for specific identifiers as needed.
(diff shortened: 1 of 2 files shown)


docs/context/spec.md
```diff
@@ -104,6 +104,39 @@ It may be as simple as adding an optional `$version` property to types, but it c
...
+
+Both Typescript and JSON Schema allow for a type of polymorphism in types and interfaces that is hard to represent in other languages: allowing the type of a variable to be a ['union'](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#union-types) of other, unrelated types. E.g.: in typescript
+
+```ts
+type Example = SomeOtherType | YetAnotherType;
+```
+
+Similar constructs are allowed in JSON Schema by [combining or composing schemas](https://json-schema.org/understanding-json-schema/reference/combining) using the `anyOf` or `oneOf` keywords to specify that a value can take the form defined in one-or-more or one-of-several sub-schemas.
+
+```JSON
```
kriswest, line 117:
```suggestion
```json
```

docs/context/spec.md
```diff
@@ -104,6 +104,39 @@ It may be as simple as adding an optional `$version` property to types, but it c
...
+            "$ref": "contact.schema.json#"
+        },
+        {
+            "$ref": "contactList.schema.json#"
+        }
+    ]
+}
+```
+
+However, other languages can be less less-flexible. In most languages, polymorphism of object types is possible via the implementation and/or extension of an interface (for example all context types are derived from the [Context](ref/Context) schema, which can be modelled as an interface). However, this approach is not possible if one of the types in the union is a primitive, meaning it's not a class and can't be modified to implement an interface, e.g.:
```
hughtroeger, line 132:
```suggestion
However, other languages can be less flexible. In most languages, polymorphism of object types is possible via the implementation and/or extension of an interface (for example all context types are derived from the [Context](ref/Context) schema, which can be modelled as an interface). However, this approach is not possible if one of the types in the union is a primitive, meaning it's not a class and can't be modified to implement an interface, e.g.:
```

docs/context/spec.md
```diff
@@ -104,6 +104,39 @@ It may be as simple as adding an optional `$version` property to types, but it c
 
 This Standard defines a number of conventions for the fields of context types that all context objects SHOULD adhere to in order to reduce or prevent competing conventions from being established in both standardized types and proprietary types created by app developers.
 
+#### Avoid union types / composition of primitive types
+
+Both Typescript and JSON Schema allow for a type of polymorphism in types and interfaces that is hard to represent in other languages: allowing the type of a variable to be a ['union'](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#union-types) of other, unrelated types. E.g.: in typescript
```
bingenito, line 109: nitpick - remain consistent with cased "TypeScript".

bingenito (approved):

docs/context/spec.md
kriswest:
```suggestion
Both Typescript and JSON Schema allow for a type of polymorphism in types and interfaces that is hard to represent in other languages: allowing the type of a variable to be a ['union'](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#union-types) of other, unrelated types. E.g.: in TypeScript
```

hughtroeger (approved):

PR #1212, opened 2024-05-14
Add missing argument in bridging message for findIntentByContext

### LINKED ISSUE #1211: FindIntentsByContextAgentRequest doesn't allow for supplying resultType
## Minor Issue

Interface is this:

```
export interface FindIntentsByContextAgentRequestPayload {
    context: ContextElement;
}
```

However the API is this:
```
findIntentsByContext(context: Context, resultType?: string): Promise<Array<AppIntent>>;
```

### Area of Issue
[ ] App Directory
[x] API
[ ] Context Data
[ ] Intents
[x] Desktop Agent Bridging
[ ] Use Cases
[ ] Other

Is that a bug?

kriswest:
resolves #1211


bingenito (approved):

hughtroeger (approved):

src/bridging/BridgingTypes.ts
```diff
@@ -5280,6 +5282,7 @@ const typeMap: any = {
     ], false),
     "FindIntentsByContextAgentRequestPayload": o([
         { json: "context", js: "context", typ: r("ContextElement") },
+        { json: "resultType", js: "resultType", typ: u(undefined, "") },
```
hughtroeger, line 5285: Just for my own understanding, I was wondering if `typ: u(undefined, "")` violates the 'no unions of primitive types' guidelines from this PR https://github.com/finos/FDC3/pull/1200 , but I assume this is how optional parameters are expressed here, and this is fine because it's really just an optional string.

kriswest: I know its fine in this case as we already have optional parameters elsewhere and those can generally be implemented via alternative function signatures in other languages. I'd probably also exclude null from the set of primitive types when considering this as most languages have someway of supporting that (e.g. `Nullable<int> i = null;` in C#, or using the `Integer` class in Java instead of int.

kriswest:
This PR when in without a changelog entry, which we will need to add before the next version of FDC3 is adopted

PR #1213, opened 2024-05-14
Filter CVE scans on PRs a bit more

kriswest:
These are double triggering on PRs, ~not sure why - this won't entirely solve it but will reduce it~

We were triggering on both pushes and PRs, finessed the triggers as we only accept changes via PRs and we can use the `synchronize` trigger for when those are updated.

### DIFF
--- a/.github/workflows/cve-scanning.yml
+++ b/.github/workflows/cve-scanning.yml
@@ -8,15 +8,15 @@ on:
       - 'toolbox/fdc3-workbench/package.json'
       - '.github/workflows/cve-scanning.yml'
       - 'website/package.json'
-  push:
-    paths:
-      - 'package.json'
-      - 'toolbox/fdc3-workbench/package.json'
-      - '.github/workflows/cve-scanning.yml'
-      - 'website/package.json'
-    schedule:
-      # Run every day at 5am and 5pm
-      - cron: '0 5,17 * * *'
+  # push:
+  #   paths:
+  #     - 'package.json'
+  #     - 'toolbox/fdc3-workbench/package.json'
+  #     - '.github/workflows/cve-scanning.yml'
+  #     - 'website/package.json'
+  schedule:
+    # Run every day at 5am and 5pm
+    - cron: '0 5,17 * * *'
 
 jobs:
   build:
--- a/.github/workflows/cve-scanning.yml
+++ b/.github/workflows/cve-scanning.yml
@@ -2,7 +2,7 @@ name: Node.js CVE Scanning
 
 on:
   pull_request:
-    types: [opened, reopened, edited, ready_for_review]
+    types: [opened, reopened, edited, synchronize, ready_for_review]
     paths:
       - 'package.json'
       - 'toolbox/fdc3-workbench/package.json'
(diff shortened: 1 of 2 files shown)


kriswest:
@robmoffat this should be good to - you use 'synchronize' to run whenever a PR is updated. Applied to both CVE scans and static code analysis

.github/workflows/cve-scanning.yml
```diff
@@ -2,20 +2,21 @@ name: Node.js CVE Scanning
 
 on:
   pull_request:
+    types: [opened, reopened, edited, synchronize, ready_for_review]
```
bingenito, line 5: Can you explain the intent here?  What was happening when it was just all pull_request across all branches for the filtered paths? The repetition of removing push checks out and keeping the schedule run is good.  Based on prior indentation was the cron schedule even running?

kriswest: @bingenito 
I noted that the CVE scans were running twice for every push to PR, I did 3 pushes when updating one and generated 6 runs in rapid succession - a waste of resources for FINOS, but was also generating double notificaitons on issues. As we only accept changes through PRs due to branch protection the push trigger can go and be replaced by triggering on:

- open (first run)
- reopened (recheck as time may have passed and CVE may have developed)
- edited (target branch changed - not sure if this is needed or not)
- synchronize: PR branch was pushed to
- ready_for_review: converted from a draft PR (it will still run on draft pull requests - but seems sensible to re-run when converted so that status is up to date or reviewers)

The cron schedule was NOT running AFAIK - I wasn't seeing any scheduled runs. We won't know if its worked until merged as that only applies to the default branch (main) and only when its on that branch.

kriswest: > For "edited", this can even be triggered if the title and description of the PR changes. Is that too aggressive? For ready_for_review, aren't the default activity types (opened/reopened/synchronized) already run and changing from draft to ready is basically the same as just a title change?

Yes we could probably drop edited... I think ready_for_review is a separate event from edited however and probalby worth keeping (for something thats been a draft for a while).

.github/workflows/semgrep.yml
```diff
@@ -1,6 +1,11 @@
 name: Static code analysis
 
-on: [push, pull_request]
+on:
+  pull_request:
+    types: [opened, reopened, edited, synchronize, ready_for_review]
```
kriswest, line 5:
```suggestion
    types: [opened, reopened, synchronize, ready_for_review]
```

.github/workflows/cve-scanning.yml
kriswest, line 5:
```suggestion
    types: [opened, reopened, synchronize, ready_for_review]
```

kriswest: `edited` dropped - I don't think editing the PR metadata affects anything that would change the results of the scans nor indicates that they should run again (where I think `ready_for_review` does)

bingenito: How does ready_for_review change the content of what is being reviewed?  If there were any commits or updates it would be handled by synchronize and run before it is changed out of draft?  If we remove ready_for_review we have the default list and can remove it completely.

bingenito (approved):

.github/workflows/cve-scanning.yml
bingenito: I realized after that given this is cve check there can be considerable duration between last commit and flipped to ready to review so it makes sense to make sure dependencies are ok.

PR #1216, opened 2024-05-23
Update standard-wg-meeting.md

kriswest:
Correcting CSL link in meeting agenda/minutes template - it was only wrong in the SWG meeting template and not the others.


bingenito (approved):
Looks good.  Interesting that the other two templates were already pointed at right path.

PR #1259, opened 2024-07-11
Set package write permission in package.yml

kriswest:
Update the package publish workflow in the 2.1.1 release branch. To be replicated on main if it resolves a number of issues:

- [x] Github package repo permissions (already fixed on main)
- [x] Check whether package is published in each repo (currently just checks NPM each time)
- [x] ensure build checks are not skipped but succeed so that they can pass status checks for docs and website changes (the build check is skipped unless there are /src changes)

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -7,14 +7,13 @@ name: npm-build
 on:
   push:
     branches:
-      - main
-      - release/*
+      - release/**
     paths-ignore:
       - '.github/workflows/docusaurus.yml'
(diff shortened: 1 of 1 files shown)


kriswest:
Tested here that the npm-build tasks now always run (so they can be used in status checks) but skip the actual build steps unless relevant files (matching a filter) changed. 

The Publish step remains correctly inhibited and should also only run when relvant files change AND its a push event. Publish only runs on push events to release branches (and NOT main) so that releases only come from release branches.

PR #1260, opened 2024-07-12
Fix build and publish workflow on main

kriswest:
Updates the package workflow to allow the build tasks to be used as status checks, and to ensure that the publishing workflow checks the status in each package manager repository correctly (instead of always checking NPM)

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -20,6 +20,8 @@ on:
       - release/**
     paths-ignore:
       - '.github/workflows/docusaurus.yml'
+permissions:
+  contents: read
 jobs:
 # JOB to run change detection
   changes:
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -13,8 +13,6 @@ on:
       - release/**
     paths-ignore:
       - '.github/workflows/docusaurus.yml'
-      - 'docs/**'
-      - 'website/**'
   pull_request:
     types: [opened, reopened, edited, ready_for_review]
     branches:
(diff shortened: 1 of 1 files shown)


.github/workflows/package.yml
```diff
@@ -13,24 +13,42 @@ on:
...
     types: [opened, reopened, edited, ready_for_review]
     branches:
       - main
-      - release/*
+      - release/**
     paths-ignore:
       - '.github/workflows/docusaurus.yml'
-      - 'docs/**'
-      - 'website/**'
-permissions:
```
bingenito, line 27: we still want default root permission to be read

.github/workflows/package.yml
```diff
@@ -39,28 +57,39 @@ jobs:
...
+
+
   package-publish:
-    if: ${{ github.event_name == 'push' }}
-    needs: package-build
+    if: ${{ github.event_name == 'push' && needs.changes.outputs.builtFiles == 'true'}}
+    needs: [changes,package-build]
     name: Publish package to ${{ matrix.name }}
     runs-on: ubuntu-latest
     permissions:
```
bingenito, line 95: we need permissions write

kriswest: contents: write? We already have id-token: write and packages:write here... I don't think it needs to be able to push back changes...

.github/workflows/package.yml
kriswest: I'll get that back in

.github/workflows/package.yml
bingenito: My mistake.  The diff was cutting off the lines and not showing entire permission block.  This is good.

bingenito (approved):

.github/workflows/package.yml
kriswest: done

bingenito (approved):

mistryvinay (commented):
LGTM

Issue #1255, opened 2024-07-09
fdc3.entity.fileAttachment is only defined inline  in the fdc3.message schema and example has wrong type


kriswest:
While working on generating context documentation from schemas I noticed that the `fdc3.message` type defines an `fdc3.entity.fileAttachment` type inline: https://github.com/finos/FDC3/blob/fdf060fd675a9cbbd84090e4df47cb97b6fc1ebb/schemas/context/message.schema.json#L46

and the type name does not agree with that used in the example (`fdc3.fileAttachment`):
https://github.com/finos/FDC3/blob/fdf060fd675a9cbbd84090e4df47cb97b6fc1ebb/schemas/context/message.schema.json#L89

FDC3 context types should each have their own definition and schema to make them usable and to properly reserve the type name. Hence, this definition should be separated out into its own schema and be referenced into fdc3.message.

@Yannick-Malins & @mistryvinay

PR #1261, opened 2024-07-15
1255 Make fdc3.fileAttachment an independent type

Yannick-Malins:
https://github.com/finos/FDC3/issues/1255


Yannick-Malins:
@kriswest

schemas/context/message.schema.json
```diff
@@ -38,38 +38,7 @@
                 "$ref": "action.schema.json#"
```
kriswest, line 38: @Yannick-Malins can you change the `additionalProperties.anyOf` to `oneOf` - the semantics are virtually the same, but the code generation via QuickType differs (it'll merge properties with `anyOf` as you could be multiple of the referenced types at once, where we intend it to be `oneOf` the referenced types)

kriswest (commented):
LGTM - although I think Message's entities element needs to be defined with oneOf rather than anyOf. Could you change that, add a Changelog entry and then I'll approve and merge the change (new governance restrictions will invalidate any reviews that happen before a later change)

schemas/context/fileAttachment.schema.json
```diff
@@ -0,0 +1,53 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "https://fdc3.finos.org/schemas/next/context/attachment.schema.json",
```
kriswest, line 3: the filename and this `$id` need to match, please change to: https://fdc3.finos.org/schemas/next/context/fileAttachment.schema.json

schemas/context/message.schema.json
kriswest: @Yannick-Malins The change I'm after is on line 36 above. Otherwise, this is good to merge, apart from a Changelog entry and the `$id` I flagged above.

kriswest:
@Yannick-Malins lemme know if you can make these two quick changes (I didn't add suggestions or I won't be able to review the change after you merge them) - alternatively let me know to re-raise this PR and have you review instead.

kriswest:
Todo:
- [ ] the filename and this $id need to match in fileAttachment.schema.json
- [ ] define message entities with `oneOf` rather than `anyOf` in message.schema.json
- [ ] add a changelog

kriswest:
Thanks @Yannick-Malins. I'm not seeing the new type in the preview - but that could be because the PR is '199 commits behind finos/FDC3:main'. Could you merge changes from main then run `npm run build`. That should re-generate ContextTypes.ts so that goes in and the preview should generate correctly afterwards (those two are unrelated but we need to check both).

Otherwise looks all good to me. I'll get around to adding a PR template soon so the above steps are documented!

Yannick-Malins:
should be good now!

kriswest:
LGTM, thanks @Yannick-Malins 

Docs page:
https://deploy-preview-1261--fdc3.netlify.app/docs/next/context/ref/FileAttachment

Page that links to new docs page:
https://deploy-preview-1261--fdc3.netlify.app/docs/next/context/ref/Message
image

kriswest (approved):
LGTM

bingenito (approved):

kemerava (approved):

hughtroeger (approved):

PR #1272, opened 2024-07-18
Updated for 2.2 Conformance

robmoffat:


robmoffat:
@kriswest anything else to add to this?  Perhaps something around the channel change listener?

kriswest:
superseded by #1425 which incorporates the new connection test, but also removes 1.2 and 2.0 labels and goes with just labelling new tests introduced in that version (so only 2.2 labels).

PR #1285, opened 2024-07-23
Use Pascal case for Chat context type titles

kriswest:
@bingenito THis will fix the titles/sidebar labels of the generated Chat context docs to match the current pages


bingenito (approved):
Quick turnaround.  Looks good.

PR #1289, opened 2024-07-24
Small bug fix to stop 2.1 DAs being flagged when running the 2.0 version (as its the same)

kriswest:
At present DAs that report version 2.1 are flagged as mismatch on version in the fdc3-workbench, when 2.0 and 2.1 API specs are the same - which is acknowledged in teh version selector already:

image

After the fix:

image

### DIFF
--- a/toolbox/fdc3-workbench/src/components/Header.tsx
+++ b/toolbox/fdc3-workbench/src/components/Header.tsx
@@ -166,7 +166,7 @@ export const Header = (props: { fdc3Available: boolean }) => {
 								<tr>
 									<th scope="row">FDC3 Version</th>
 									{appInfo?.fdc3Version ? (
-										chosenVersion === appInfo.fdc3Version ? (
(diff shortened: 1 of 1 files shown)


toolbox/fdc3-workbench/src/components/Header.tsx
```diff
@@ -166,7 +166,7 @@ export const Header = (props: { fdc3Available: boolean }) => {
 								<tr>
 									<th scope="row">FDC3 Version</th>
 									{appInfo?.fdc3Version ? (
-										chosenVersion === appInfo.fdc3Version ? (
+										chosenVersion === appInfo.fdc3Version || chosenVersion === "2.0" && appInfo.fdc3Version == "2.1" ? (
```
kriswest, line 169:
```suggestion
										((chosenVersion === appInfo.fdc3Version) || (chosenVersion === "2.0" && appInfo.fdc3Version === "2.1")) ? (
```

bingenito (approved):

bingenito:
Simple fix, looks good.

PR #1304, opened 2024-08-01
Drop Intent.yaml, simplify standard intents.json, add missing intents

kriswest:
/src/intents/Intent.yaml was defined early on in FDC3's history and (AFAIK) is not used by anyone. It originally defined properties for an intent that are instead defined in AppDirectory records relating to individual apps these days, while the displayNames have since been deprecated.

/src/intents/standard intents.json contains a list of Standardized intent names, but appears to be missing a few recent additions. It is useful as a source-of-truth for Standardized intent names and could be used to generate datastructures for specific languages in future, but otherwise has no use case in the Standard. Its format could be simplified.

This PR:
- Removes Intent.yaml (as its anachronistic)
- Simplifies the format of standard intents.json and adds the missing entries


kriswest:
@rikoe as the person most likely to remember the discussion about/the creation of _Intents.yaml_ I wanted to flag this to you. Personally, I think its anachronistic and could go...

bingenito (approved):

kriswest:
@hughtroeger could we get your blessing on this?

hughtroeger (approved):

PR #1325, opened 2024-08-29
update google analytics tracking id

TheJuanAndOnly99:
FINOS marketing would like to update the Google Analytics tracking ID to use gtag.


kriswest (commented):
@TheJuanAndOnly99 have you been able to test that this is working via a preview?

TheJuanAndOnly99:
@kriswest yes confirmed. Thank you!

PR #1345, opened 2024-09-10
Rewrite channel selector to use updated message types

julianna-ciq:


julianna-ciq:
@robmoffat I've finished cleaning up the aesthetics, so this PR is ready for you to review and merge.

packages/addon/src/channel_selector.ts
```diff
@@ -35,40 +40,95 @@ function fillChannels(data: ChannelDetails[], selected: string | null, messageCl
...
+  const myPort = mc.port1;
+  myPort.start();
+  myPort.onmessage = ({data}) => {
+    console.debug("Received message: ", data);
+    switch(data.type){
+      case "iframeHandshake": {
+        collapse();
+        break;
+      }
+      case "iframeChannels": {
```
kriswest, line 55: Needs updating to the `fdc3UserInterface*` messsages after we changed the name(s)

```suggestion
      case "fdc3UserInterfaceChannels": {
```

packages/addon/src/channel_selector.ts
```diff
@@ -35,40 +40,95 @@ function fillChannels(data: ChannelDetails[], selected: string | null, messageCl
...
+        logo.removeEventListener("click", expand);
+        const {userChannels, selected} = data.payload as IframeChannelsPayload;
+        fillChannels(userChannels, selected, (channelStr) => {
+          myPort.postMessage({
+            type: "iframeChannelSelected",
```
kriswest, line 60: Needs update 
```suggestion
            type: "fdc3UserInterfaceSelected",
```

packages/addon/src/channel_selector.ts
```diff
@@ -35,40 +40,95 @@ function fillChannels(data: ChannelDetails[], selected: string | null, messageCl
...
-      myPort.postMessage({ type: "SelectorMessageChoice", channelId: id })
-    })
-
-    // ask the parent container to increase the window size
-    changeSize(true)
-  })
+  };
+
+  parent.postMessage({
+    type: "iframeHello",
```
kriswest, line 76:
```suggestion
    type: "fdc3UserInterfaceHello",
```

packages/addon/src/channel_selector.ts
```diff
@@ -35,40 +40,95 @@ function fillChannels(data: ChannelDetails[], selected: string | null, messageCl
...
+        position: "fixed"
+
+      }
+    }
+  }, "*", [mc.port2]);
+
+  const expand = () => {
+    document.body.setAttribute("data-expanded", "true");
+    myPort.postMessage({
+      type: "iframeRestyle",
```
kriswest, line 94:
```suggestion
      type: "fdc3UserInterfaceRestyle",
```

packages/addon/src/channel_selector.ts
```diff
@@ -35,40 +40,95 @@ function fillChannels(data: ChannelDetails[], selected: string | null, messageCl
...
+          "z-index": "1000",
+          position: "fixed"
+        }
+      }
+    });
+  }
 
+  const collapse = () => {
+    myPort.postMessage({
+      type: "iframeRestyle",
```
kriswest, line 111:
```suggestion
      type: "fdc3UserInterfaceRestyle",
```

packages/addon/src/channel_selector.ts
```diff
@@ -35,40 +40,95 @@ function fillChannels(data: ChannelDetails[], selected: string | null, messageCl
...
+  myPort.onmessage = ({data}) => {
+    console.debug("Received message: ", data);
+    switch(data.type){
+      case "iframeHandshake": {
+        collapse();
+        break;
+      }
+      case "iframeChannels": {
+        logo.removeEventListener("click", expand);
+        const {userChannels, selected} = data.payload as IframeChannelsPayload;
```
kriswest, line 57: IframeChannelsPayload needs updating to the new type name. 

```suggestion
        const {userChannels, selected} = data.payload as Fdc3UserInterfaceChannelsPayload;
```
However, you are also using one of the generated components of the top level type. The naming of these is unstable and they should not be used directly. Instead, you should type the `data` element as `Fdc3UserInterfaceChannels` instead and rework accordingly

packages/addon/src/channel_selector.ts
```diff
@@ -35,40 +40,95 @@ function fillChannels(data: ChannelDetails[], selected: string | null, messageCl
...
+        fillChannels(userChannels, selected, (channelStr) => {
+          myPort.postMessage({
```
kriswest, line 59: It might be better to initialise the message as an object and use the Fdc3UserInterfaceChannelSelected type to check it, then post it.

packages/addon/src/channel_selector.ts
```diff
@@ -35,40 +40,95 @@ function fillChannels(data: ChannelDetails[], selected: string | null, messageCl
...
+          zIndex: "1000",
+          "z-index": "1000",
+          position: "fixed"
+        }
+      }
+    });
+  }
 
+  const collapse = () => {
+    myPort.postMessage({
```
kriswest, line 110: Same comment as above - I would set up the message as an object and apply typing, then post it.

packages/addon/src/channel_selector.ts
```diff
@@ -35,40 +40,95 @@ function fillChannels(data: ChannelDetails[], selected: string | null, messageCl
...
+      }
+    });
+
+    // If you immediately change to the logo, before the iframe has a chance to finish restyling,
+    // you see a flicker of a giant, colored logo.
+    // Here, we wait a negligible amount of time, and hope that the restyling has finished. This avoids the flicker.
+    // It's not a *good* idea, it's just the best available, since we don't know when the restyle finishes.
+    setTimeout(() => {
+      document.body.setAttribute("data-expanded", "false");
+    }, 15);
```
kriswest, line 131: Doh, do we need to add a response message to the restyle message so that the DA can indicate that it has applied the change? That's likely to add some lag however, so we might be just exchanging one artefact for another...

This is exactly the sort of issue that we need to explore and provide advice on handling so thank you for identifying it. We'll need to think though reliable solutions / designs that avoid this...

packages/addon/src/intent_resolver.ts
```diff
@@ -1,23 +1,27 @@
 import { Icon } from "@kite9/fdc3";
 import { AppIntent } from "@kite9/fdc3";
-import { ResolverIntents, ResolverMessageChoice, SingleAppIntent } from "@kite9/fdc3-common";
+import { IframeResolveActionPayload, IframeResolvePayload } from "@kite9/fdc3-common";
```
kriswest, line 3: Don't import the subtypes as their naming may be unstable. Please try and use the top-level types instead (and add comments about this).
Also needs updating to the FDC3UserInterface messages

packages/addon/src/intent_resolver.ts
```diff
@@ -58,22 +66,22 @@ function createIcon(icons: Icon[] | undefined): HTMLElement {
   return img
 }
 
-const fillList = (ai: AppIntent[], intent: string, callback: (s: SingleAppIntent) => void) => {
+const fillList = (ai: AppIntent[], intent: string, callback: (s: IframeResolveActionPayload) => void) => {
```
kriswest, line 69: Update to FDC3UserInterface messages and switch to using the top-level type rather than the payload type

packages/addon/src/intent_resolver.ts
```diff
@@ -132,34 +155,61 @@ const fillList = (ai: AppIntent[], intent: string, callback: (s: SingleAppIntent
...
-      const details = e.data as ResolverIntents
-      // console.log(JSON.stringify("INTENT DETAILS: " + JSON.stringify(details)))
-
-      setup(details, callback)
+  const myPort = mc.port1;
+  myPort.start();
+  myPort.onmessage = ({data}) => {
+    console.debug("Received message: ", data);
+    switch(data.type){
+      case "iframeHandshake": {
```
kriswest, line 166:
```suggestion
      case Fdc3UserInterfaceHandshake": {
```

packages/addon/src/intent_resolver.ts
```diff
@@ -132,34 +155,61 @@ const fillList = (ai: AppIntent[], intent: string, callback: (s: SingleAppIntent
...
+        break;
+      }
+      case "iframeResolve": {
```
kriswest, line 169:
```suggestion
      case "Fdc3UserInterfaceResolve": {
```

packages/addon/src/intent_resolver.ts
```diff
@@ -132,34 +155,61 @@ const fillList = (ai: AppIntent[], intent: string, callback: (s: SingleAppIntent
...
+        myPort.postMessage({
+          type: "iframeRestyle",
```
kriswest, line 171:
```suggestion
          type: "fdc3UserInterfaceRestyle",
```

packages/addon/src/intent_resolver.ts
```diff
@@ -132,34 +155,61 @@ const fillList = (ai: AppIntent[], intent: string, callback: (s: SingleAppIntent
...
+  const myPort = mc.port1;
+  myPort.start();
+  myPort.onmessage = ({data}) => {
+    console.debug("Received message: ", data);
+    switch(data.type){
+      case "iframeHandshake": {
+        break;
+      }
+      case "iframeResolve": {
+        myPort.postMessage({
```
kriswest, line 170: Init and type the message as an object, then post

packages/addon/src/intent_resolver.ts
```diff
@@ -132,34 +155,61 @@ const fillList = (ai: AppIntent[], intent: string, callback: (s: SingleAppIntent
...
+        });
+
+        setup(data.payload, (s) => {
+          document.querySelector("dialog")?.close();
+          myPort.postMessage({
+            type: "iframeResolveAction",
+            payload: s
+          });
+
+          myPort.postMessage({
```
kriswest, line 190: Init and type message then post

packages/addon/src/intent_resolver.ts
```diff
@@ -132,34 +155,61 @@ const fillList = (ai: AppIntent[], intent: string, callback: (s: SingleAppIntent
...
+            type: "iframeRestyle",
```
kriswest, line 191:
```suggestion
            type: "fdc3UserInterfaceRestyle",
```

packages/addon/src/intent_resolver.ts
```diff
@@ -132,34 +155,61 @@ const fillList = (ai: AppIntent[], intent: string, callback: (s: SingleAppIntent
...
     }
-  })
-
-  document.getElementById("cancel")!!.addEventListener("click", () => {
-    callback(null);
-  })
-
+  };
+
+  parent.postMessage({
```
kriswest, line 205: Init and type message then post

packages/addon/src/intent_resolver.ts
```diff
@@ -132,34 +155,61 @@ const fillList = (ai: AppIntent[], intent: string, callback: (s: SingleAppIntent
...
+    type: "iframeHello",
```
kriswest, line 206:
```suggestion
    type: "fdc3UserInterfaceHello",
```

kriswest (changes requested):
Thanks for having a go at this @julianna-ciq. Needs a few updates for new message type names. Could also ensure typing is applied to messages sent and that you only ever reference the top-level types for the messages, rather than using the payload types directly (as their naming may be unstable.

Thanks for identifying the styling race - what do you think about ways of avoiding it without using a `setTimeout`

packages/addon/src/channel_selector.ts
robmoffat: @kriswest those changes aren't merged into `fdc3-for-web-impl` so @julianna-ciq's code is correct as it stands.

kriswest:
Nah, that just means there's stuff the other end to change too :-p

I wasn't saying she got it wrong, rather just highlighting the places that it will need to change as we changed the message names.

PR #1360, opened 2024-09-19
Add missing id/name fields to context schema docs

kriswest:
supersedes #1233

Adds missing id and name fields from the context base schema to context schemas that use them (but don't define sub-properties) so that they show up in documentation correctly, in the following types:

- Contact
- ContactList
- Country
- InstrumentList
- OrderList
- Organization
- Portfolio
- Position
- TradeList

Also fixes:
- Rendering of internal references (e.g. internal $refs to $defs or definitions used in a schema), which was broken in ChatSearchCriteria (which defines a SearchCriteria element).
- Bulleting of polymorphic array items (oneOf, anyOf etc.)
- A number of broken links that have slipped through in other PRs which docusaurus reports on a manual build


hughtroeger (approved):
Looks good! All the names and ids are there and the updated links navigate as expected.

PR #1361, opened 2024-09-19
Set explicit background colours for website

kriswest:
The website doesn't currently explicitly set background colors - rather it assume the browser's default background is white (which is a user preference in some browsers, while many electron-based products use a dark background by default - leaving the page impossible to read).

This PR sets an explicit background colour for each mode (dark and light)

### DIFF
--- a/website/src/css/customTheme.css
+++ b/website/src/css/customTheme.css
@@ -1,4 +1,12 @@
 
+html[data-theme='light'] {
+    background-color: white;
+}
+
+html[data-theme='dark'] {
+    background-color: black;
+}
+
 @font-face {
     src: url("/fonts/Electrolize-Regular.ttf");
     font-family: 'Electrolize'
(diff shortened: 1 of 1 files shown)


bingenito:
Any thoughts on a bit lighter like #1b1b1d?  You can see with dark theme on docusaurus.org for an example.

bingenito (approved):

website/src/css/customTheme.css
```diff
@@ -1,4 +1,12 @@
 
+html[data-theme='light'] {
+    background-color: white;
+}
+
+html[data-theme='dark'] {
+    background-color: black;
```
kriswest, line 7:
```suggestion
    background-color: #1b1b1b;
```

website/src/css/customTheme.css
```diff
@@ -1,4 +1,12 @@
 
+html[data-theme='light'] {
+    background-color: white;
```
kriswest, line 3:
```suggestion
    background-color: #FFFFFF;
```

kriswest:
@bingenito adjusted to #1b1b1b, does seem a little less stark

PR #1191, opened 2024-04-15
FDC3 for Web specification

kriswest:
resolves #896 
resolves #1385 
resolves #1402 

This PR adds specifications and documentation related to enabling FDC3 for the web. This was based off of the committee's [working doc](https://tick42-my.sharepoint.com/:w:/g/personal/finsemble_datastore_interop_io/EZ0dfTCdRlJCnIF3C_1Oit0BF3fsXyvlMbisXp722DC9Kg?e=JNXSRi&isSPOFile=1&clickparams=eyJBcHBOYW1lIjoiVGVhbXMtRGVza3RvcCIsIkFwcFZlcnNpb24iOiI0OS8yNDAyMDIwNTUxNCIsIkhhc0ZlZGVyYXRlZFVzZXIiOmZhbHNlfQ%3D%3D) and an initial PR ([#1167](https://github.com/finos/FDC3/pull/1167)) with the draft by @thorsent.

This assumes that "@finos/fdc3" will implement getAgent() and communication mechanisms to work with a Desktop Agent running in a different browser window.

The protocols were codified as "Web Connection Protocol (WCP)" for negotiating the handshake between the library and desktop agents, and "Browser Communication Protocol (BCP)" which encompasses the complete "wire protocol". The former is entirely specific to working in a Web Browser, however the latter might be reused in other implementations and hence may change name (e.g. 'The FDC3 Wire Protocol' or similar).

This documentation followed three streams:

- Modifications for users of FDC3 - primarily related to using getAgent() for connectivity instead of relying on the global fdc object
- Instructions for desktop agent implementers - provides a specification for "browser-resident desktop agents" as well as "preload desktop agents", along with typescript definitions.
- A specification for getAgent() - to be used by implementers of the "@finos/fdc3" library, and for reference by desktop agent implementers.

Some changes and additions were made from the original working document:

- The optional initialization fields for channel selector and intent resolver were eliminated. The DA knows whether it can provide these UI elements and so therefore it should be the DA that decides how to accomplish this. However, there are cases where a DA cannot reasonably provide these UI elements in a browser, therefore this spec mandates that "@finos/fdc3" provide built-in UI capabilities that a DA can optionally enable them.
- This spec also covers window disconnects by suggesting use of a well known polling mechanism.
- A new folder called "specs" was created to house many of the specifications involved. Previously, there was no true distinction between implementation specs and user specs in the FDC3 overview.

Deep links to docs for review:
[...]

PR #1379, opened 2024-09-27
Context Clearing

### LINKED ISSUE #1197: Context Clearing
### Use Case:

Explicit / first-class support for context clearing via the API.

For example: 
> Often contexts are used to set filters, which may later be removed by a user. E.g. filter a data-grid by a particular instrument, trade, timerange etc

This is a fairly generic use-case that could be applied to many contexts.

### Workflow Description

- Application connects to a channel with multiple contexts and filters some data based on one or more of those contexts
- User removes one of the context filters
- Expectation is that only that context is removed but others remain set on the channel

### Workflow Examples

- User opens an application that shows a list of `Trade` entries filtered by `Product`
- User further filters by `TimeRange`
- User removes the `TimeRange` filter

### Additional Information

The issue of context clearing has previously been raised and closed at https://github.com/finos/FDC3/issues/301 and appears to have been subsequently discussed several times at the WG meetings.

From what I can tell the recommended solution was to use the new `fdc3.nothing` context or to broadcast a context with a null `id`.

Neither of these solutions feel ergonomic, particularly when dealing with multi-context channels.

##### `fdc3.nothing`

The semantics of this approach appear to be limited to clearing *all* contexts on a channel since there is no way to specify a target sub-type. Therefore it is not suitable for clearing individual contexts.

It also leaves it up to each application to correctly deal with the actual clearing of contexts.

##### null `id` field

This approach does support clearing of individual contexts and is likely well-supported as long as applications have correctly implemented against the type definitions.

However it requires "empty" versions of each context type to be explicitly created in order to maintain the type contract. It's not enough to just set `id` to null as for many context types there are other required fields. E.g. `Position`, both `holding` and `instrument` are required fields.

In some cases `id` is nullable, in others it is not e.g. (`Product`). It might be fine to pass an empty object in that case, but again this is not very clear or ergonomic.

For a one off use, the above is manageable, but for example if building internal libraries and helpers to work with FDC3 this becomes painful to write abstractions around without becoming coupled into individual context implementations.

##### Other

For our internal use-cases we've also explored extending `fdc3.nothing` with a subType - which helps to some extent but raises further complications around actually clearing the value for `getCurrentContext` etc. and is also only going to work for internal applications that are aware of the semantics.

#### Proposed solution (1)

Add an explicit `Channel.clearContext` method for clearing contexts:

```TypeScript
interface Channel {
  ...
  broadcast(context: Context): Promise<void>;  
  getCurrentContext(contextType?: string): Promise<Context|null>;
  clearContext(contextType?: string): Promise<void>;
  addContextListener(contextType: string | null, handler: ContextHandler): Promise<Listener>;
}
```

This could work with similar semantics to `getCurrentContext` and `addContextListener` in that providing a `contextType` targets that specific context, or leaving it undefined would apply to all contexts.

Some consideration would be needed for how to pass the context type of a cleared context to the callback on a `ContextHandler` subscribed to *all* context updates. One option would be adding context type info to `ContextMetadata`.

#### Alternate solution (2)

Add an optional `subType` field to `fdc3.nothing`:

```TypeScript
interface Nothing {
  type: 'fdc3.nothing';
  subType?: string;
}
```

Add documentation on the semantics of this to the effect of providing a subType indicates that an individual context should be cleared.

Likely some kind of API change would also be needed to allow implementations to then deal with this and provide a means to clear current context without triggering a separate broadcast.

kemerava:
Closes #1197 

Adding context cleaning 

@michael-bowen-sc

### DIFF
--- a/docs/api/ref/Channel.md
+++ b/docs/api/ref/Channel.md
@@ -36,6 +36,7 @@ interface Channel {
   broadcast(context: Context): Promise<void>;
   getCurrentContext(contextType?: string): Promise<Context|null>;
   addContextListener(contextType: string | null, handler: ContextHandler): Promise<Listener>;
+  clearContext(contextType?: string): Promise<void>;
   
   //deprecated functions
   /**
@@ -57,6 +58,7 @@ interface IChannel: IIntentResult
     Task Broadcast(IContext context);
     Task<IContext?> GetCurrentContext(string? contextType);
     Task<IListener> AddContextListener<T>(string? contextType, ContextHandler<T> handler) where T : IContext;
+    Task ClearContext(string? contextType);
 }
 ```
 
@@ -430,6 +432,99 @@ catch (Exception ex)
 - [`ChannelError`](Errors#channelerror)
 - [`broadcast`](#broadcast)
 - [`addContextListener`](#addcontextlistener)
+
+### `clearContext`
+
+<Tabs groupId="lang">
+<TabItem value="ts" label="TypeScript/JavaScript">
+
+```ts
+public clearContext(contextType?: string): Promise<void>;
+```
+
+</TabItem>
+<TabItem value="dotnet" label=".NET">
+
+```csharp
+Task ClearContext(string? contextType);
+```
+
+</TabItem>
+</Tabs>
+
+Used to clear the specified context type if provided, otherwise, clear all contexts. Desktop Agent MUST update internal context. Any future calls to [`getCurrentContext`](#getcurrentcontext) and any new joiners on that channel (through [`joinUserChannel`](DesktopAgent#joinUserChannel) or [`addContextListener`](DesktopAgent#addContextListener)) will not receive anything for specified context type or all contexts accordingly. 
+Desktop Agent MUST update all channel listeners subscribed to the  [`fdc3.nothing`](../../context/ref/Nothing.md/#nothing) type. If `contextType` is provided, then `subType` will be specified, otherwise, it is ommitted. 
+
+
+**Examples:**
+
+Without specifying a context type:
+
+<Tabs groupId="lang">
+<TabItem value="ts" label="TypeScript/JavaScript">
+
+```ts
+try {
+    const context = await channel.clearContext();
+} catch (err: ChannelError) {
+    // handle error
+}
+```
+
+</TabItem>
+<TabItem value="dotnet" label=".NET">
+
+```csharp
+try
+{
+    var context = await channel.ClearContext();
+}
+catch (Exception ex)
+{
+    // handle error
+}
+```
+
+</TabItem>
+</Tabs>
+
+Specifying a context type:
+
+<Tabs groupId="lang">
+<TabItem value="ts" label="TypeScript/JavaScript">
+
+```ts
+try {
+    const contact = await channel.clearContext('fdc3.contact');
+} catch (err: ChannelError) {
+    // handler error
+}
+```
+
+</TabItem>
+<TabItem value="dotnet" label=".NET">
+
+```csharp
+try
+{
+    var context = await channel.ClearContext("fdc3.contact");
+}
+catch (Exception ex)
+{
+    // handle error
+}
+```
+
+</TabItem>
+</Tabs>
+
+
+**See also:**
+
+- [`getCurrentContext`](#getcurrentcontext)
+- [`addContextListener`](DesktopAgent#addContextListener)
+- [`joinUserChannel`](DesktopAgent#joinUserChannel)
+- [`fdc3.nothing`](../../context/ref/Nothing.md/#nothing)
 
 ## Deprecated Functions
 
--- a/src/api/Channel.ts
+++ b/src/api/Channel.ts
@@ -74,6 +74,15 @@ export interface Channel {
   addContextListener(contextType: string | null, handler: ContextHandler): Promise<Listener>;
 
   /**
+   * Provides the ability to clear context from the channel. Subsequent connection to this channel or calls to get current context will not return anything (accounting for contextType if specified). 
+   * 
+   * If a `contextType` is provided, only contexts of that type will be cleared. 
+   * 
+   * If no `contextType` is provided, all contexts will be cleared.
+   */
+  clearContext(contextType?: string): Promise<void>;
+
+  /**
    * @deprecated use `addContextListener(null, handler)` instead of `addContextListener(handler)`.
    */
   addContextListener(handler: ContextHandler): Promise<Listener>;
--- a/docs/api/spec.md
+++ b/docs/api/spec.md
@@ -241,7 +241,7 @@ When raising an intent a specific context is provided as input. The type of the 
 
 A context type may also be associated with multiple intents. For example, an `fdc3.instrument` could be associated with `ViewChart`, `ViewNews`, `ViewAnalysis` or other intents.
 
-To raise an intent without a context, use the [`fdc3.nothing`](../context/ref/Nothing) context type. This type exists so that applications can explicitly declare that they support raising an intent without a context (when registering an `IntentHandler` or in an App Directory).
+To raise an intent without a context, use the [`fdc3.nothing`](../context/ref/Nothing) context type. This type exists so that applications can explicitly declare that they support raising an intent without a context (when registering an `IntentHandler` or in an App Directory). This type is also used when the context is cleared for the channel. If the optional context type is provided when performing [`clearContext`](ref/Channel.md#clearcontext), that type will be recorded in the field `subType` of [`fdc3.nothing`](../context/ref/Nothing) context type.
 
 As an alternative to raising a specific intent, you may also raise an unspecified intent with a known context allowing the Desktop Agent or the user (if the intent is ambiguous) to select the appropriate intent and then to raise it with the specified context for resolution.
 
@@ -585,7 +585,7 @@ Apps can join *User channels*.  An app can only be joined to one User channel at
 
 When an app is joined to a User channel, calls to [`fdc3.broadcast`](ref/DesktopAgent#broadcast) will be routed to that channel and listeners added through [`fdc3.addContextListener`](ref/DesktopAgent#addcontextlistener) will receive context broadcasts from other apps also joined to that channel. If an app is not joined to a User channel [`fdc3.broadcast`](ref/DesktopAgent#broadcast) will be a no-op and handler functions added with  [`fdc3.addContextListener`](ref/DesktopAgent#addcontextlistener) will not receive any broadcasts. However, apps can still choose to listen and broadcast to specific channels (both User and App channels) via the methods on the [`Channel`](ref/Channel) class.
 
-When an app joins a User channel, or adds a context listener when already joined to a channel, it will automatically receive the current context for that channel.
+When an app joins a User channel, or adds a context listener when already joined to a channel, it will automatically receive the current context for that channel, unless the context was cleared through [`clearContext`](ref/Channel.md#clearcontext).
 
 It is possible that a call to join a User channel could be rejected.  If for example, the desktop agent wanted to implement controls around what data apps can access.
 
@@ -829,6 +829,9 @@ The `PrivateChannel` type also supports synchronization of data transmitted over
 The [Context specification](../context/spec#assumptions) recommends that complex context objects are defined using simpler context types for particular fields. For example, a `Position` is composed of an `Instrument` and a holding amount. This leads to situations where an application may be able to receive or respond to context objects that are embedded in a more complex type, but not the more complex type itself. For example, a pricing chart might respond to an `Instrument` but doesn't know how to handle a `Position`.
 
 To facilitate context linking in such situations it is recommended that applications `broadcast` each context type that other apps (listening on a User Channel or App Channel) may wish to process, starting with the simpler types, followed by the complex type. Doing so allows applications to filter the context types they receive by adding listeners for specific context types - but requires that the application broadcasting context make multiple broadcast calls in quick succession when sharing its context.
+
+### Context clearning on channels
+Channel interface provides the ability to [`clearContext`](ref/Channel.md#clearcontext) on the channel, either for the specific context type, if provided, or for all contexts on that channel. If provided, the specified type will be recorded in the `subType` field of the [`fdc3.nothing`](../context/ref/Nothing) context type. Once cleared, any subsequent new joiners on the channel, new context listeners on the context or calls to [`getCurrentContext`](ref/Channel.md#getcurrentcontext) will not return anything to the caller. 
 
 ### Originating App Metadata
 
--- a/schemas/context/nothing.schema.json
+++ b/schemas/context/nothing.schema.json
@@ -10,6 +10,10 @@
       "properties": {
         "type": {
           "const": "fdc3.nothing"
+        },
+        "subtype": {
+          "type": "string",
+          "description": "A subtype of the context type. This can be used to provide a specific context which should be cleared, if ommitted, all contexts will be cleared"
         }
       }
     },
@@ -17,7 +21,8 @@
   ],
   "examples": [
     {
-      "type": "fdc3.nothing"
+      "type": "fdc3.nothing",
+      "subtype": "fdc3.timeRange"
     }
   ]
 }


docs/api/ref/Channel.md
```diff
@@ -36,6 +36,7 @@ interface Channel {
   broadcast(context: Context): Promise<void>;
   getCurrentContext(contextType?: string): Promise<Context|null>;
   addContextListener(contextType: string | null, handler: ContextHandler): Promise<Listener>;
+  clearContext(contextType?: string): Promise<void>;
```
kriswest, line 39: I'm wondering if we should require an explicit null argument (`contextType: string | null`) to clear all types - which would match `addContectListener` or stick with this (matching `getCurrentContext`). Either works of course.

docs/api/ref/Channel.md
```diff
@@ -431,6 +433,99 @@ catch (Exception ex)
...
+<TabItem value="dotnet" label=".NET">
+
+```csharp
+Task ClearContext(string? contextType);
+```
+
+</TabItem>
+</Tabs>
+
+Used to clear the specified context type if provided, otherwise, clear all contexts. Desktop Agent MUST update internal context. Any future calls to [`getCurrentContext`](#getcurrentcontext) and any new joiners on that channel (through [`joinUserChannel`](DesktopAgent#joinUserChannel) or [`addContextListener`](DesktopAgent#addContextListener)) will not receive anything for specified context type or all contexts accordingly.
```
kriswest, line 455:
```suggestion
Used to clear the specified context type if provided, otherwise, clear all context types present in the channel. The Desktop Agent MUST update its internal representation of the context in the channel and ensure that subsequent calls to [`getCurrentContext`](#getcurrentcontext) and any new joiners to that channel (through [`joinUserChannel`](DesktopAgent#joinUserChannel) or [`addContextListener`](DesktopAgent#addContextListener)) will not receive anything for either specified context types or the most recent context until new context has been broadcast to the channel. 
```

docs/api/ref/Channel.md
```diff
@@ -431,6 +433,99 @@ catch (Exception ex)
...
+
+```csharp
+Task ClearContext(string? contextType);
+```
+
+</TabItem>
+</Tabs>
+
+Used to clear the specified context type if provided, otherwise, clear all contexts. Desktop Agent MUST update internal context. Any future calls to [`getCurrentContext`](#getcurrentcontext) and any new joiners on that channel (through [`joinUserChannel`](DesktopAgent#joinUserChannel) or [`addContextListener`](DesktopAgent#addContextListener)) will not receive anything for specified context type or all contexts accordingly. 
+Desktop Agent MUST update all channel listeners subscribed to the  [`fdc3.nothing`](../../context/ref/Nothing.md/#nothing) type. If `contextType` is provided, then `subType` will be specified, otherwise, it is ommitted.
```
kriswest, line 456:
```suggestion
Desktop Agents MUST also immediately broadcast the  [`fdc3.nothing`](../../context/ref/Nothing.md/#nothing) type, which applications may listen to to be notified of the cleared context. If a `contextType` was provided, then the `subType` field will be be set to that type, otherwise, it is omitted. 
```

src/api/Channel.ts
```diff
@@ -73,6 +73,15 @@ export interface Channel {
    */
   addContextListener(contextType: string | null, handler: ContextHandler): Promise<Listener>;
 
+  /**
+   * Provides the ability to clear context from the channel. Subsequent connection to this channel or calls to get current context will not return anything (accounting for contextType if specified).
```
kriswest, line 77:
```suggestion
   * Clears context from the channel, and broadcasts an `fdc3.nothing` context to notify existing listeners that the context was cleared. Listeners added to the channel and calls to [`getCurrentContext`](#getcurrentcontext) will not receive any existing context until new context is broadcast to the channel. 
```

src/api/Channel.ts
```diff
@@ -73,6 +73,15 @@ export interface Channel {
    */
   addContextListener(contextType: string | null, handler: ContextHandler): Promise<Listener>;
 
+  /**
+   * Provides the ability to clear context from the channel. Subsequent connection to this channel or calls to get current context will not return anything (accounting for contextType if specified). 
+   * 
+   * If a `contextType` is provided, only contexts of that type will be cleared.
```
kriswest, line 79:
```suggestion
   * If a `contextType` is provided, only contexts of that type will be cleared and the `contextType` of the `fdc3.nothing` context will be set to that type name. 
```

src/api/Channel.ts
```diff
@@ -73,6 +73,15 @@ export interface Channel {
    */
   addContextListener(contextType: string | null, handler: ContextHandler): Promise<Listener>;
 
+  /**
+   * Provides the ability to clear context from the channel. Subsequent connection to this channel or calls to get current context will not return anything (accounting for contextType if specified). 
+   * 
+   * If a `contextType` is provided, only contexts of that type will be cleared. 
+   * 
+   * If no `contextType` is provided, all contexts will be cleared.
```
kriswest, line 81:
```suggestion
   * If no `contextType` is provided, all contexts will be cleared and the `contextType` of the `fdc3.nothing` context will be omitted.
```

kriswest (commented):
Many thanks for taking this on @kemerava.

Could you add an entry to the CHANGLOG.md file please.

I think we should keep this open until after fdc3 for the web (#1191) is merged, when we will need to add schemas for a request and response message. The same *may* be needed in the bridging (the `fdc3.nothing` ewill propagate through normal broadcast messages, but I think this will need a specific message to indicate it happened through the clearContextAPI call).

docs/api/ref/Channel.md
kemerava: To me it makes sense to keep it closer to getCurrentContext, because they seem most related to me, but happy to hear any thoughts/suggestions.

docs/api/ref/Channel.md
kemerava: Question, Kris, should I change "Desktop Agent" to "Desktop Agents" everywhere in my descriptions?

docs/api/ref/Channel.md
kriswest: There is a historical reason for addCOntextListener having an explicit null argument, the type argument was originally optional so the `handler` argument could be either first or second (when a type was specified). This was good in practice so we moved to an explicit `null` argument and deprecated the single argument override:
https://fdc3.finos.org/docs/2.0/api/ref/DesktopAgent#addcontextlistener-deprecated

We don't have that problem here, so I agree its fine as is (matching `getCurrentContext`).

kriswest (commented):
LGTM - although we could do with an example of listening for cleared context somewhere.

Holding approval until after fdc3-for-web is merged as we'll then need to add schemas to support it there.  Bridging schemas are also needed, those should be added before this is merged.

Many thanks for the contribution @kemerava, much appreciated.

kemerava:
@kriswest could you please take a look at the conformance tests that I added? Thanks

toolbox/fdc3-conformance/App-Channel-Tests.md
```diff
@@ -52,3 +52,6 @@
 - `ACContextHistoryTyped`: Perform above test.
 - `ACContextHistoryMultiple`: **B** Broadcasts multiple history items of both types.  Ensure that only the last version of each type is received by **A**.
 - `ACContextHistoryLast`: In step 5. **A** retrieves the _untyped_ current context of the channel via `const currentContext = await testChannel.getCurrentContext()`. Ensure that A receives only the very last broadcast context item _of any type_.
+- `ACClearHistorySpecificContext`: Perform the above test, except after step 3 clear context with `await testChannel.clearContext('fdc3.instrument')`. Ensure that in step 5 nothing is received for `const instrument = await testChannel.getCurrentContext('fdc3.instrument')` and context is received successfully for `const instrument = await testChannel.getCurrentContext('fdc3.contact')`
```
kriswest, line 55:
```suggestion
- `ACClearHistorySpecificContext`: Perform the above test, except after step 3 clear context with `await testChannel.clearContext('fdc3.instrument')`. Ensure that in step 5 nothing is received for `const instrument = await testChannel.getCurrentContext('fdc3.instrument')` and context is received successfully for `const contact = await testChannel.getCurrentContext('fdc3.contact')`
```

toolbox/fdc3-conformance/App-Channel-Tests.md
```diff
@@ -52,3 +52,6 @@
...
+- `ACClearHistoryAllContexts`: Perform the above test, except after step 3 clear context with `await testChannel.clearContext('fdc3.nothing')`. Ensure that in step 5 nothing is received for `const instrument = await testChannel.getCurrentContext('fdc3.instrument')` and for `const instrument = await testChannel.getCurrentContext('fdc3.contact')`
```
kriswest, line 56:
```suggestion
- `ACClearHistoryAllContexts`: Perform the above test, except after step 3 clear context with `await testChannel.clearContext('fdc3.nothing')`. Ensure that in step 5 nothing is received for `const instrument = await testChannel.getCurrentContext('fdc3.instrument')` and for `const contact = await testChannel.getCurrentContext('fdc3.contact')`
```

toolbox/fdc3-conformance/App-Channel-Tests.md
```diff
@@ -52,3 +52,6 @@
...
+- `ACClearHistoryAllContextsSubscribedToNothing`: Perform the above test, except after step 3 clear context with `await testChannel.clearContext('fdc3.nothing')` and in step 4 call `await testChannel.addContextListener("fdc3.nothing", handler)` instead. Ensure that after clearing context, the handler is called.
```
kriswest, line 57: A few thoughts on this case:
- I think you want to clear fdc3.instrument or all types (null) rather than fdc3.nothing
- As written, the nothing handler is added after calling clear, in that case it won't receive anything on an app channel. The handler needs to be in place before the clear.
- I think this needs to be 2 cases, one for clearing all types and another for a specific type that confirms the other is not affected and that the fdc3.nothing quotes the correct type.
- Finally, I think we need at least one case that checks the behavior of `user` channele-  you'll work with the channel as an app channel to clear it (either retrrieving it by name, via `getUserChannels`, or `getCUrrentChannel`) but we should also confirm that it gets the fdc3.nothing broadcast and that it doesn't retain the cleared context.

kriswest (changes requested):
Thanks for adding the conformance test cases! I think we'llneed a couple f additional cases that look the effect on user channels.

kriswest:
@kemerava, @robmoffat and I were talking about this recently and wondering if we should rethink the fdc3.nothing use and instead go with events... We could bring `addEventListener` to Channel (currently only in `PrivateChannel` and allow aps to listen for context clearing events (on the DA for the current user channel on the channel object for specific channels). I think the fdc3.nothing use is probably the weakest or most awkward part of the proposal (particularly if someone broadcasts fdc3.nothing from an app).

WDYT?

kemerava:
> @kemerava, @robmoffat and I were talking about this recently and wondering if we should rethink the fdc3.nothing use and instead go with events... We could bring `addEventListener` to Channel (currently only in `PrivateChannel` and allow aps to listen for context clearing events (on the DA for the current user channel on the channel object for specific channels). I think the fdc3.nothing use is probably the weakest or most awkward part of the proposal (particularly if someone broadcasts fdc3.nothing from an app).
> 
> WDYT?

@kriswest updated based on what I understood the logic would be. Please let me know if I need to change something else or if I misunderstood your suggestion

packages/fdc3-context/schemas/context/nothing.schema.json
```diff
@@ -10,14 +10,19 @@
       "properties": {
         "type": {
           "const": "fdc3.nothing"
+        },
+        "contextType": {
+          "type": "Context Type",
```
Roaders, line 15: I don't think that we still need this change now that we're not using the `fdc3.nothing` context?

packages/fdc3-standard/src/api/Events.ts
```diff
@@ -46,6 +46,16 @@ export interface FDC3ChannelChangedEvent extends FDC3Event {
   };
 }
 
+/**
+ * Type defining the format of event `contextCleared` objects
+ */
+export interface FDC3ContextClearedEvent extends FDC3Event {
+  readonly type: 'contextCleared';
+  readonly details: {
+    type: string | null;
```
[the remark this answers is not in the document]
Roaders: this is pretty minor but might be worth making consistent unless I am missing something.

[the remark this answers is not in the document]
kemerava: Thanks, changed it!

packages/fdc3-context/schemas/context/nothing.schema.json
kemerava: Removed, thanks!

CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 * Added details of and procedures for resolving fully-qualified appIds and unqualified appIds in the API and Bridging Parts of the Standard. ([#1523](https://github.com/finos/FDC3/pull/1523))
 
+* Added `clearContext` function to the `Channel` API, to be able to clear specific or all context types from the channel. ([#1379](https://github.com/finos/FDC3/pull/1379))
```
kriswest:
```suggestion
* Added `clearContext` function and associated `contextClearedEvent` to the `Channel` API, to be able to clear specific or all context types from the channel. ([#1379](https://github.com/finos/FDC3/pull/1379))
```

packages/fdc3-agent-proxy/src/channels/DefaultChannel.ts
```diff
@@ -98,4 +102,30 @@ export class DefaultChannel implements Channel {
...
+    };
+    await this.messaging.exchange<ClearContextResponse>(request, 'clearContextResponse', this.messageExchangeTimeout);
+  }
+
+  async addEventListener(type: string | null, handler: EventHandler): Promise<Listener> {
+    let listener: RegisterableListener;
+    switch (type) {
+      case 'contextCleared':
+        listener = new EventListener(this.messaging, 'contextCleared', handler);
+        break;
```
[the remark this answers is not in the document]
kriswest: Copilot has a point there. It looks like this is also wrong where we listener for the channelChanged event:

https://github.com/finos/FDC3/blob/d4a261a043d994670130ecbe177ca9aabea4e87b/packages/fdc3-agent-proxy/src/DesktopAgentProxy.ts#L77-L85

Co-pilot's suggestion to fix isn't right however. Rather EventListener will probbably ned to be adjusted to be able to take multiple types:
https://github.com/finos/FDC3/blob/d4a261a043d994670130ecbe177ca9aabea4e87b/packages/fdc3-agent-proxy/src/listeners/EventListener.ts#L12

As we need to know the specific messages it should respond to (from a set that is wider than the scope of the listener).

packages/fdc3-standard/src/api/Channel.ts
```diff
@@ -73,6 +74,38 @@ export interface Channel {
    */
   addContextListener(contextType: string | null, handler: ContextHandler): Promise<Listener>;
 
+  /**
+   * Clears context from the channel, and broadcasts an `fdc3.nothing` context to notify existing listeners that the context was cleared. Listeners added to the channel and calls to [`getCurrentContext`](#getcurrentcontext) will not receive any existing context until new context is broadcast to the channel. 
+   * 
+   * If a `contextType` is provided, only contexts of that type will be cleared and the `contextType` of the `fdc3.nothing` context will be set to that type name.
```
robmoffat: I think this needs updating now that we have the event listener.

packages/fdc3-agent-proxy/src/channels/DefaultChannel.ts
```diff
@@ -98,4 +102,30 @@ export class DefaultChannel implements Channel {
     await listener.register();
     return listener;
   }
+
```
robmoffat, line 105: we're going to need some cucumber tests for this otherwise you'll see the coverage dip!

packages/fdc3-agent-proxy/src/channels/DefaultChannel.ts
```diff
@@ -98,4 +102,30 @@ export class DefaultChannel implements Channel {
...
+  async clearContext(contextType?: string): Promise<void> {
+    // first, ensure channel state is up-to-date
+    const request: ClearContextRequest = {
+      meta: this.messaging.createMeta(),
+      payload: {
```
robmoffat, line 110: Also, there will need to be support for this event added to the `BroadcastHandler` in the `fdc3-web-impl`, plus tests of the same there.

website/docs/api/conformance/App-Channel-Tests.md
```diff
@@ -60,3 +60,19 @@ hide_title: true
...
+| A   | 2. Add Context Listener | Add an _typed_ context listener for `fdc3.instrument`, using: <br />`await testChannel.addContextListener("fdc3.instrument",handler)`|
+| B   | 3. Retrieve `Channel`   | Retrieve a `Channel` object representing the same 'App' channel A did (`test-channel`)|
+| B   | 4. Broadcast            | B broadcasts both an `fdc3.instrument` context and an `fdc3.contact` context, using: <br /> `testChannel.broadcast(<fdc3.instrument>)` <br /> `testChannel.broadcast(<fdc3.contact>)`|
+| A   | 5. Receive Context      | An fdc3.instrument context is received by the handler added in step 2.<br />Ensure that the fdc3.instrument received by A is identical to that sent by B<br />Ensure that the fdc3.contact context is NOT received.                                                                   |
+| A   | 6. Add Event Listener    | Add an event listener for `contextCleared` event, using: <br />`await testChannel.addEventListener("contextCleared",handler)`  |
+| B   | 7. Clear Context            | B clears context using `testChannel.clearContext()`|
```
hughtroeger, line 75: Is it necessary to add a final step to verify that the contextCleared event listener on App A was fired? Or is that implied?

website/docs/api/ref/Channel.md
```diff
@@ -588,6 +644,99 @@ if result.Err != null {
...
+<TabItem value="dotnet" label=".NET">
+
+```csharp
+Task ClearContext(string? contextType);
+```
+
+</TabItem>
+</Tabs>
+
+Used to clear the specified context type if provided, otherwise, clear all context types present in the channel. The Desktop Agent MUST update its internal representation of the context in the channel and ensure that subsequent calls to [`getCurrentContext`](#getcurrentcontext) and any new joiners to that channel (through [`joinUserChannel`](DesktopAgent#joinUserChannel) or [`addContextListener`](DesktopAgent#addContextListener)) will not receive anything for either specified context type or the most recent context until new context has been broadcast to the channel.
```
hughtroeger:
```suggestion
Used to clear the specified context type if provided, otherwise, clear all context types present in the channel. The Desktop Agent MUST update its internal representation of the context in the channel and ensure that subsequent calls to [`getCurrentContext`](#getcurrentcontext) and any new joiners to that channel (through [`joinUserChannel`](DesktopAgent#joinUserChannel) or [`addContextListener`](DesktopAgent#addContextListener)) will not receive anything for either the specified context type or the most recent context until new context has been broadcast to the channel. 
```

website/docs/api/ref/Channel.md
```diff
@@ -588,6 +644,99 @@ if result.Err != null {
...
+</TabItem>
+</Tabs>
+
+
+**See also:**
+
+- [`getCurrentContext`](#getcurrentcontext)
+- [`addContextListener`](DesktopAgent#addContextListener)
+- [`joinUserChannel`](DesktopAgent#joinUserChannel)
+- [`fdc3.nothing`](../../context/ref/Nothing.md/#nothing)
```
hughtroeger: Should this reference be dropped now? Maybe replaced with `addEventListener` ?

website/docs/api/spec.md
```diff
@@ -256,7 +256,7 @@ When raising an intent a specific context is provided as input. The type of the
 
 A context type may also be associated with multiple intents. For example, an `fdc3.instrument` could be associated with `ViewChart`, `ViewNews`, `ViewAnalysis` or other intents.
 
-To raise an intent without a context, use the [`fdc3.nothing`](../context/ref/Nothing) context type. This type exists so that applications can explicitly declare that they support raising an intent without a context (when registering an `IntentHandler` or in an App Directory).
+To raise an intent without a context, use the [`fdc3.nothing`](../context/ref/Nothing) context type. This type exists so that applications can explicitly declare that they support raising an intent without a context (when registering an `IntentHandler` or in an App Directory). This type is also used when the context is cleared for the channel. If the optional context type is provided when performing [`clearContext`](ref/Channel.md#clearcontext), that type will be recorded in the field `subType` of [`fdc3.nothing`](../context/ref/Nothing) context type.
```
hughtroeger, line 259: I'm unsure if this is still current. @kriswest indicated that the approach had been changed from using `fdc3.nothing` to the event based workflow. Is this still accurate?

website/docs/api/conformance/App-Channel-Tests.md
kemerava: Added, thanks!

website/docs/api/ref/Channel.md
kemerava: Done

packages/fdc3-standard/src/api/Channel.ts
kemerava: Updated

PR #1386, opened 2024-10-15
Make window.fdc3 optional

### LINKED ISSUE #1385: window.fdc3 should be marked as optional
### Use Case:
I want to test to see if `window.fdc3` is defined

For example: 
In writing and testing a `getAgent` implementation it's important to see if an agent has already been created and assigned to `window.fdc3`. Currently the [typing ](https://github.com/finos/FDC3/blob/fffcb331fccb36c7363f50180c35986c1a487fdb/src/index.ts#L37)specified it as being `DesktopAgent`:

```ts
declare global {
  interface Window {
    fdc3: DesktopAgent;
  }
}
```

This however should be marked as optional so any code accessing that property check to see if it's defined first:

```ts
declare global {
  interface Window {
    fdc3?: DesktopAgent;
  }
}
```

Roaders:
Currently `window.fdc3` is incorrectly marked as always being set. `getAgent` code will need to be able to determine if this has been set already and app code should also be able to determine if it is currently set. At the moment this is difficult due to the incorrect types.

fixes #1385 

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Changed
 
+* `window.fdc3` is now an optional property and may or may not be defined.
 * `Listener.unsubscribe()` was made async (the return type was changed from `void` to `Promise<void>`) for consistency with the rest of the API. ([#1305](https://github.com/finos/FDC3/pull/1305))
 * Added reference materials and supported platforms information for FDC3 in .NET via the [finos/fdc3-dotnet](https://github.com/finos/fdc3-dotnet) project. ([#1108](https://github.com/finos/FDC3/pull/1108))
 * The supported platforms page in the FDC3 documentation was moved into the API section as the information it provides all relates to FDC3 Desktop Agent API implementations. ([#1108](https://github.com/finos/FDC3/pull/1108))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Changed
 
-* `window.fdc3` is now an optional property and may or may not be defined.
+* `window.fdc3` is now an optional property and may or may not be defined. Applications should use `getAgent()` is now the recommended way of retrieving a reference to the FDC3 API. ([#1386](https://github.com/finos/FDC3/pull/1386))
 * `Listener.unsubscribe()` was made async (the return type was changed from `void` to `Promise<void>`) for consistency with the rest of the API. ([#1305](https://github.com/finos/FDC3/pull/1305))
 * Added reference materials and supported platforms information for FDC3 in .NET via the [finos/fdc3-dotnet](https://github.com/finos/fdc3-dotnet) project. ([#1108](https://github.com/finos/FDC3/pull/1108))
 * The supported platforms page in the FDC3 documentation was moved into the API section as the information it provides all relates to FDC3 Desktop Agent API implementations. ([#1108](https://github.com/finos/FDC3/pull/1108))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Changed
 
-* `window.fdc3` is now an optional property and may or may not be defined. Applications should use `getAgent()` is now the recommended way of retrieving a reference to the FDC3 API. ([#1386](https://github.com/finos/FDC3/pull/1386))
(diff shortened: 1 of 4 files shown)


src/context/ContextTypes.ts
Roaders, line 1: I didn't make any changes to this file. I think it was an automated husky script that did this.

kriswest: Don't worry about this one, every PR is making the same changes, something got merged without generation being run. We'll make sure this is sorted out before the 2.2 release.

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Changed
 
+* `window.fdc3` is now an optional property and may or may not be defined.
```
kriswest, line 25:
```suggestion
* `window.fdc3` is now an optional property and may or may not be defined. Applications should use `getAgent()` is now the recommended way of retrieving a reference to the FDC3 API. ([#1386](https://github.com/finos/FDC3/pull/1386))
```

src/api/DesktopAgent.ts
```diff
@@ -13,7 +13,6 @@ import { ImplementationMetadata } from './ImplementationMetadata';
 import { PrivateChannel } from './PrivateChannel';
 import { AppIdentifier } from './AppIdentifier';
 import { AppMetadata } from './AppMetadata';
-import { DesktopAgentDetails } from './GetAgent';
```
kriswest, line 16: Not actually used or re-exported - so yes this should go.

However I think we're missing a separate export of everything from GetAgent in index.ts... Probably need to add:

```ts
export * from './api/GetAgent';
```

Roaders: I've added that

kriswest (approved):
LGTM, thanks @Roaders.

src/context/ContextTypes.ts
```diff
@@ -734,34 +734,31 @@ export interface ChatSearchCriteria {
      *
      * Empty search criteria can be supported to allow resetting of filters.
      */
-    criteria: SearchCriteria[];
+    criteria: Array<OrganizationObject | string>;
```
psmulovics, line 737: What would the string entail in this case?

kriswest: @psmulovics Any search string. Easiest to understand from the example here: https://fdc3.finos.org/docs/next/context/ref/ChatSearchCriteria#example

This could be described in the comment of course.

Not actually relevant to this PR, a previous PR missed regenerating these types so these changes are showing up on them all.

kriswest:
@Roaders this change causes build errors in /src/Methods.ts, which needs to have guards installed to ensure that `window.fdc3` is defined. It's already set up to reject promises if that's the case, but TypeScript needs an explicit if/else to get that.

I can't find an elegant way to do that so in each function we need to make changes like:

Current:

```ts
export function open(app: AppIdentifier | string, context?: Context): Promise<AppIdentifier> {
  if (isString(app)) {
    return rejectIfNoGlobal(() => window.fdc3.open(app, context));
  } else {
    return rejectIfNoGlobal(() => window.fdc3.open(app, context));
  }
}
```

Updated:

```ts
export function open(app: AppIdentifier | string, context?: Context): Promise<AppIdentifier> {
  if (isString(app)) {
    return window.fdc3 ? window.fdc3.open(app, context) : Promise.reject(UnavailableError);
  } else {
    return window.fdc3 ? window.fdc3.open(app, context) : Promise.reject(UnavailableError);
  }
}
```

I.e. inline the `rejectIfNoGlobal` function.

You can find a copy of the file with those updates at: https://github.com/finos/FDC3/blob/fdc3-for-web-impl-update/packages/fdc3-standard/src/api/Methods.ts

However, don't apply the change to the import of Context on line 25.

Roaders:
would

```ts
return window.fdc3?.open() ?? Promise.reject(UnavailableError);
```

work? It should do. If fdc3 is not defined then undefined would be returned rather than a promise so in that case the reject promise would be returned.

kriswest:
> ```ts
> return window.fdc3?.open() ?? Promise.reject(UnavailableError);
> ```

Seems to work yes.

bingenito (approved):

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Changed
 
+* `window.fdc3` is now an optional property and may or may not be defined. Applications should use `getAgent()` is now the recommended way of retrieving a reference to the FDC3 API. ([#1386](https://github.com/finos/FDC3/pull/1386))
```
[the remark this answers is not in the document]
kriswest: That was a Norfolk accent creeping in 😱

PR #1278, opened 2024-07-19
Use TS Morph to generate type predicates

Roaders:
adresses #1268 

This is a very rough cut and really just proves that what I was planning does work.

We could generate the predicates in a different file if we wanted to. We could also generate union types of the request and response messages if we wanted to.

I had some issues getting ts-node working initially hence the extra tsconfig file. We can move the `generate-type-predicates` file into the root of the project if desired.

---

THIS SOFTWARE IS CONTRIBUTED SUBJECT TO THE TERMS OF THE FINOS CORPORATE CONTRIBUTOR LICENSE AGREEMENT.

THIS SOFTWARE IS LICENSED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE AND ANY WARRANTY OF NON-INFRINGEMENT, ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE. THIS SOFTWARE MAY BE REDISTRIBUTED TO OTHERS ONLY BY EFFECTIVELY USING THIS OR ANOTHER EQUIVALENT DISCLAIMER IN ADDITION TO ANY OTHER REQUIRED LICENSE TERMS.

Issue #1268, opened 2024-07-17
Enhancements to messages defined in BrowserTypes


Roaders:
### Use Case:
I am a developer implementing a Desktop Agent Proxy that will use the proposed messaging protocol to communicate with a Desktop Agent. The types defined in [BrowserTypes ](https://github.com/finos/FDC3/blob/fdc3-for-web/src/api/BrowserTypes.ts)are very helpful but there are things that can be done to make things easier.

- Have a union type of all messages called `FDC3Message` (name is just an example. I am not good at naming things! feel free to suggest a different name)
- The `FDC3Message` union type should be composed of 2 other union types `RequestMessage` and `ResponseMessage`
- Have a `RequestMessageBase` and a `ResponseMessageBase`. All response messages should have a `source: AppIdentifier` and request / response uuid in the meta object for example (I am not sure how practical this is but some commanality between messages would be great)
- Have a type predicate defined for each message type: 
```ts
export function isAddContextListenerRequest(value: any): value is AddContextListenerRequest{
    return (value as AddContextListenerRequest).type === "addContextListenerRequest"; 
}
```
This isn't necessarily the desired implementation - we could probably use some of the logic from the `Convert` class to do a better check that it is in fact a valid message. The important bit is the type predicate return type.

We have already implemeted a lot of type predicates and union types but I am not sure how useful it would be to contribute these as I understand that the `BrowserTypes` are generated by magic so that magic should probably also generate the predicates and union types.

kriswest:
@Roaders There exist AppRequestMessage, AgentResponseMessage and AgentEvent message schemas and types. See:
- https://github.com/finos/FDC3/blob/d94171986ccca791d570439ecfc366e4f9665ca0/src/api/BrowserTypes.ts#L86-L101
- https://github.com/finos/FDC3/blob/d94171986ccca791d570439ecfc366e4f9665ca0/src/api/BrowserTypes.ts#L172-L188
- https://github.com/finos/FDC3/blob/d94171986ccca791d570439ecfc366e4f9665ca0/src/api/BrowserTypes.ts#L233-L247

The types are structurally compatible with all messages derived from them so you can use them to write generalized TypeScript code in your implementation. However, as we generate the TypeScript code from the schemas, the type definitions of specific message types don't formally extend these types - but as TypeScript uses structural compatibility rather than than more traditional declarative typing that should not matter (other than it feeling a little odd to those of us that grew up with OO languages).

[...]

PR #1387, opened 2024-10-15
Use TS Morph to generate type predicates

Roaders:
adresses #1268 

(apologies for the second PR on this. I had stupidly created my last PR #1278  from a silly branch and then overwrote it).

This is a very rough cut and really just proves that what I was planning does work.

We could generate the predicates in a different file if we wanted to. We could also generate union types of the request and response messages if we wanted to.

I had some issues getting ts-node working initially hence the extra tsconfig file. We can move the `generate-type-predicates` file into the root of the project if desired.

---

### DIFF
--- a/code-generation/generate-type-predicates.ts
+++ b/code-generation/generate-type-predicates.ts
@@ -1,0 +1,50 @@
+import { InterfaceDeclaration, MethodDeclaration, Project, SyntaxKind } from "ts-morph"
+
+// open a new project with just BrowserTypes as the only source file 
+const project = new Project();
+const sourceFile = project.addSourceFileAtPath("./src/api/BrowserTypes.ts")
+
+//get a list of all interfaces in the file
+const interfaces = sourceFile.getChildrenOfKind(SyntaxKind.InterfaceDeclaration);
+
+// get a list of all conversion functions in the Convert class
+const convert = sourceFile.getClass("Convert");
+const convertFunctions = (convert?.getChildrenOfKind(SyntaxKind.MethodDeclaration) ?? []).filter(func => func.getReturnType().getText() === "string");
+
+// generate a list of Interfaces that have an associated conversion function
+const matchedInterfaces = convertFunctions.map(func => {
+    const valueParameter = func.getParameter("value");
+
+    const matchingInterface = interfaces.find(interfaceNode => {
+        return valueParameter?.getType().getText(valueParameter) === interfaceNode.getName();
+    });
+
+
+    if(matchingInterface != null) {
+        return {func, matchingInterface};
+    }
+
+    return undefined;
+}).filter(((value => value != null) as <T>(value: T | null | undefined) => value is T));
+
+// write a type predicate for each matched interface
+matchedInterfaces.forEach(matched => writePredicate(matched.matchingInterface, matched.func));
+
+
+function writePredicate(matchingInterface: InterfaceDeclaration, func: MethodDeclaration): void{
+    const predicateName = `is${matchingInterface.getName()}`;
+
+    sourceFile.addStatements(`export function ${predicateName}(value: any): value is ${matchingInterface.getName()} {
+    try{
+        Convert.${func.getName()}(value);
+        return true;
+    } catch(e: any){
+        return false; 
+    }
+}`);
+
+}
+
+sourceFile.formatText();
+
+project.saveSync();
--- a/code-generation/generate-type-predicates.ts
+++ b/code-generation/generate-type-predicates.ts
@@ -30,11 +30,15 @@ const matchedInterfaces = convertFunctions.map(func => {
 // write a type predicate for each matched interface
 matchedInterfaces.forEach(matched => writePredicate(matched.matchingInterface, matched.func));
 
+writeUnionType("RequestMessage", interfaces, "Request");
+writeUnionType("ResponseMessage", interfaces, "Response");
+writeUnionType("EventMessage", interfaces, "Event");
 
 function writePredicate(matchingInterface: InterfaceDeclaration, func: MethodDeclaration): void{
     const predicateName = `is${matchingInterface.getName()}`;
 
-    sourceFile.addStatements(`export function ${predicateName}(value: any): value is ${matchingInterface.getName()} {
+    sourceFile.addStatements(`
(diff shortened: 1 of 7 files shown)


src/api/BrowserTypes.ts
```diff
@@ -5861,3 +5861,738 @@ const typeMap: any = {
...
+export function isWebConnectionProtocolMessage(value: any): value is WebConnectionProtocolMessage {
+    try {
+        Convert.webConnectionProtocolMessageToJson(value);
+        return true;
+    } catch (e: any) {
+        return false;
+    }
+}
+
+export type RequestMessage = AddContextListenerRequest | AddEventListenerRequest | AddIntentListenerRequest | BroadcastRequest | ContextListenerUnsubscribeRequest | CreatePrivateChannelRequest | EventListenerUnsubscribeRequest | FindInstancesRequest | FindIntentRequest | FindIntentsByContextRequest | GetAppMetadataRequest | GetCurrentChannelRequest | GetCurrentContextRequest | GetInfoRequest | GetOrCreateChannelRequest | GetUserChannelsRequest | HeartbeatAcknowledgementRequest | IntentListenerUnsubscribeRequest | IntentResultRequest | JoinUserChannelRequest | LeaveCurrentChannelRequest | OpenRequest | PrivateChannelAddEventListenerRequest | PrivateChannelDisconnectRequest | PrivateChannelUnsubscribeEventListenerRequest | RaiseIntentForContextRequest | RaiseIntentRequest;
```
Roaders, line 6594: These names `RequestMessage` `ResponseMessage` and `EventMessage` probably need to change. I am open to suggestions!

src/api/BrowserTypes.ts
```diff
@@ -5861,3 +5861,738 @@ const typeMap: any = {
         "WCP6Goodbye",
     ],
 };
+
+export function isAddContextListenerRequest(value: any): value is AddContextListenerRequest {
+    try {
+        Convert.addContextListenerRequestToJson(value);
```
Roaders, line 5867: Are we happy with this "brute force" check of the types. It would be much harder to split the `Convert` class up to allow us to just check the conformance of rhte value rather than just trying to convert it to a string (which does conformance checking anyway).

code-generation/generate-type-predicates.ts
Roaders, line 1: Is this the best place for this config / code? I really didn't know so just stuck it here. Please suggest a better place to move it to if required.

src/api/BrowserTypes.ts
kriswest: Seeing this, there is another way. We already generate 3 types that will only validate against these groupings:

- `AppRequestMessage`, 
- `AgentResponseMessage`,
- `AgentEventMessage` 

and each has a union of the `type` field string  values that apply:  

- `RequestMessageType`, 
- `ResponseMessageType`, 
- `EventMessageType`

E.g. https://github.com/finos/FDC3/blob/6cf70dcac3096b6104417d613ed87516cacd3592/src/api/BrowserTypes.ts#L528-L564

These could also be used to implement a type guard.

Could you add a usage example or two so that we can evaluate what the best (and easiest to maintain) solution is?

kriswest (commented):
Added a few thoughts. Happy to discuss at Thursday's meeting.

code-generation/tsconfig.json
```diff
@@ -0,0 +1,37 @@
+{
+    // see https://www.typescriptlang.org/tsconfig to better understand tsconfigs
+   "files": ["generate-type-predicates.ts"],
+    "compilerOptions": {
+    "module": "CommonJS",
```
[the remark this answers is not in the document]
Roaders: I am pretty sure that I would have only added this as I was getting errors without but hopefully I can get it to work without if I have another go.

Roaders:
The reason for `RequestMessage` is to support Discriminated unions:

```ts

public onMessage(message: RequestMessage): void{
    switch(message.type){
         case: "raiseIntentRequest":
                return this.handleRaiseIntentRequest(message); // here message is typed as RaiseIntentRequestMessage rather than RequestMessage
        case "raiseIntentForContextRequest":
                return this.handleRaiseIntentForContextRequest(message); // message again correctly typed
        default:
                const unhandled: never = message.type; // compiler error if we have not handled all messages
                console.log(`Unknown message type encountered: ` + message.type);
    }

}

private handleRaiseIntentRequest(message: RaiseIntentRequestMessage): void {}

private handleRaiseIntentForContextRequest(message: RaiseIntentForContextMessage): void {}

```

code-generation/tsconfig.json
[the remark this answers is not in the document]
kriswest: The original one has an include filter that would have excluded your codegeneration directory: https://github.com/finos/FDC3/blob/1a21b027f506d9ed88d5e3a1d55be38ac03ae828/tsconfig.json#L3C25-L3C25

The only other difference is your `files` entry above - could it have been that it just wasn't applying to your file?

Roaders:
Notes from FDC3 for Web Browsers Discussion group 17th October 2024 #1391 

It would be better if didn't rely on the interface name to build the union types. If we could instead check that a given interface is logically compatible with `AgentResponseMessage`, `AgentEventMessage` or `AppRequestMessage` and use that to build the union types that would be better.

robmoffat:
Hi @Roaders ,

I would like to merge this into the fdc3-for-web-impl branch, which will form the basis of the implementation for 2.2
 Can we rebase this PR against that?  I could probably have a go at it if you don't want to.

The PR for this is #1309, which also changes the FDC3 project to use a monorepo structure.  Your wor would need to go into the `packages/fdc3-schema` module, which generates BrowserTypes.ts.  You might want to also have something in `packages/fdc3-context` which generates the various FDC3 standard context typescript definitions.  

Let me know if you want to work on this, otherwise I might be able to take a look at it

robmoffat:
@kriswest should this be closed?

kriswest:
No, not sure how I did that. I think it might have been automated after I deleted the branch it targeted.

kriswest:
Re-opened and now pointed at main, could do with an update and `npm run build` or a rebase onto the fdc3-for-web-impl branch

robmoffat:
I'm going to merge this into fdc3-for-web-impl - turns out we need it there already!

PR #1424, opened 2024-11-08
Updates from impl review

julianna-ciq:


packages/fdc3-agent-proxy/src/Messaging.ts
```diff
@@ -1,6 +1,6 @@
 import { Connectable, AppIdentifier, ImplementationMetadata } from "@kite9/fdc3-standard";
 import { RegisterableListener } from "./listeners/RegisterableListener";
-import { AddContextListenerRequest, AddContextListenerRequestMeta } from "@kite9/fdc3-schema/generated/api/BrowserTypes";
+import { AddContextListenerRequestMeta } from "@kite9/fdc3-schema/generated/api/BrowserTypes";
```
kriswest, line 3: Only top-level types should be imported from BrowserTypes (as the naming of the lower-level types can be unstable),. Hence, I'd import `AppRequestMessage` and then extract the type for its meta element:

```ts
import { AppRequestMessage} from "@kite9/fdc3-schema/generated/api/BrowserTypes";

type RequestMetadata = AppRequestMessage["meta"];
```
then use that down on line 33.

packages/fdc3-agent-proxy/src/channels/DefaultChannel.ts
```diff
@@ -25,27 +25,29 @@ export class DefaultChannel implements Channel {
...
             meta: this.messaging.createMeta(),
             payload: {
                 channelId: this.id,
                 context
             },
             type: "broadcastRequest"
-        } as BroadcastRequest, 'broadcastResponse')
-        console.log("broadcast done", done)
+        };
+        await this.messaging.exchange<BroadcastResponse>(request, 'broadcastResponse')
```
kriswest, line 36: 👍 👍 👍

packages/fdc3-agent-proxy/src/messaging/AbstractMessaging.ts
```diff
@@ -30,48 +29,59 @@ export abstract class AbstractMessaging implements Messaging {
...
     async exchange<X>(message: any, expectedTypeName: string, timeoutErrorMessage?: string): Promise<X> {
         const errorMessage = timeoutErrorMessage ?? `Timeout waiting for ${expectedTypeName} with requestUuid ${message.meta.requestUuid}`
-        const prom = this.waitFor(m =>
+        const messageResponseWaiter = this.waitFor<X>(m =>
             (m.type == expectedTypeName)
             && (m.meta.requestUuid == message.meta.requestUuid), errorMessage)
         this.post(message)
-        const out: any = await prom
+        const out = await messageResponseWaiter
+        // @ts-expect-error: TODO: We should make a list of all of the possible messages that can be called through this, and make that list the generic. This will ensure better type safety.
```
kriswest, line 82: I'm pretty sure that relevant types already exist for that: AppRequestMessage, AgentResponseMessage, AgentEventMessage were intended for such usage and contain unions of the relvant type strings for messages in their set.

packages/fdc3-agent-proxy/src/messaging/AbstractMessaging.ts
```diff
@@ -102,15 +112,18 @@ export abstract class AbstractMessaging implements Messaging {
         this.implementationMetadata = null;
     }
 
-    getImplementationMetadata(): Promise<ImplementationMetadata> {
-        return Promise.resolve(this.implementationMetadata!!)
+    async getImplementationMetadata(): Promise<ImplementationMetadata> {
+        if(this.implementationMetadata === null){
```
kriswest, line 116: Reasonable to throw an error - where I hit this in another branch I generated impl metadata with "unknown" values. It shouldn't happen, but it pays to be defensive and to handle the error if it ever comes up.

packages/fdc3-agent-proxy/src/messaging/AbstractMessaging.ts
```diff
@@ -102,15 +112,18 @@ export abstract class AbstractMessaging implements Messaging {
...
+            throw new Error("Unconnected. No metadata available.")
+        }
+        return this.implementationMetadata
     }
 
     private async exchangeValidationWithId<X>(message: any, connectionAttemptUuid: string): Promise<X> {
```
kriswest, line 122: I'm moving the identity validation into the Loaders and out of here in another branch. Its part of the connection flow for Web and should be separated from messaging (that could be used with a different connection).

packages/fdc3-get-agent/src/strategies/TimeoutLoader.ts
```diff
@@ -26,9 +26,9 @@ export class TimeoutLoader implements Loader {
         this.done = true;
     }
 
-    get(params: GetAgentParams): Promise<DesktopAgent | void> {
+    get({timeoutMs = 0}: GetAgentParams): Promise<DesktopAgent | void> {
```
kriswest, line 29: There is a constant for the default timeout elsewhere - but I'm also eliminating the timeoutLoader in another branch (with id validation moving into the loader, loaders can do some ops outside of the timeout so they need to manage the timeouts themselves

packages/fdc3-get-agent/src/ui/DefaultDesktopAgentChannelSelector.ts
```diff
@@ -25,30 +25,32 @@ export class DefaultDesktopAgentChannelSelector extends AbstractUIComponent impl
...
-        // also send to the iframe
-        this.port!!.postMessage({
+        const message: Fdc3UserInterfaceChannels = {
             type: 'Fdc3UserInterfaceChannels',
             payload: {
                 selected: channelId,
                 userChannels: availableChannels.map(ch => {
                     return {
+                        // TODO: Is this the correct type to return for these channels?
+                        type: "user",
```
kriswest, line 45: Yes thats the correct type.
```suggestion
                        type: "user",
```

packages/fdc3-get-agent/test/step-definitions/intent-resolver.steps.ts
```diff
@@ -4,7 +4,7 @@ import { handleResolve } from "@kite9/testing";
 import { DefaultDesktopAgentIntentResolver } from "../../src/ui/DefaultDesktopAgentIntentResolver";
 import { INTENT_RESPOLVER_URL } from "../support/MockFDC3Server";
 
-
+// TODO: Replace 'any' with useful typings
 const contextMap: Record<string, any> = {
```
kriswest, line 8: The `Context` type would be the obvious one 
```suggestion
import { Context } from "@kite9/fdc3-context";

const contextMap: Record<string, Context> = {
```

toolbox/fdc3-for-web/demo/src/client/ui/channel-selector.ts
```diff
@@ -4,8 +4,9 @@ type IframeChannels = BrowserTypes.Fdc3UserInterfaceChannels
 type IframeRestyle = BrowserTypes.Fdc3UserInterfaceRestyle
 type IframeHello = BrowserTypes.Fdc3UserInterfaceHello
```
kriswest, line 5: Probably best to stick to the actual (new / replacement) names for these types

kriswest (approved):
This looks like a solid set of changes that will reduce the number of warnings from linting through better typing. I'll make a couple of quick changes to deal with some of my comments.

@robmoffat once done, I think you should merge this. I'll need to deal with some conflicts in my branch thereafter.

packages/fdc3-agent-proxy/src/messaging/AbstractMessaging.ts
kriswest: Those types can't be easily applied yet due to these functions serving both WCP and DACP - I'll fix this in my branch where I've unpicked that already

kriswest:
Resovled conflicts @robmoffat - this should be ready for you.

One thing to watch out for: don't import the various generated types in BrowserTypes other than the top-level message types, as the naming of the autogenerated types may be unstable. That also goes for constants used for message `type` fields. Instead you can:
- Apply a top-level message type to a whole message - it will highlight if the string type is wrong and help you correct
- Use 'indexed access types' for typing of properties. E.g. if you just want the payload of a message type then do:
    
    ```ts
    import { Fdc3UserInterfaceRestyle } from "@kite9/fdc3-schema/dist/generated/api/BrowserTypes";
    type RestylePayload = Fdc3UserInterfaceRestyle ["payload"]
    ```
    That way you are not dependent on the auto-generated sub-types.

robmoffat:
Have you tried running the tests on this @kriswest?  

I get this error (consistent with the coverage run): 

```
TypeError: Class extends value undefined is not a constructor or null
    at Object.<anonymous> (/home/runner/work/FDC3/FDC3/packages/fdc3-get-agent/src/ui/DefaultDesktopAgentIntentResolver.ts:8:1)
    at Module._compile (node:internal/modules/cjs/loader:1469:14)
    at Module.replacementCompile (/home/runner/work/FDC3/FDC3/node_modules/append-transform/index.js:60:13)
    at Module.m._compile (/home/runner/work/FDC3/FDC3/node_modules/ts-node/src/index.ts:1618:23)
    at module.exports (/home/runner/work/FDC3/FDC3/node_modules/default-require-extensi
```

This might be fixable by merging main back into the branch, perhaps?

kriswest:
@robmoffat found it! It was the import of FDC3_VERSION in [packages/fdc3-get-agent/src/ui/AbstractUIComponent.ts](https://github.com/finos/FDC3/pull/1424/files?show-deleted-files=true&show-viewed-files=true&file-filters%5B%5D=#diff-956eac765f30be876bdbafffa26437ff49c21b56e4a0b0854d995e38b7aa3457). The PR fixes an incorrect payload in an `Fdc3UserInterfaceHandshake` message by adding the missing FDC3 version number, but creates a circular dependency.

I've corrected the dependency by moving the FDC3 version into its own file - although that should probably be moved to the top of the dependency chain (fdc3-standard package)? That then revelaed some other issues (unused imports in tests). I've corrected those and now it runs, but seems to get stuck, its also timing out in a few tests - I'm not sure why. Could you run the tests and see if you can replicate?  I may be on a different node version to you (20.11.1) which might make a difference

PR #1428, opened 2024-11-12
Move addon package to toolbox/fdc3-for-web/reference-ui

julianna-ciq:


kriswest (commented):
Noted a few out of date message names (which may not be the subject of this PR - but they were edited in this one so thought to point out). I'm also uncertain about the build outputting directly into the website's static dir. It might be better to output within the reference-ui project then copy to website location.

toolbox/fdc3-for-web/reference-ui/src/channel_selector.ts
```diff
@@ -49,11 +49,11 @@ window.addEventListener("load", () => {
   myPort.onmessage = ({ data }) => {
     console.debug("Received message: ", data);
     switch (data.type) {
-      case "iframeHandshake": {
+      case "IframeHandshake": {
         collapse();
         break;
       }
-      case "fdc3UserInterfaceChannels": {
+      case "Fdc3UserInterfaceChannels": {
```
robmoffat, line 56: @kriswest is there a constant defined somewhere with these in?  I think there should be

robmoffat (changes requested):

toolbox/fdc3-for-web/reference-ui/src/channel_selector.ts
julianna-ciq: While BrowserTypes.ts does define the type `Fdc3UserInterfaceMessageType`, which lists all of the available names, it's not a constant that could be used elsewhere.

So, I've updated the files to refer to the message type. This way, if there are changes to the structure (or renaming the individual types), typescript will be aware of which files need to change.

kriswest: @robmoffat I think @Roaders TsMorph PR is closely related to your question. If memory serves it achieves two related things: 
- you can use the `RequestMessage`, `ResponseMessage` and `EventMessage` to get smarter typing through a switch statement like this (see https://github.com/finos/FDC3/pull/1387#issuecomment-2419628543)
- it creates a bunch of `isFdc3UserInterfaceChannels` type functions that you could use in the switch statement instead of the string (although I'd question the utility of that as the string is embedded in that function name and would change along with it).

QuickType was generating enums you could reference, but we switched everywhere to use unions instead (where you can't directly refer to individual values): https://github.com/finos/FDC3/blob/aaa58067f32ece40cb0009f7a615ca2d394f4f9d/packages/fdc3-schema/generated/api/BrowserTypes.ts#L1842-L1842C286
I don;t remember why we did, but I do remember there was a good reason to do so - I think because you *had* to use the enum reference which is not equal to the string it represents.

So to sum up, I think Julianna is right to use the types as that will highlight incorrect values. The types could (possibly) be better (via the TSMorph PR). THere are no accessible constants generated currently.

robmoffat: Yes, I think we're going to have to bite the bullet and merge in the TSMorph stuff to get this...

Issue #1351, opened 2024-09-12
CommonJs vs ES6


robmoffat:
For Background:

@Lecss @kriswest and I built the "combined" @finos/fdc3 module in the FDC3 repo, combining all the other modules.  For now, this uses ES6 module structure, rather than the FDC3 rollup approach used originally. 

However, the original FDC3 module was commonJS.  I'm not sure of the ins and outs of this, so:

1.  Is it backwards incompatible to move to ES6?  Or does Javascript correctly handle both?
2. ES6 import seems like the way we want to go forward is this the right time to change?
3. If we have an old, legacy CJS module, should we expect people to import the monorepo's submodules e.g. fdc3-context?  This seems unlikely to work given the amount of literature around the old version.

I want to open this up for further discussion as I'm not an expert on this.

robmoffat:
@novavi

robmoffat:
@julianna-ciq

kriswest:
My limited understanding is that the current module contains both:
https://unpkg.com/browse/@finos/fdc3@2.1.1/dist/
and I would _assume_ its not backwards compatible to move to ESM only... but again I'm no expert on that topic. @julianna-ciq or @thorsent might know more.

thorsent:
TLDR; either configure rollup to produce both commonjs and esm (if you want to be a good citizen), or just go with esm (you'll be fine)

The approach taken by most libraries these days is to produce both commonjs and esm. Rollup can do this pretty easily (the "main" entry in package.json points to the cjs version, while the "module" entry points to the esm version).

Some libraries have chosen to only support esm (an effort to strong-arm the world into moving forward), so people are learning to deal with it.

In general, commonjs is only necessary for node (and usually only in legacy situations). On the web, browsers already support esm as do all the major bundlers. Node however is very persnickety about esm , requiring either that all modules are esm (via a switch) or that modules specifically identify themselves using the .cjs and .mjs notation to indicate their module type.

We most often see this issue rear its head when building unit tests, since mocha and JEST run in node.  Anyone in the world who has incorporated the FDC3 library, or needs to incorporate the FDC3 library, in a mocha suite that isn't configured for esm will have issues (as they will with many other public libraries - the solution usually being to use an older version...)

Typically, if you don't have a commonjs option then you have to go "all in" on esm in node in your unit test pipeline, which means including extensions in your imports:

`import "../mymodule.js"` instead of `import "../mymodule"`

[...]

Issue #1297, opened 2024-07-29
FDC3 monorepo


robmoffat:
## Enhancement Request

We should use the new Node monorepo features so that we can isolate different parts of the FDC3 build and incorporate FDC3 for the web, which is already using a monorepo structure.

```[tasklist]
### Tasks
- [x] Moved FDC3 For the Web code into the main FDC3 Repo
- [x] Convert from Yarn to NPM workspaces (fdc3-for-web-implementation part)
- [x] Apply monorepo structure to the whole of the FDC3 project
- [x] Rationalise / clean up package structure
- [x] Move FDC3 For Web demo / da-server somewhere else
- [x] FDC3 Workbench - runs fine with npm run dev but fails with build
- [x] Error message: /bin/sh: ../node_modules/.bin/quicktype: No such file or directory
- [x] Fix issue #1209
- [x] fdc3-core -> fdc3-standard (Alex)
- [x] Split out Schema into a module (Alex)
- [x] Remove fdc3-common (RM)
- [x] FDC3 Core Tests Are Breaking (Rob)
- [x] move da-server/demo/addon into toolbox
- [x] move husky to root of project - Added GitHub action instead
- [x] Rename client to fdc3-get-agent
- [x] Rename fdc3-core to fdc3-standard
- [x] Work on the main (@finos/fdc3) FDC3 npm package build so that it will contain FDC3 for the web. Handle root exports. We'll do this in ES6 format for now. Handle package publication so that individual and combined packages are released to npm. Handle root exports / naming clashes (Alex, Kris and Rob)
- [x] Rename da-proxy to fdc3-agent-proxy
- [x] Rename da-server to fdc3-web-impl
- [x] Ensure all package.json files have consistent project metadata
- [x] Look at docusaurus build and make sure schemas are imported correctly (Kris, Julianna)
- [x] Remove ag-grid / ChartIQ / fix semgrep issues (RM, Julianna)
- [x] move addon into toolbox/fdc3-for-web/reference-uis, rename and apply this into the docusaurus build in website/package.json (RM, Julianna)
- [x] Add Coverage report to GitHub Actions (ALEX)
- [x] add nyc to fdc3-standard
- [x] report test results
- [x] write READMEs for the different sub-projects
- [ ] commonJS support (RM)
```

robmoffat:
- #1298 

Handles box 2

robmoffat:
## New Repo Structure

```
package.json  # Top of the Monorepo
packages / 
  - client
[...]

PR #1452, opened 2024-11-28
Fdc3 for web impl commonjs

robmoffat:
This is to complete #1297 and address https://github.com/finos/FDC3/issues/1351, adds rollup to the `fdc3` package so that common-js is still supported.

Since I know CJS is used in node, I ran the following:

```
var fdc3CJS= require('@kite9/fdc3-commonjs')

console.log("Got FDC3", fdc3CJS)
```

From within the `fdc3-for-web/demo` process and got back:

```
Got FDC3 {
  AgentError: {
    AgentNotFound: 'AgentNotFound',
    AccessDenied: 'AccessDenied',
    ErrorOnConnect: 'ErrorOnConnect',
    InvalidFailover: 'InvalidFailover'
  },
  OpenError: {
    AppNotFound: 'AppNotFound',
    ErrorOnLaunch: 'ErrorOnLaunch',
...
  Convert: [class Convert],
  addContextListener: [Function: addContextListener],
  addEventListener: [Function: addEventListener],
  addIntentListener: [Function: addIntentListener],
  broadcast: [Function: broadcast],
  compareVersionNumbers: [Function: compareVersionNumbers],
  createPrivateChannel: [Function: createPrivateChannel],
  fdc3Ready: [Function: fdc3Ready],
  findInstances: [Function: findInstances],
  findIntent: [Function: findIntent],
  findIntentsByContext: [Function: findIntentsByContext],
  getAgent: [Function: getAgent],
  getAppMetadata: [Function: getAppMetadata],
  getCurrentChannel: [Function: getCurrentChannel],
  getInfo: [Function: getInfo],
  getOrCreateChannel: [Function: getOrCreateChannel],
  getSystemChannels: [Function: getSystemChannels],
  getUserChannels: [Function: getUserChannels],
  isStandardContextType: [Function: isStandardContextType],
  isStandardIntent: [Function: isStandardIntent],
  joinChannel: [Function: joinChannel],
  joinUserChannel: [Function: joinUserChannel],
  leaveCurrentChannel: [Function: leaveCurrentChannel],
  open: [Function: open],
  raiseIntent: [Function: raiseIntent],
  raiseIntentForContext: [Function: raiseIntentForContext],
  versionIsAtLeast: [Function: versionIsAtLeast]
}
```

Which looks ok to me?   Obviously, I can't call getAgent, so not sure how further to test this.  

Nb, this is very much simplified from the original `rollup.config.js` (preserved in the `packages/fdc3/old` dir) but I think a lot of the assumptions of that are not necessarily relevant anymore, perhaps?


kriswest:
The old config built two sets of files into the module, which consumers could switch between by setting a NODE_ENV variable, which this config won't produce. I'm no export on these config, but it also looks like you are only producing common.js files, and in their development form (all comments etc. retained as terser is not used to dtrip them out).

robmoffat:
HI @kriswest,

Yes that's right.  The main difference between the two versions being that the production one was minified using terser.  

Based on @thorsent's feedback about this, it seems like the use case for CJS this is automated tests on very legacy versions of node.  Given that, it seemed best that we provide the un-minified version by default so at least people can figure out where their tests are going wrong.

kriswest:
One of the participating teams fed back that they were still using CommonJS in builds. 

The advantage of having both types in the module is that, based on whether you are doing a dev or prod build, it includes the relevant version. If you then bundle and treeshake your own app you app code only then includes the version you use, so prod builds are leaner, while dev builds are unminimized, including comments, maps etc..  I have no idea if an app can do that minification itself if we don't.

At least thats my (limited) understanding. @julianna-ciq  probably understands it better?

robmoffat:
Do we actually know who is using CJS?  It might be simpler to just ask that one person!  I have a feeling this is super-niche and not worth spending a lot of time on, but at the same time it'd be nice not to give that one person a headache.

robmoffat:
The reason it's so different is that basically I had to start over and read the rollup docs on how to do this - the original script (preserved) didn't work out-of-the-box.   

As I looked into it, it seemed like a lot of the code added to it probably wasn't relevant to what we were doing anymore.  For example, it referenced react, but we don't have any react in the fdc3 module.  

So it's tricky. We could spend effort trying to preserve the NODE_ENV stuff if that's what you want?  But (going by @thorsent's notes)  it seems like an odd thing to have in there just to reduce the size of a download for the node environment, especially given the node environment is going to be pulling in the entire dependency into it's `node_modules`:  effectively having a minified version is just going to increase the size of the download!

kriswest:
The react reference is only a minor part of that config (a couple of lines), the majority is related to preparing the variations on the module. I'm not convinced its producing both CommonJs and ES6 versions now. Also, the goal is not to reduce the module size downloaded by NPM, but rather the output bundle size of a project consuming the module. The module is going to increase in size with 2.2 and the loss of a minified version to pull in _may_ exacerbate that increase in size for downstream projects.

I think we need to understand the impact of changes to the module, before releasing a version with those changes.

kriswest:
The react part of the config is probably inherited from tsdx's rollup config, as that is what was originally used to build the module (and also where the dev/prod build setup comes from).

robmoffat:
@kriswest did this get raised in the SWG meeting last week at all?

kriswest:
I mentioned it, few opinions were offered and no conclusion drawn unfortunately

robmoffat:
Shall I close this then?

packages/fdc3/rollup.config.mjs
```diff
@@ -0,0 +1,38 @@
...
+            format: "cjs",
+            freeze: false,
+            esModule: false,
+            sourcemap: true,
+            exports: "named",
+        },
+
+        plugins: [
+            nodeResolve(),
+            commonjs({
```
kriswest, line 29: Still needs switching back to ES6 here

kriswest:
> Shall I close this then?

Sorry to clarify my previous response: I mentioned the idea to publish two modules, ES6 and CommonJS, allowing for a simple resolution for anyone's build that gets broken by switching to ES6 only and there were no objections. I think @novavi liked the idea of being able to track usage and retire the CommonJs module when its not in use.

Hence, I think we should proceed as we've discussed offline @robmoffat, and it looks like you are on your way there already.

robmoffat:
yeah think this is done as discussed.  WDYT?

README.md
```diff
@@ -93,7 +93,9 @@ This project (the FDC3 Standard repo) is now a monorepo containing the following
...
 | `toolbox/fdc3-for-web/demo`          | -not released-                                         | A reference implementation of an FDC3-For-The-Web Desktop Agent, using the above implementation and conforming to FDC3 2.0 from the  [FDC3 Conformance Framework](https://github.com/finos/FDC3-conformance-framework) |                       |
-| `packages/fdc3`                      | `@finos/fdc3` (npm)                                    | A complete roll-up of the `fdc3-agent-proxy`, `fdc3-context`, `fdc3-schema` and `fdc3-get-agent` sub-modules.  This is intended to be the main entry point for typescript / javascript applications using FDC3         |                       |
+| `packages/fdc3`                      | `@finos/fdc3` (npm)                                    | Imports `fdc3-agent-proxy`, `fdc3-context`, `fdc3-schema` and `fdc3-get-agent` sub-modules.  This is intended to be the main entry point for typescript / javascript applications using FDC3         |                       |
```
kriswest, line 96: fdc3-agent-proxy is imported by fdc3-get-agent, rather than directly by this module. `fdc3-standard` needs to be added to the list
```suggestion
| `packages/fdc3`                      | `@finos/fdc3` (npm)                                    | Imports `fdc3-standard`, `fdc3-context`, `fdc3-schema` and `fdc3-get-agent` sub-modules.  This is intended to be the main entry point for typescript / javascript applications using FDC3         |                       |
```

kriswest, line 96: fdc3-agent-proxy is imported by fdc3-get-agent, rather than directly by this module. `fdc3-standard` needs to be added to the list
```suggestion
| `packages/fdc3`                      | `@finos/fdc3` (npm)                                    | Imports `fdc3-standard`, `fdc3-context`, `fdc3-schema` and `fdc3-get-agent` sub-modules.  This is intended to be the main entry point for typescript / javascript applications using FDC3         |                       |
```

kriswest (changes requested):
Needs a changelog adding (for the siwtch to searpate ES6 and commonJS modules).

Also, these packages will (I assume!) publish with the readme file from each package. `fdc3-commonjs` is lacking a Readme file currently  as is `fdc3`, see https://www.npmjs.com/package/@kite9/fdc3 for what that looks like.

In the past the NPM package has inherited the repositoy's README file, which is not ideal. This is good oppurtunity to develop something more relevant, including content for both App and Desktop Agent developers as very basic getting started content. Each readme should also include an early reference to the other package and an explanation of which to use.

packages/fdc3/rollup.config.mjs
kriswest: was done

kriswest:
LGTM

kriswest (approved):
LGTM

packages/fdc3-commonjs/README.md
```diff
@@ -0,0 +1,56 @@
+# FDC3 CommonJS Module
+
+This is the CommonJS module version of the Financial Desktop Connectivity and Collaboration Consortium (FDC3) javascript API. 
+## Warning
```
kriswest, line 4:
```suggestion

## Warning
```

packages/fdc3-commonjs/README.md
```diff
@@ -0,0 +1,56 @@
...
+// Example usage
+const contact = {
+  type: 'fdc3.contact',
+  name: 'Jane Doe',
+  id: {
+    email: 'jane@mail.com'
+  }
+}
+
+fdc3.raiseIntent('ViewProfile', contact)
```
kriswest, line 35:
```suggestion
};

fdc3.raiseIntent('ViewProfile', contact);
```

packages/fdc3/README.md
```diff
@@ -0,0 +1,52 @@
...
+// Example usage
+const contact = {
+  type: 'fdc3.contact',
+  name: 'Jane Doe',
+  id: {
+    email: 'jane@mail.com'
+  }
+}
+
+fdc3.raiseIntent('ViewProfile', contact)
```
kriswest, line 31:
```suggestion
};

fdc3.raiseIntent('ViewProfile', contact);
```

packages/fdc3/README.md
```diff
@@ -0,0 +1,52 @@
+# FDC3
+
+This npm module is the main entry point for the Financial Desktop Connectivity and Collaboration Consortium (FDC3) javascript API.
```
kriswest, line 3:
```suggestion
This npm module is the main entry point for the Financial Desktop Connectivity and Collaboration Consortium (FDC3) JavaScript API.

Since version 2.2.0 this module only provides an ES6 module. Applications requiring a CommonJS module should either update to ES6 or switch to the [@finos/fdc3-commonjs module](https://npmjs.com/package/@finos/fdc3-commonjs) (provided for backwards compatibility).
```

PR #1472, opened 2024-12-13
(DO NOT MERGE) Testing CI checks III

kriswest:
## Describe your change

Testing CI check when run on a PR from a repository fork (which are failing at the last step, posting a comment with test results)

### DIFF
--- a/packages/fdc3-agent-proxy/src/listeners/AbstractListener.ts
+++ b/packages/fdc3-agent-proxy/src/listeners/AbstractListener.ts
@@ -1,3 +1,4 @@
+import { AgentEventMessage } from '@kite9/fdc3-schema/generated/api/BrowserTypes';
 import { Messaging } from '../Messaging';
 import { RegisterableListener } from './RegisterableListener';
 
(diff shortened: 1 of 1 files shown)


kriswest:
No dice @robmoffat, this PR doesn't touch the workflow file at all, it's got the pull request perms in it already, but it still fails.

This thread from the mocha team explains the issue well, we're getting a read-only token as the run is being triggered by a PR from a fork: https://github.com/mochajs/mocha/issues/4553

There is more from Github here that explains possible workarounds using the pull_request_target workflow - but also that we probably shouldn't use it as it will mean 3rd parties can execute code in the context of the repository. I've only skim-read it, there may be more.

As the bit that is failing is the posting of the comment, a way forward would be to just remove that. Let the check pass or fail and we rely on the check's status indicator rather than a posted message. That seems sufficient. However, I f you wreally want to retain the comment you could do so by just inhibiting it for PRs from forks, which you can detect by comparing the head and base repos in the PR, as they do here: https://github.com/danger/danger-js/issues/918#issuecomment-645814203

Feel free to close this PR once you've had a read of this.

PR #1541, opened 2025-03-18
User channel change issue

### LINKED ISSUE #1540: When the User Channel Changes, DA Proxy ContextListeners fail to recognise the change
## Minor Issue

This occurs when you change channel via the channel selector rather than using the internal `fdc3.join..` function.  See associated PR for fix.  

### Area of Issue
 - [ ] App Directory
 - [x] API
 - [ ] Context Data
 - [ ] Intents
 - [ ] Desktop Agent Bridging
 - [ ] Use Cases
 - [ ] Other

### Issue Description: 
On the server the channel definition might get updated.  This means that messages get sent on the new user channel to the client, but pre-existing listeners aren't moved over to the new channel, meaning they don't receive messages.

### Additional Context:
PR Incoming.   This is all to do with the `channelChangedEvent` message.  I've also extended this message so that Desktop Agents can warn the DA Proxy about other user channel details changing.

robmoffat:
## Describe your change

DA proxy code needs to listen for channelChanged events from the DA and action them (if necessary) in order to handle channel changes performed outside of the window (e.g. in DAs that load apps into iframes and display their own channel selector, rather than an injected one, outside of the window).

### Related Issue

resolves #1540

## Review Checklist

- [x] **Issue**: If a change was made to the FDC3 Standard, was an issue linked above? #1540


robmoffat:
Actually a single commit - the rest is from the release branch which we need to merge first

packages/fdc3-schema/schemas/api/channelChangedEvent.schema.json
```diff
@@ -42,6 +42,14 @@
 							"type": "null"
 						}
 					]
+				},
+				"userChannels": {
```
kriswest, line 46: I don't think this should be here - the event should not have to carry all the user channel definitions, which are supposed to remain constant (there are no event s for changes in the user channel definitions which are supposed to remain fixed). These should already have been retrieved by the client code and hence this duplication is inefficient.

At most the definition of the channel being selected could be included - but even that should not be necessary.

Also this change to the message schema will need to be approved by the SWG, where a change to the implementation only can be approved by the maintainers.

kriswest (changes requested):
I don't think we should be changing the message schema to resolve this issue - there is already state in this class containing the user channel defs that should be used, rather than re-transmitting it with every event. 

I want to spend a bit more time with the client code to confirm that we're not listening to this event elsewhere - I could have sworn we had this handled already.

kriswest:
> Actually a single commit - the rest is from the release branch which we need to merge first

Changed the base to point at that release branch (and filled in summary/issue) - if that base is merged first the base will automatically change itself to main.

robmoffat:
Would you prefer if I 

a) reverted the change to `channelChangedEvent.schema.json` and then changed `DefaultChannelSupport` to ask for the list of channels to something like:

```
// DefaultChannelSupport.ts

  this.addChannelChangedEventHandler(async (e: ApiEvent) => {
      const cce = e.details as ChannelChangedEventPayload;
      Logger.debug('Desktop Agent reports channel changed: ', cce.newChannelId);
      const newChannel = (await this.getUserChannels()).find(uc => uc.id === cce.newChannelId) ?? null;
      this.userChannelListeners.forEach(l => l.changeChannel(newChannel));
      this.channelSelector.updateChannel(newChannel?.id ?? null, this.userChannels);
    });
```

The advantage of this is that we wouldn't need to change the event message, but I do want to query the user channels in case they have changed, hence `(await this.getUserChannels())`.  The disadvantage of which is a roundtrip back to the da.

kriswest (dismissed):
LGTM, will need another approver (or for me to not be the last push + reapprove)

robmoffat:
>  Hence, we should (IMHO) support the edge-case, but not focus on it (by querying the channels every time).

The problem with this is the channel selector:  unless you call getUserChannels() or you are changing to a new channel, then it will be displaying the old user channels.  Thus, if the desktop agent adds channel 5, but you're on channel 2, you won't know about it.  You'll have to restart the app to see the correct channels again, which is a workaround but not ideal.

> The default web fdc3 channel selector wasn't working for me from fdc3.finos.org as its pointing at a (now non-existent) netlify PR preview URL, I've fixed that.

nice, thanks

> FDC3 workbench still has to be started locally for the demo - I think that can now switch to the published version - I have NOT fixed that.

yes, great idea.

kriswest:
> The problem with this is the channel selector: unless you call getUserChannels() or you are changing to a new channel, then it will be displaying the old user channels. Thus, if the desktop agent adds channel 5, but you're on channel 2, you won't know about it. You'll have to restart the app to see the correct channels again, which is a workaround but not ideal.

Just retrieving the channel data from the the DA doesn't push it into the injected channel selectors - something extra would be needed for that. Also the channel being set should be the thing triggering such an update as by that point a channel has already been selected in an external channel selector (which wouldn't normally be used alongside an injected one).

The user channel set changing has never been a thing in FDC3 which is why there are no events telling apps that the set of channels has changed. It is something we could do - but its really a new feature and we need to do more than we are here (e.g. an event from DA saying channels have changed). Hence, if we want to talk about doing it a different issue should be raised and discussed for 2.3 IMHO.

robmoffat:
> Just retrieving the channel data from the the DA doesn't push it into the injected channel selectors - something extra would be needed for that.

The reason I ended up with my original commit (with the change to the message etc.) was because of this line:

```
      this.channelSelector.updateChannel(newChannel?.id ?? null, this.userChannels);
```

... in which we are pushing potentially state data (this.userChannels) out to the injected channel selectors.   

Now, as I said in the previous comment, I think I can work around this by ignoring the user channel information and passing it separately with web sockets, but it still seems like we missed a step by not adding this to the incoming message or making sure it's correct before we send this out.

kriswest:
> Shall I raise another issue for this?

I think you should. The description on the message schema for FDC3UserInterfaceChannels: https://github.com/finos/FDC3/blob/9e0c4042c2552e988eaadcd37b78c1a7d1cb88e1/packages/fdc3-schema/schemas/api/fdc3UserInterfaceChannels.schema.json#L5

Describes it as being used for initial set-up, not for handling changes to the set of channels. I assume its *also* being used to set the channel when the app calls `joinUserChannel` (i.e. channel wasn't set by the channel selector so it doesn't know about the change unless told) - which is how we got to here... if we'd thought about that case we probably would have allowed for [Fdc3UserInterfaceChannelSelected](https://github.com/finos/FDC3/blob/main/packages/fdc3-schema/schemas/api/fdc3UserInterfaceChannelSelected.schema.json) to be sent in either direction and then it wouldn't have been including the set of channels.

As I said, the standard doesn't explicitly say that the set of User channels doesn't change, just that they "are created and named by the desktop agent". However, it also makes no provision for changes (I suspect because no one actually thought that they would get changed) and if we start doing that it should be done properly (i.e. discussed and approved by the SWG) and with a complete solution (e.g. include an event for app's to know that the set of channels has changed, create a message for the DA to notify the proxy - which would be shared with the aforementioned event).

The issue that this PR address is related but different and, IMHO, is just an implementation issue we can resolve through the maintainers, as channels being changed by the DA (with an external channel selector or similar feature) is an existing use case and easily solvable with the proposed code tweaks - it can go out in a 2.2.x NPM release.

Hence, if raising an issue I'd call it something like "Support changes to the set of User Channels in the Desktop Agent API" and we can confirm (or refute) that it is thing you should be able to do as part of that.

robmoffat:
ok will do.  I am baffled by the idea that no one has ever wanted to change the user channels.

robmoffat:
Would like to merge this but I need to merge v2.2.0-beta.3 PR first.

kriswest:
> ok will do. I am baffled by the idea that no one has ever wanted to change the user channels.

They're simply a set of predefined colour channels for a user to select from, the use case has always been changing the selected channel rather than the set you select from. I'm aware of multiple products that will accept a configuration for the set (and of course the standard recommends a particular, fixed set of channel defs), but to my knowledge no one has expressed a use case or need to change them at runtime. 

Where you have the ability to save or share a workspace or layout of apps, with channel selections, a changeable set of channels presents a problem as the links between apps in the layout might be broken by a channel used in the layout not existing in a particular user's set-up. Hence, I've mostly seen people change the set via config applied to all users on start-up. I can envisage someone coming up with a use case for adding to the set of user channels for some specific purpose (although such use cases are often served by app channels - which are not administrated by the user but handled by the apps instead), but changing or removing channels I mostly see as creating challenges for little gain.

kriswest (dismissed):
Still LGTM, @robmoffat does it test up ok on your end?

kriswest:
Feel free to set another release branch up and I'll review for ya - perhaps we should change the package.yml workflow back to running on merge into main at the same time...

That said, don't forget that you could just `npm pack --workspace` and then `npm install <package-name>.tar.gz` in the project you need to test it in, for a local test. However, it would be good to have a beta out with this change.

robmoffat:
@novavi could you take a look and approve this please?

robmoffat:
Can someone from the maintainer team review this please?

kriswest:
> Can someone from the maintainer team review this please?

@julianna-ciq I know you guys will need this to work...

mistryvinay (dismissed):
Approve these additions.

mistryvinay (approved):
LGTM

Issue #1585, opened 2025-04-25
Erroneous Property / Structure for Channel Events


robmoffat:
## Minor Issue
Please use minor issues for discrepancies or errors found in the spec or supporting documentation, examples, and other materials.  If you wish to propose a new feature or a revision to an existing feature, please use the *Enhancement Request* template.

### Area of Issue
 - [ ] App Directory
 - [x] API
 - [ ] Context Data
 - [ ] Intents
 - [ ] Desktop Agent Bridging
 - [ ] Use Cases
 - [ ] Other

This is a bit confusing:

```
// Events.ts

/**
 * Type defining the format of event `userChannelChanged` objects
 */
export interface FDC3ChannelChangedEvent extends FDC3Event {
  readonly type: 'userChannelChanged';
  readonly details: {
    currentChannelId: string | null;
  };
}
```

```
// BrowserTypes.ts

export interface ChannelChangedEvent {
  /**
   * Metadata for messages sent by a Desktop Agent to an app notifying it of an event.
   */
  meta: BroadcastEventMeta;
  /**
   * The message payload contains details of the event that the app is being notified about.
   */
  payload: ChannelChangedEventPayload;
  /**
   * Identifies the type of the message and it is typically set to the FDC3 function name that
   * the message relates to, e.g. 'findIntent', with 'Response' appended.
   */
  type: 'channelChangedEvent';
}

/**
 * The message payload contains details of the event that the app is being notified about.
 */
export interface ChannelChangedEventPayload {
  /**
   * The Id of the channel that the app was added to or `null` if it was removed from a
   * channel.
   */
  newChannelId: null | string;
}
```

the problem is that one is taking a newChannelId and the other is currentChannelId. 
... and the event type changes too

ok, this is a weird issue, definitely an error:

```
/**
 * Type defining the format of event objects that may be received
 * via the FDC3 API's `addEventListener` function.
 */
export interface FDC3Event extends ApiEvent {
  readonly string: FDC3EventTypes;
  readonly details: any;
}

/**
 * Type defining the format of event `userChannelChanged` objects
 */
export interface FDC3ChannelChangedEvent extends FDC3Event {
  readonly type: 'userChannelChanged';
  readonly details: {
    currentChannelId: string | null;
  };
}

```
There's a property called "string" on the FDC3Event!

[3:01](https://finos-lf.slack.com/archives/D037QSHKLCW/p1745589660879449)
I'll raise an issue

Message Kris West

PR #1586, opened 2025-04-25
Fdc3 new conformance 2.2

robmoffat:
Added two new conformance tests and got FDC3 Demo to pass the new manual test.
Raised #1585


.husky/pre-commit
```diff
@@ -1,2 +1,2 @@
 #!/bin/sh
-npx lint-staged
+# npx lint-staged # Why do we need this?  We have github actions for this and this breaks my IDE.
```
Roaders, line 2: Was there ever a discussion about using husky in the first place?

We have never used it as not all clients work with husky - the most obvious being Visual studio code. If you can't rely on the hooks running the IMHO it's better to remove husky completely and rely on GitHub actions.

packages/fdc3-agent-proxy/src/channels/DefaultChannelSupport.ts
```diff
@@ -54,23 +54,24 @@ export class DefaultChannelSupport implements ChannelSupport {
...
+      const cce = e as FDC3ChannelChangedEvent;
+      const currentChannelId = cce.details.currentChannelId;
+      Logger.debug('Desktop Agent reports channel changed: ', currentChannelId);
 
       let theChannel: Channel | null = null;
 
       // if theres a newChannelId, retrieve details of the channel
-      if (cce.newChannelId) {
-        theChannel = this.userChannels.find(uc => uc.id == cce.newChannelId) ?? null;
+      if (currentChannelId) {
```
Roaders, line 64: it is better to do a more explicit comparison

```ts
if (currentChannelId != null) {
```

in this case if the channel id is `0` or an empty string then the if statement will not fire.

I've had some very long frustrating debug session caused by JavaScript type coercion "feature" where it has not fired an if for a value of `0`

Roaders (approved):

Roaders:
Is there a github actions to ensure that prettier has been run? WE should run lint (without fixes) to ensure that no code gets into the repo that does not pass linting.

As I said you can't rely on husky at all. It does not run from visual studio code when you do a commit (at least it didn't, I am not aware of any fixes for this).

I assume that most people contributing would use VS code?

We should probably add a recommended extension setting so that people install the vscode prettier plugin so that changes are made whenever you save a file.

kriswest:
If the huskey precommit hook isn't working using that shell script, then we can instead configure it in the package.json file by adding:
 
```json
"husky": {
  "hooks": {
    "pre-commit": "npx lint-staged"
  }
}
```

That might provide better cross-platform support. Original set-up instructions I used to (restore prettier use in the repo after refactor) came from prettier: https://prettier.io/docs/install#git-hooks

robmoffat:
This is passing for me now.

robmoffat:
@kriswest and @Roaders I've just fixed the build here - lots of linting issues.  Conformance tests are all passing for the `demo` DA.  I'm away for the next couple of weeks - it would be good if you could pick this up now and try to get it merged with the other maintainers if possible.

thanks!

robmoffat:
I'll leave you to figure out what to do about husky / pre-commit checks.

.husky/pre-commit
julianna-ciq: To elaborate on what Kris said - it may looks like this step is just running a linter to *check* for formatting errors. In fact, it is running a linter to *fix* formatting errors, and then *check those fixes in*.

If you only had a GHA to check for linting issues, you'd be constantly annoyed by failing linting that you'd have to fix and re-publish. If you run linting regularly , it will frequently complain about unrelated files that aren't the ones you're trying to commit.

lint-staged fixes the easy problems in the background, keeping the code clean with minimal annoyance to the developer.

I would highly recommend keeping lint-staged in place.

julianna-ciq:
I have forwarded this PR to our engineers. I'll follow up with you when I have new information, @robmoffat .

Roaders:
I had a go at running this against our implementation and it did seem to correctly run the tests up to a point. As last time we tried running we got to a certain point and then the tests stopped running. For this test run it got to 93% and got no further. The last test to run was:

> "after each" hook: afterEach for "(2.0-ACBasicUsage1) Should receive context when app a adds a listener and app B broadcasts to the same app channel"

However I also tried to run against the reference implementation and I got lots of failures and an imncomplete test there as well:

image

I got quite a few:

> Error: App didn't return close context within .5 secs

but also other timeouts of 5000ms, 20000ms and various assertion errors and AppTimeout. I don't think that there is a report that I can copy and paste to show all the results.

robmoffat:
this is a long shot, but is popup blocking turned off?

package.json
```diff
@@ -30,7 +30,7 @@
     "toolbox/fdc3-for-web/reference-ui",
     "toolbox/fdc3-for-web/demo",
     "toolbox/fdc3-workbench",
-    "toolbox/fdc3-conformnance"
```
robmoffat: @kriswest @openfin-johans this typo is why the deploy isn't on the fdc3.finos.org.   We need to merge this PR too

kriswest: fixed in #1650, along with linting and other issues. If you could review and merge that @robmoffat all should be well again when you rebase other PRs

robmoffat: done

robmoffat:
Hi I ran this not long ago and it worked for me, although there are known race conditions that cause it to fail sometimes. @Roaders are you running this on a PC?  I've observed the race conclude differently there than on the Mac.  

In my other PR, you can see I've been messing about with the timings for this very reason:  

https://github.com/finos/FDC3/pull/1696/files#diff-c22f660199ac437ce4a57e10356b4739c5ae21c9abd6536f44e854f85b46b378

This is the server-context-refactor PR.  Obviously, a key test for me there is that conformance still passes after the refactor!

toolbox/fdc3-conformance/src/test/manual/fdc3.manual.ts
```diff
@@ -1,25 +1,25 @@
...
-declare let fdc3: DesktopAgent;
-
 /**
  * Details on the mock apps used in these tests can be found in /mock/README.md
  */
 
-export let fdc3ResolveAmbiguousIntentTarget_2_0 = () =>
-  describe('ResolveAmbiguousIntentTarget_2.0', () => {
+export const fdc3ResolveAmbiguousIntentTarget_2_0 = () =>
+  describe('2.0-ResolveAmbiguousIntentTarget', async () => {
```
openfin-johans, line 15: Maybe I am just confused but why the different approach to adding 2.0? Sometimes after, sometimes before?

toolbox/fdc3-conformance/src/test/support/intent-support-2.0.ts
```diff
@@ -218,7 +225,7 @@ export class RaiseIntentControl2_0 {
       streamedNumberStart += 1;
     });
 
-    timeout = window.setTimeout(() => {
+    const timeout = window.setTimeout(() => {
```
openfin-johans, line 228: Does this actually add a delay or time?

openfin-johans (commented):
Looks good to me - just some questions for my interest...

toolbox/fdc3-conformance/src/test/support/intent-support-2.0.ts
kriswest: If the timer isn't cancelled (on receiving the expected contexts) within `constants.WaitTime`, then it fails the test. It gets cleared above on line 222

openfin-johans: Thanks @kriswest - I hadn't expanded to see the whole function...

toolbox/fdc3-conformance/src/test/manual/fdc3.manual.ts
robmoffat: the text string is what appears on the screen.  This is consistent.  The constant is used in the code and is also consistent but just with the other constants :)

.vscode/launch.json
```diff
@@ -0,0 +1,17 @@
+{
```
kriswest: Do we need this? Perhaps .gitignore it if not

packages/fdc3-agent-proxy/src/channels/ChannelSupport.ts
```diff
@@ -8,5 +8,9 @@ export interface ChannelSupport {
   leaveUserChannel(): Promise<void>;
   joinUserChannel(id: string): Promise<void>;
   addContextListener(handler: ContextHandler, type: string | null): Promise<Listener>;
-  addChannelChangedEventHandler(handler: EventHandler): Promise<Listener>;
+
+  /**
+   * TODO: Move out of ChannelSupport when there are other, non-channel related event types to support.
+   */
+  addEventListener(handler: EventHandler, type: FDC3EventTypes | null): Promise<Listener>;
```
kriswest, line 15: While I understand the comment (this is here because its being user for UserChannelChanged, which is actually registered on the DA rather than a channel), I think this is actually the right place for this to go/it will be needed here later - there is an event coming in 3.0 thats relevant to all channels (userChannelChanged does not apply, but contextCleared will in future). 

It could have a more precise type. The current type will cause issues with overriding it in PrivateChannelSupport, so this might need to be ApiEvent for now... Otherwise we'll need a new type that filters for events that apply to channels only.
```suggestion
  /**
   * TODO: Move handling for userChannelChanged out of ChannelSupport and update type to filter to channel events only.
   */
  addEventListener(handler: EventHandler, type: FDC3EventTypes | null): Promise<Listener>;
```

packages/fdc3-agent-proxy/src/listeners/DesktopAgentEventListener.ts
```diff
@@ -0,0 +1,93 @@
...
+  AgentEventMessage,
+  ChannelChangedEvent,
+} from '@finos/fdc3-schema/generated/api/BrowserTypes';
+import { Messaging } from '../Messaging';
+import { AbstractListener } from './AbstractListener';
+
+function wrapHandler(handler: EventHandler): (msg: AgentEventMessage) => void {
+  return (m: AgentEventMessage) => {
+    if (m.type === 'channelChangedEvent') {
+      const cm: ChannelChangedEvent = m;
```
kriswest: Why not just use `m`?

packages/fdc3-agent-proxy/src/listeners/DesktopAgentEventListener.ts
```diff
@@ -0,0 +1,93 @@
...
+      const restructured: FDC3ChannelChangedEvent = {
+        type: 'userChannelChanged',
+        details: {
+          currentChannelId: cm.payload.newChannelId,
+        },
+        string: 'userChannelChanged', // ISSUE:  #1585
```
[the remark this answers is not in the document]
kriswest: we can correct the type and lose this erroneous field

kriswest:
```suggestion
      const restructured: FDC3ChannelChangedEvent = {
        type: 'userChannelChanged',
        details: {
          //backwards compatibility, see issue #1585
          currentChannelId: cm.payload.newChannelId ?? cm.payload.currentChannelId 
        }
```

kriswest:
```suggestion
      const restructured: FDC3ChannelChangedEvent = {
        type: 'userChannelChanged',
        details: {
          //backwards compatibility, see issue #1585
          currentChannelId: m.payload.newChannelId ?? m.payload.currentChannelId 
        }
```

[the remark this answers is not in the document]
kriswest: we can correct the type and lose this erroneous field

packages/fdc3-agent-proxy/src/listeners/DesktopAgentEventListener.ts
```diff
@@ -0,0 +1,93 @@
...
+      };
+
+      handler(restructured);
+    } else {
+      // currently unused but meeting issue #1585
```
kriswest, line 24: will be used shortly for contextCleared
```suggestion
```

packages/fdc3-agent-proxy/test/features/user-channel-sync.feature
```diff
@@ -13,6 +13,7 @@ Feature: Updating User Channel State
     Given "resultHandler" pipes context to "contexts"
     And I call "{api}" with "addContextListener" with parameters "fdc3.instrument" and "{resultHandler}"
     When I call "{api}" with "joinUserChannel" with parameter "one"
+    And we wait for a period of "600" ms
```
kriswest: blergh, but inevitable. The standard doesn't require that the current context is received before this call returns, only that it is required, only that "it will automatically receive the current context for that channel." (see https://fdc3.finos.org/docs/api/spec#joining-user-channels).

Hence, not an implementation flaw.

Hence, I'd consider bumping to 1 second

```suggestion
    And we wait for a period of "1000" ms
```

packages/fdc3-agent-proxy/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@finos/fdc3-agent-proxy",
-  "version": "2.2.0",
+  "version": "2.2.1-beta.4",
```
kriswest: these can all be updated to 2.2.2-beta.1

kriswest: (release flow changed so its safe to do so - there are a few other PRs to merge until we go to 2.2.2-rc.1)

kriswest: (release flow changed so its safe to do so - there are a few other PRs to merge until we go to 2.2.2-rc.1)

toolbox/fdc3-conformance/src/test/advanced/fdc3.findIntent.ts
```diff
@@ -59,7 +59,7 @@ export default () =>
       );
     });
 
-    it("(IntentAppDMultiple2) Should find intent 'sharedTestingIntent2' belonging to multiple apps (intent-a & intent-b) filtered by specific context 'testContextY'", async () => {
+    it("(2.0-IntentAppDMultiple2) Should find intent 'sharedTestingIntent2' belonging to multiple apps (intent-a & intent-b) filtered by specific context 'testContextY'", async () => {
```
kriswest, line 62: will be removed again in a subsequent PR

toolbox/fdc3-conformance/src/test/basic/fdc3.basic.ts
```diff
@@ -7,11 +7,20 @@ import { closeMockAppWindow } from '../fdc3-2_0-utils';
 import { assert, expect } from 'chai';
 import { handleFail } from '../../utils';
 
-let basicCL1 = (fdc3: DesktopAgent, documentation: string) => {
+const getAgent2_2 = (fdc3: DesktopAgent, documentation: string) => {
+  it('(GetAgentAPI) Method is callable', async () => {
+    const info = await fdc3.getInfo();
+    assert.isTrue(info.fdc3Version.startsWith('2.'), documentation);
```
kriswest, line 13: I think you should check that the version is 2.2 (it is necessary for conformance for the right version number to be reported) and that that value should be drawn from a constant that we can easily update in subsequent releases.

toolbox/fdc3-conformance/src/test/manual/fdc3.manual.ts
kriswest: I've got a PR ready to raise to remove *all* the 2.0s, extending on from Rob's that removes them from the names - doing so will simplify maintenance in future versions. We can apply after this is in.

toolbox/fdc3-conformance/src/test/manual/fdc3.manual.ts
```diff
@@ -100,3 +108,25 @@ export let fdc3ResolveAmbiguousContextTargetMultiInstance_2_0 = () =>
...
+      try {
+        const agent = await getAgent();
+        agent.addEventListener('userChannelChanged', event => {
+          const changedEvent: FDC3ChannelChangedEvent = event as FDC3ChannelChangedEvent;
+          const currentChannel = changedEvent.details.currentChannelId;
+          console.log('User channel changed', event, currentChannel);
+          channels.push(currentChannel);
+        });
+
+        await wait(8000);
```
kriswest, line 125: This could come from a constant - I added detail on it to the test name above (which could also use the constant)...
```suggestion
        await wait(10000);
```

toolbox/fdc3-conformance/src/test/manual/fdc3.manual.ts
```diff
@@ -100,3 +108,25 @@ export let fdc3ResolveAmbiguousContextTargetMultiInstance_2_0 = () =>
       }
     });
   });
+
+export const fdc3ChannelChangedEvent_2_2 = () =>
+  describe('2.2-ChannelChangedEvent', () => {
+    it('(2.2-ChannelChangedEvent) Should receive an event when the user changes channel.  This is a manual test, please change the channel a few times in your browser to get this to pass.', async () => {
```
kriswest, line 114:
```suggestion
    it('(2.2-ChannelChangedEvent) Should receive an event when the user changes channel.  This is a manual test, please change the channel a few times in the next 10 seconds in your browser to get this to pass.', async () => {
```

toolbox/fdc3-conformance/src/test/manual/fdc3.manual.ts
```diff
@@ -100,3 +108,25 @@ export let fdc3ResolveAmbiguousContextTargetMultiInstance_2_0 = () =>
...
+          const currentChannel = changedEvent.details.currentChannelId;
+          console.log('User channel changed', event, currentChannel);
+          channels.push(currentChannel);
+        });
+
+        await wait(8000);
+        const uniqueChannels = new Set(channels);
+        expect(uniqueChannels.size).to.be.greaterThan(0);
+      } catch (ex) {
+        handleFail(`Didn't get any channel change events: ${JSON.stringify(channels)}`, ex);
```
kriswest, line 129:
```suggestion
        handleFail(`Didn't get any channel change events (ensure that you manually change the user channel in your UI): ${JSON.stringify(channels)}`, ex);
```

toolbox/fdc3-conformance/static/directories/preview-conformance.v2.json
```diff
@@ -0,0 +1,960 @@
+{
```
kriswest: Remove before merge as it'll break when the preview is deleted!

toolbox/fdc3-for-web/demo/README.md
```diff
@@ -94,4 +94,6 @@ As shown in the screenshot above, there are 7 demo applications provided by Demo
 
 - **Intent Result**: On launch, raises a `ViewQuote` intent with an `fdc3.instrument` context.  This is useful for testing out intent resolution without having to bother with the FDC3 Workbench.
 
-- **Intent Result 2**: On launch, raises a `ViewNews` intent with an `fdc3.instrument` context.    Tip: Try changing the intent resolver in the UI to the default one and see what happens.
\ No newline at end of file
+- **Intent Result 2**: On launch, raises a `ViewNews` intent with an `fdc3.instrument` context.    Tip: Try changing the intent resolver in the UI to the default one and see what happens.
+
+- **FDC3 Conformance 2.0**: This app is used to run the FDC3 2.0 conformance tests.
```
kriswest: Can you add a note about starting the conformance project, as it needs to be running locally!
```suggestion
- **FDC3 Conformance 2.0**: This app is used to run the FDC3 2.0 conformance tests, which should be running locally. See [/toolbox/fdc3-conformance/README.md#running-the-conformance-tests](../../fdc3-conformance/README.md#running-the-conformance-tests#running-the-conformance-tests).  
```

toolbox/fdc3-for-web/fdc3-web-impl/src/handlers/BroadcastHandler.ts
```diff
@@ -201,13 +226,49 @@ export class BroadcastHandler implements MessageHandler {
         // handling state synchronization of channels
         case 'getCurrentContextRequest':
           return this.handleGetCurrentContextRequest(msg as GetCurrentContextRequest, sc, from);
+
+        // TODO: move this out when we no longer handle just channel-based events.
```
kriswest: This will remain for the contextClearedEvent coming in 3.0. 
```suggestion
        // Handle channel events
```

website/docs/api/conformance/Basic-Tests.md
```diff
@@ -17,7 +17,7 @@ _These are some basic sanity tests implemented in the FDC3 Conformance Framework
 | App | Step            | Description                                              |
 |-----|-----------------|----------------------------------------------------------|
 | A   | `getAgent`      | App A calls `getAgent` and waits for the promise to resolve to a `DesktopAgent` instance. |
-| A   | `getInfo`       | App A can call the `getInfo()` method on the `DesktopAgent` instance to get the `ImplementationMetadata` object. <br /> Check that fdc3Version is set to 2.2.  <br />Check that provider and providerVersion are populated. |
+| A   | `getInfo`       | App A can call the `getInfo()` method on the `DesktopAgent` instance to get the `ImplementationMetadata` object. <br /> Check that fdc3Version is set to 2.x.  <br />Check that provider and providerVersion are populated. |
```
kriswest, line 20: This should remain 2.2 or (better) the correct FDC3 version - see comment in the implementation about checking it is actually 2.2 (as we decided to use tags of teh codebase for deploying test versions in future)

```suggestion
| A   | `getInfo`       | App A can call the `getInfo()` method on the `DesktopAgent` instance to get the `ImplementationMetadata` object. <br /> Check that fdc3Version is set to the correct FDC3 spec version.  <br />Check that provider and providerVersion are populated. |
```

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## [Unreleased]
 * Enhanced method binding for FDC3 API objects to support destructuring. All public methods of `Channel`, `PrivateChannel`, and `IntentResolution` objects are now properly bound to their instances using `.bind(this)` in their constructors. ([#1645](https://github.com/finos/FDC3/issues/1645))
 
+* Increased FDC3 Conformance Test WindowCloseWaitTime to 2000 (from 1000)
```
kriswest:
```suggestion
* Enhanced method binding for FDC3 API objects to support destructuring. All public methods of `Channel`, `PrivateChannel`, and `IntentResolution` objects are now properly bound to their instances using `.bind(this)` in their constructors. ([#1645](https://github.com/finos/FDC3/issues/1645))
* Increased FDC3 Conformance Test WindowCloseWaitTime to 2000 (from 1000). ([#1586](https://github.com/finos/FDC3/issues/1586))
```

kriswest: Both of these changelogs need moving into the Fixed section please

website/docs/api/conformance/Basic-Tests.md
robmoffat: Screenshot 2026-01-30 at 16 17 03

We are missing code coverage for the context clearing in this PR. Is it added elsewhere?

robmoffat: The conformance test actually checks for >=2 .

```
  expect(implMetadata, `ImplementationMetadata did not have property fdc3Version${getInfoDocs}`).to.have.property(
      'fdc3Version'
    );
    expect(parseFloat(implMetadata.fdc3Version)).to.be.greaterThanOrEqual(2);
    expect(implMetadata, `ImplementationMetadata did not have property provider${getInfoDocs}`).to.have.property(
      'provider'
    );
```

Remember, these are 2.0 conformance tests, not 2.2 ones

kriswest: these *were* the 2.0 ones, but are becoming the 2.2's right? I thought/think the plan is to just maintain the tests against the current version and have you deploy older release tags for past test versions. Hence, this PR can focus on 2.2 tests.

kriswest: > We are missing code coverage for the context clearing in this PR. Is it added elsewhere?

It may be needed still... Context clearing is supposed to be going out in 3.0 - we probably need to get merged and then on the release branch rip out anything related to context clearing... or agree to leave it in as relatively harmless and @experimental. 

The 3.0 conformance tests MUST include specific tests on context clearing if its a MUST in the 3.0 standard (which I think it should be).

kriswest: Finally, I have some additional changes queued up to rebase onto this that affect the events code. Was planning to raise that after we get this in and I rebase/fix up any conflicts.

kriswest:
> Also we need to merge in #1726 before we merge this one.

@robmoffat how do you feel about merging this *before* #1726? I've got a further set of changes to apply to #1726 where I've removed every 2.0 etc. (not just from the names, but filenames, test names etc.) to align fully with the 'just maintain the conformance tests against the current version and use tags for older versions' plan.  It touches a lot of files and it might be easiest to make sure this tests up, then get the renaming PR in next?

packages/fdc3-schema/generated/api/BrowserTypes.ts
```diff
@@ -1372,11 +1372,16 @@ export interface ChannelChangedEvent {
...
+   * Deprecated - allowed for backwards compatibility.  The Id of the channel that the app was
+   * added to or `null` if it was removed from a channel.
+   */
+  newChannelId?: null | string;
   /**
    * The Id of the channel that the app was added to or `null` if it was removed from a
    * channel.
    */
-  newChannelId: null | string;
+  currentChannelId?: null | string;
```
kriswest, line 1384: @robmoffat this is probably the thing that broke the typescript build - quicktype is making the two possible fields both optional because it can't express in TypeScript that one or other must be set (as JSONSchema can). Will need handling in the code

packages/fdc3-standard/src/api/Events.ts
```diff
@@ -32,7 +32,6 @@ export type FDC3EventTypes = 'userChannelChanged';
  * via the FDC3 API's `addEventListener` function.
  */
 export interface FDC3Event extends ApiEvent {
-  readonly string: FDC3EventTypes;
   readonly details: any;
```
kriswest, line 35: I think we do want the `type` field here to narrow down the types from §§string§ as used in APIEvent:

```suggestion
  readonly type: FDC3EventTypes
  readonly details: any;
```

However, we do need to watch out and make sure we use APIEvent as the type in `Channel`'s `addEventListener(event: APIEvent`) function as that gets extended in PrivateChannel to include the PrivateChannel events - so the `PrivateChannel` version becomes `addEventListener(event: FDC3Event | PrivateChannelEvent)`.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 * Enhanced method binding for FDC3 API objects to support destructuring. All public methods of `Channel`, `PrivateChannel`, and `IntentResolution` objects are now properly bound to their instances using `.bind(this)` in their constructors. ([#1645](https://github.com/finos/FDC3/issues/1645))
+* Increased FDC3 Conformance Test WindowCloseWaitTime to 2000 (from 1000). ([#1586](https://github.com/finos/FDC3/issues/1586))
```
kriswest: need to move these down into 'Changed'

I realise you didn't add line 8 but if you could move it at the same time as 9 that'd be grand.

CHANGELOG.md
```diff
@@ -21,6 +22,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Changed
 
+* Added Conformance tests for FDC3 2.2 ([#1586](https://github.com/finos/FDC3/pull/1586))
```
kriswest: Move to the Added section?

packages/fdc3-standard/src/api/Events.ts
kriswest: That `type` field is really the only reason for this type to exist!

robmoffat:
Still looking good after your PR @kriswest 

Screenshot 2026-02-03 at 13 31 42

toolbox/fdc3-conformance/src/test/basic/fdc3.basic.ts
kriswest: I still think we should check specific version (you need to report the matching version number of the test set and to use the test set for the right version...). We can pick up the version number from the fdc3-standard - but we can pick that up in a subsequent PR.

CHANGELOG.md
```diff
@@ -4,9 +4,6 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
-## [Unreleased]
```
kriswest, line 7: Gah, we need to keep the `## [Unreleased]` section heading @robmoffat   - otherwise, thank you for the cleanup

kriswest (approved):
LGTM

PR #1590, opened 2025-05-09
(WIP) implement addIntentListenerWithContext

Roaders:
## Describe your change

Adds a new function `addIntentWithContextListener` to the `DesktopAgent` interfaces and implements it within `DesktopAgentProxy`

### Related Issue
resolves #1545

## Contributor License Agreement

- [x] I acknowledge that a contributor license agreement is required and that I have one in place or will seek to put one in place ASAP.

## Review Checklist

- [x] **Issue**: If a change was made to the FDC3 Standard, was an issue linked above?
- [x] **CHANGELOG**: Is a *CHANGELOG.md* entry included?
- [x] **API changes**: Does this PR include changes to any of the FDC3 APIs (`DesktopAgent`, `Channel`, `PrivateChannel`, `Listener`, `Bridging`)?
  - [ ] **Docs & Sources**: If yes, were both documentation (/docs) and sources updated?<br/>
        *JSDoc comments on interfaces and types should be matched to the main documentation in /docs*
  - [ ] **Conformance tests**: If yes, are conformance test definitions (/toolbox/fdc3-conformance) still correct and complete?<br/>
        *Conformance test definitions should cover all **required** aspects of an FDC3 Desktop Agent implementation, which are usually marked with a **MUST** keyword, and  optional features (**SHOULD** or **MAY**) where the format of those features is defined*
  - [ ] **Schemas**: If yes, were changes applied to the Bridging and FDC3 for Web protocol schemas?<br/>
        *The Web Connection protocol and Desktop Agent Communication Protocol schemas must be able to support all necessary aspects of the Desktop Agent API, while Bridging must support those aspects necessary for Desktop Agents to communicate with each other*
    - [ ] If yes, was code generation (`npm run build`) run and the results checked in?<br/>
        *Generated code will be found at `/src/api/BrowserTypes.ts` and/or `/src/bridging/BridgingTypes.ts`*
- [ ] **Context types**: Were new Context type schemas created or modified in this PR?
  - [ ] Were the [field type conventions](https://fdc3.finos.org/docs/context/spec#field-type-conventions) adhered to?
  - [ ] Was the `BaseContext` schema applied via `allOf` (as it is in existing types)?
  - [ ] Was a `title` and `description` provided for all properties defined in the schema?
  - [ ] Was at least one example provided?
  - [ ] Was code generation (`npm run build`) run and the results checked in?<br/>
        *Generated code will be found at `/src/context/ContextTypes.ts`*
[...]

PR #1594, opened 2025-05-17
Add intent with context listener

### LINKED ISSUE #1545: Support specifying supported context when registering an intent listener
### Use Case:
It is possible for an app that is not listed in an app directory to register an intent listener with the DesktopAgent using the `addIntentListener` function. This means that the app (that is not listed in an app directory) will be included in the app resolver UI when an intent is raised. What is not supported is to specify the contexts that the app supports. This is supported by the app directory and it seems sensible that the same level of detail should be available to apps that register intent listeners rather than relying on the app directory.

Without the ability to specify the supported contexts the user could experience apparently broken applications. If they raise an intent with an unsupported context the app will still be listed in the app resolver UI but when the intent is sent to that app it will not know what to do with it and will either do nothing (which will appear broken) or will throw an error.

We should add an additional parameter to the function to specify what context is supported:

```ts
addIntentListener(intent: string, context: Context, handler: IntentHandler): Promise<Listener>;
```

We could optionally support multiple contexts:

```ts
addIntentListener(intent: string, context: Context | Context[], handler: IntentHandler): Promise<Listener>;
```

but this could be handled by just calling the first version multiple times.

As written above this would be a breaking change which is undesirable. It could be made non-breaking by supporting it with an overload of the existing function but I understand this is hard to implement in non-javascript languages. We could also add it as an optional argument at the end:

```ts
addIntentListener(intent: string, handler: IntentHandler, context?: Context): Promise<Listener>;
```

I don't think that this parameter order makes as much sense but it would mean we could make the change non-breaking.

Roaders:
## Describe your change

Adds a new function `addIntentWithContextListener` to the `DesktopAgent` interfaces and implements it within `DesktopAgentProxy`

### Related Issue
resolves #1545
recreating #1590 - to be closed

## Review Checklist

---

### DIFF
--- a/packages/fdc3-agent-proxy/src/intents/DefaultIntentSupport.ts
+++ b/packages/fdc3-agent-proxy/src/intents/DefaultIntentSupport.ts
@@ -272,7 +272,23 @@ export class DefaultIntentSupport implements IntentSupport {
   }
 
   async addIntentListener(intent: string, handler: IntentHandler): Promise<Listener> {
-    const out = new DefaultIntentListener(this.messaging, intent, handler, this.messageExchangeTimeout);
+    const out = new DefaultIntentListener(this.messaging, intent, undefined, handler, this.messageExchangeTimeout);
+    await out.register();
+    return out;
+  }
+
+  async addIntentListenerWithContext(
+    intent: string,
+    contextType: string | string[],
+    handler: IntentHandler
+  ): Promise<Listener> {
+    const out = new DefaultIntentListener(
+      this.messaging,
+      intent,
+      Array.isArray(contextType) ? contextType : [contextType],
+      handler,
+      this.messageExchangeTimeout
+    );
     await out.register();
     return out;
   }
--- a/packages/fdc3-agent-proxy/src/intents/DefaultIntentSupport.ts
+++ b/packages/fdc3-agent-proxy/src/intents/DefaultIntentSupport.ts
@@ -74,6 +74,7 @@ export class DefaultIntentSupport implements IntentSupport {
   readonly intentResolver: IntentResolver;
   readonly messageExchangeTimeout: number;
   readonly appLaunchTimeout: number;
+  private readonly intentListeners: DefaultIntentListener[] = [];
 
   constructor(
     messaging: Messaging,
@@ -272,9 +273,7 @@ export class DefaultIntentSupport implements IntentSupport {
   }
 
   async addIntentListener(intent: string, handler: IntentHandler): Promise<Listener> {
-    const out = new DefaultIntentListener(this.messaging, intent, undefined, handler, this.messageExchangeTimeout);
-    await out.register();
-    return out;
+    return this.registerIntentListener(intent, undefined, handler);
   }
 
   async addIntentListenerWithContext(
@@ -282,14 +281,52 @@ export class DefaultIntentSupport implements IntentSupport {
     contextType: string | string[],
     handler: IntentHandler
   ): Promise<Listener> {
+    return this.registerIntentListener(intent, Array.isArray(contextType) ? contextType : [contextType], handler);
+  }
+
+  private async registerIntentListener(
+    intent: string,
+    contextTypes: string[] | undefined,
+    handler: IntentHandler
+  ): Promise<Listener> {
+    this.throwIfConflicting(intent, contextTypes);
+
     const out = new DefaultIntentListener(
       this.messaging,
       intent,
-      Array.isArray(contextType) ? contextType : [contextType],
+      contextTypes,
       handler,
-      this.messageExchangeTimeout
+      this.messageExchangeTimeout,
+      () => {
+        const index = this.intentListeners.indexOf(out);
+        if (index !== -1) {
+          this.intentListeners.splice(index, 1);
+        }
+      }
     );
     await out.register();
+    this.intentListeners.push(out);
     return out;
+  }
+
+  /**
+   * A new intent listener conflicts with an existing one registered for the same intent when either
+   * listener is unfiltered (no context types, so it handles all contexts) or their declared context
+   * types overlap.
+   */
+  private throwIfConflicting(intent: string, contextTypes: string[] | undefined): void {
+    const conflict = this.intentListeners.some(existing => {
+      if (existing.intent !== intent) {
+        return false;
+      }
+      if (existing.contextTypes == null || contextTypes == null) {
+        return true;
+      }
+      return existing.contextTypes.some(ct => contextTypes.includes(ct));
+    });
+
+    if (conflict) {
+      throw new Error(ResolveError.IntentListenerConflict);
+    }
   }
 }
--- a/packages/fdc3-schema/schemas/api/addIntentListenerRequest.schema.json
+++ b/packages/fdc3-schema/schemas/api/addIntentListenerRequest.schema.json
@@ -1,46 +1,54 @@
 {
-	"$schema": "http://json-schema.org/draft-07/schema#",
-	"$id": "https://fdc3.finos.org/schemas/next/api/addIntentListenerRequest.schema.json",
-	"type": "object",
-	"title": "AddIntentListener Request",
-	"description": "A request to add an Intent listener for a specified intent type.",
-	"allOf": [
-		{
-			"$ref": "appRequest.schema.json"
-		},
-		{
-			"type": "object",
-			"properties": {
-				"type": {
-					"$ref": "#/$defs/AddIntentListenerRequestType"
-				},
-				"payload": {
-					"$ref": "#/$defs/AddIntentListenerRequestPayload"
-				},
-				"meta": true
-			},
-			"additionalProperties": false
-		}
-	],
-	"$defs": {
-		"AddIntentListenerRequestType": {
-			"title": "AddIntentListener Request Message Type",
-			"const": "addIntentListenerRequest"
-		},
-		"AddIntentListenerRequestPayload": {
-			"title": "AddIntentListener Request Payload",
-			"type": "object",
-			"properties": {
-				"intent": {
-					"title": "Intent name",
-					"description": "The name of the intent to listen for.",
-					"type": "string"
-				}
-			},
-			"additionalProperties": false,
-			"required": [
-				"intent"
-			]
-		}
-	}
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "$id": "https://fdc3.finos.org/schemas/next/api/addIntentListenerRequest.schema.json",
+    "type": "object",
+    "title": "AddIntentListener Request",
+    "description": "A request to add an Intent listener for a specified intent type.",
+    "allOf": [
+        {
+            "$ref": "appRequest.schema.json"
+        },
+        {
+            "type": "object",
+            "properties": {
+                "type": {
+                    "$ref": "#/$defs/AddIntentListenerRequestType"
+                },
+                "payload": {
+                    "$ref": "#/$defs/AddIntentListenerRequestPayload"
+                },
+                "meta": true
+            },
+            "additionalProperties": false
+        }
+    ],
+    "$defs": {
+        "AddIntentListenerRequestType": {
+            "title": "AddIntentListener Request Message Type",
+            "const": "addIntentListenerRequest"
+        },
+        "AddIntentListenerRequestPayload": {
+            "title": "AddIntentListener Request Payload",
+            "type": "object",
+            "properties": {
+                "intent": {
+                    "title": "Intent name",
+                    "description": "The name of the intent to listen for.",
+                    "type": "string"
+                },
+                "contextTypes": {
+                    "title": "Context types",
+                    "description": "The types of context to listen for.",
+                    "type": "array",
+                    "items": {
+                        "type": "string"
+                    }
+                }
+            },
+            "additionalProperties": false,
+            "required": [
+                "intent"
+            ]
+        }
+    }
 }
--- a/packages/fdc3-schema/schemas/api/addIntentListenerRequest.schema.json
+++ b/packages/fdc3-schema/schemas/api/addIntentListenerRequest.schema.json
@@ -1,54 +1,54 @@
 {
-    "$schema": "http://json-schema.org/draft-07/schema#",
-    "$id": "https://fdc3.finos.org/schemas/next/api/addIntentListenerRequest.schema.json",
-    "type": "object",
-    "title": "AddIntentListener Request",
-    "description": "A request to add an Intent listener for a specified intent type.",
-    "allOf": [
-        {
-            "$ref": "appRequest.schema.json"
-        },
-        {
-            "type": "object",
-            "properties": {
-                "type": {
-                    "$ref": "#/$defs/AddIntentListenerRequestType"
-                },
-                "payload": {
-                    "$ref": "#/$defs/AddIntentListenerRequestPayload"
-                },
-                "meta": true
-            },
-            "additionalProperties": false
-        }
-    ],
-    "$defs": {
-        "AddIntentListenerRequestType": {
-            "title": "AddIntentListener Request Message Type",
-            "const": "addIntentListenerRequest"
-        },
-        "AddIntentListenerRequestPayload": {
-            "title": "AddIntentListener Request Payload",
-            "type": "object",
-            "properties": {
-                "intent": {
-                    "title": "Intent name",
-                    "description": "The name of the intent to listen for.",
-                    "type": "string"
-                },
-                "contextTypes": {
-                    "title": "Context types",
-                    "description": "The types of context to listen for.",
-                    "type": "array",
-                    "items": {
-                        "type": "string"
-                    }
-                }
-            },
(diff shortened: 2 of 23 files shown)


packages/fdc3-schema/schemas/api/addIntentListenerRequest.schema.json
Roaders, line 1: sorry about the whitespace changes on here. Do we have anything in place to lint `json` changes to prevent this sort of change?

Roaders:
I think this covers most of the functionality but as far as I can see there is still a gap.

In the reference implementation I cannot see how an app that registers an intent with `addIntentListener` is displayed in the app resolution dialog. I assume that these apps come from the directory but as far as I can see the records in the directory are never updated based on incoming `addIntentListener` requests.

@robmoffat - can you give me some guidance on how this works?

Roaders:
Converted to draft as docs, unit tests and conformance tests still need to be updated

julianna-ciq (dismissed):
lgtm

packages/fdc3-standard/src/api/DesktopAgent.ts
```diff
@@ -351,6 +351,23 @@ export interface DesktopAgent {
    */
   addIntentListener(intent: Intent, handler: IntentHandler): Promise<Listener>;
 
+  /**
+   * Like `addIntentListener`, but allows filtering by one or more context types.
+   * The listener will only be triggered for the specified intent and context type(s).
```
julianna-ciq, line 356:
```suggestion
   * Adds a listener for incoming intents raised by other applications, via calls to `fdc3.raiseIntent` or `fdc3.raiseIntentForContext, but filters intents for one or more context types. See `addIntentListener` for details and restrictions for both usage and implementation.
```

packages/fdc3-schema/schemas/api/addIntentListenerRequest.schema.json
kriswest: Apparently not. The schema files should probably be being processed by prettier to do that. For some reason JSON files are being ignored:
https://github.com/finos/FDC3/blob/63eceaa9d1f1790c0c6d0f463b010965a187dd07/.prettierignore#L6

Feel free to change that and see what it does to the schema files - although I'd suggest doing so in a different PR

robmoffat:
Hi @Roaders,

To answer your question above, take a look at `IntentHandler.ts`:  it keeps a list of registrations 

```
  private registrations: ListenerRegistration[] = [];
```

when someone adds an intent listener.  The directory isn't changed itself - this is static based on what is loaded from the AppD endpoints.  

You're going to need to add some tests (see `find-intent.feature` and `raise-intent.feature`) for your new method on the server side and also in the proxy side (see `find-intents.feature` and `raise-intents.feature`)

HTH?

robmoffat:
> Can you point to the code that gets the list of application to display in the resolution dialog?

Yeah I can't see it either.   Seems like it only looks at what the appD record declares.  I'm not sure if this is correct behaviour or not, but It passes the conformance tests so if it isn't we are probably missing some tests.   

My feeling is this might be a gap in the spec:  the directory `interop` section should be consistent with the intent listeners that are registered otherwise this is going to cause either errors (when an app says it's going to register an intent listener but doesn't) or inconsistent behaviour (the app registers an unexpected intent listener, meaning the intents will get resolved differently depending on which apps are running).  @kriswest?

Roaders:
I thought about a conformance test but it would have to be a manual one as the user would have to select the registered app in the app resolver for the test to pass...
I guess...
Unless we make sure that only one app is registered I suppose.

robmoffat:
FWIW it does use registrations in findIntents.

kriswest:
@Roaders @robmoffat its upto the desktop agent to return the options for further resolution. Thats handled here in the proxy: https://github.com/finos/FDC3/blob/63eceaa9d1f1790c0c6d0f463b010965a187dd07/packages/fdc3-agent-proxy/src/intents/DefaultIntentSupport.ts#L159-L165

The DA has sent back app intents for all resolution options, which are then passed to the (injected) intent resolver. If the DA showed its won resolver then it doesn't return the appIntent as its already resolved, so this path is only for injected resolvers.

In the reference implementation intent resolution happens in this block (depending on whether and how a target was set): https://github.com/finos/FDC3/blob/63eceaa9d1f1790c0c6d0f463b010965a187dd07/toolbox/fdc3-for-web/fdc3-web-impl/src/handlers/IntentHandler.ts#L455-L474

The reference implementation doesn't currently support dynamic registration AFAIK (and should be updated to do so) as the resolution always relates to the app directory, that happens here : 

https://github.com/finos/FDC3/blob/63eceaa9d1f1790c0c6d0f463b010965a187dd07/toolbox/fdc3-for-web/fdc3-web-impl/src/handlers/IntentHandler.ts#L384-L385

This block is where it will resolve if no target was set and if it only found one option:
https://github.com/finos/FDC3/blob/63eceaa9d1f1790c0c6d0f463b010965a187dd07/toolbox/fdc3-for-web/fdc3-web-impl/src/handlers/IntentHandler.ts#L407-L423

Then finally here it returns options to show in the resolver if multiple options were found:
https://github.com/finos/FDC3/blob/63eceaa9d1f1790c0c6d0f463b010965a187dd07/toolbox/fdc3-for-web/fdc3-web-impl/src/handlers/IntentHandler.ts#L424-L439

As you guys are finding this hard to navigate perhaps we need to add some more comments to it? It does take a while to read and interpret that code...

kriswest:
> I thought about a conformance test but it would have to be a manual one as the user would have to select the registered app in the app resolver for the test to pass... I guess... Unless we make sure that only one app is registered I suppose.

A manual test is worth adding.

However, I did just have a random thought about injecting a resolver just for testing that can do the selection... but then not all DAs will use injection for the resolver so that might be too specific.

kriswest:
> Seems like it only looks at what the appD record declares. I'm not sure if this is correct behaviour or not, but It passes the conformance tests so if it isn't we are probably missing some tests.

Dynamic registration is currently optional (MAY) so we don't test for conformance on it.

> My feeling is this might be a gap in the spec: the directory interop section should be consistent with the intent listeners that are registered otherwise this is going to cause either errors (when an app says it's going to register an intent listener but doesn't) 

We have a specified error ([`ResolveError.IntentDeliveryFailed`](https://github.com/finos/FDC3/blob/63eceaa9d1f1790c0c6d0f463b010965a187dd07/packages/fdc3-standard/src/api/Errors.ts#L81-L82)) and minimum timeout allowed for registration in the spec (but no maximum timeout as one could not be agreed): 
https://github.com/finos/FDC3/blob/63eceaa9d1f1790c0c6d0f463b010965a187dd07/website/docs/api/spec.md?plain=1#L536

> or inconsistent behaviour (the app registers an unexpected intent listener, meaning the intents will get resolved differently depending on which apps are running). @kriswest?

...and thats optional support for dynamic registration. Its all covered in the spec (no gaps) but its also not entirely standardized as dynamic registration is optional (MAY).

Does that answer all questions raised in this thread?

kriswest:
@Roaders

> @kriswest I am a bit confused regarding adding a conformace test. You replied to my comment saying that we should add a conformance test to ensure that if an intent lisstener is registered that it appears in the resolution dialog but you also replied to Rob to say that dynamic registration is optional so we don't need a conformace test.
>
> It seems to me that we don't need a conformance test as this is an optional feature.

We do not desperately need a test if its not a required feature - although there is an argument that if an optional feature (MAY OR SHOULD) is present it MUST be conformant to the spec...

However, I made the comment in the context of this, perhaps, becoming a required feature. If we accept the addition of the new API call it should be a MUST (we don't have any other optional API calls AFAIK). Then we should consider whether dynamic registration should be required of a Desktop Agent or not. If it's NOT supported, then the standard is somewhat deficient on advice as to how to handle that (so I misspoke above - that is a gap!). There's no suitable error to return so most such implementations probably just accept the registration but never route anything to it/offer it during resolution. Is that the best solution? If so we should document that.

Hence, I think we have something that should be added to the spec whether we require dynamic registration or not.

> To summarise what's been said so far:
>
> - The refernce implementation does not support dynamic registration (so if we add an intent listener it will not appear in the app resolution dialog)
> - We should add support for dynamic registration to the reference implementation

💯

packages/fdc3-schema/schemas/api/addIntentListenerRequest.schema.json
Roaders: This is unfortunately not as easy as it would seem. Eslint does not support json files. I think that I am correct in saying that prettier is run via ESlint in this repo. Unfortunately I don't think that we can get prettier to handle json files when run via eslint.
In the past I have had to add a prettier call to my lint command to do this:

```
"lint": "prettier -c **/*.json && eslint"
```

Roaders:
The reference implementation now seems to correctly add dynamically registered apps to the app resolver.
I want to update the workbench to allow us to test `addIntentWithContextListener`. I did have some issues with the workbench when testing this and raise #1612

Roaders:
> What PR is that in?

This one! Sorry if that wasn't clear!

robmoffat:
Hi @Roaders,

I think this might be crossing the streams a bit:  that last part is a bug-fix that we should really have in the 2.2 branch, whereas the rest of this PR is for FDC3 2.3, I think

Roaders:
ok, raised #1613 as a seperate PR.

packages/fdc3-schema/schemas/api/addIntentListenerRequest.schema.json
kriswest: No I think its lint-staged running prettier on all projects and doing so on commit:

https://github.com/finos/FDC3/blob/d4a261a043d994670130ecbe177ca9aabea4e87b/package.json#L65-L67

https://github.com/lint-staged/lint-staged

The eslint integration with prettier is to adjust the eslint rules (which differ from prettier by default on some areas of formatting). More details here: https://prettier.io/docs/integrating-with-linters

In short, adjusting the .prettierignore _should_ apply it to our JSON files - although it will only apply on commit to ones you touched I think. To manually run it on everything to see what. changes, try `npx prettier . --write` in teh root dir.

Roaders:
remaining items that need looking at:

- add tests for DesktopAgentProxy
- DOCUMENTATION (probably need to speak to @kriswest about this)
- clarification of dynamic registration in docs
- add new add intent section to workbench

kriswest:
@Roaders you'll need to resolve conflicts in this one

kriswest:
Needs the conformance test definitions adding and then implementing

Roaders:
conformance tests all pass
{7E7492D2-0675-4B33-A827-5A266BC67AF8}

Roaders:
Locally all tests pass, build passes and workbench compiles and runs fine. Workbench now allows context to be applied when adding an intent listener. docs updated as per agents.md

I don't know why easy cli fails. It looks like all the commits in the PR belong to me so not sure why it's getting confused.

Roaders:
aha - I see why it's failing. 
wow. that user must be getting SO MANY commits attributed to them!

kriswest:
@Roaders can yuou check your failing test then we can wade into review and merge

website/docs/api/ref/DesktopAgent.md
```diff
@@ -476,6 +477,62 @@ listenerResult := <-desktopAgent.AddIntentListener("fdc3.contact", func(context
...
+  intent: string,
+  contextType: string | string[],
+  handler: IntentHandler
+): Promise<Listener>;
+```
+
+</TabItem>
+</Tabs>
+
+Adds a listener for incoming intents raised by other applications, via calls to [`fdc3.raiseIntent`](#raiseintent) or [`fdc3.raiseIntentForContext`](#raiseintentforcontext), filtered by one or more context types. The handler will only be invoked when the incoming intent's context `type` matches one of the supplied `contextType` values; all other behaviour is identical to [`addIntentListener`](#addintentlistener) — see that section for details, restrictions and result handling that also apply here.
```
kriswest, line 496: You may need something here on the new `IntentListenerConflict` Error in case a conflicting listener was already in place. WIll probably need to check the agent-proxy code to make sure its done there also.

packages/fdc3-agent-proxy/src/intents/DefaultIntentSupport.ts
```diff
@@ -281,7 +281,23 @@ export class DefaultIntentSupport implements IntentSupport {
   }
 
   async addIntentListener(intent: string, handler: IntentHandler): Promise<Listener> {
```
kriswest, line 275: @kemerava @Roaders I think we need to be checking for conflicting intent listeners here and throwing the new IntentListenerConflict Error for both the existing addIntentListener function and new addIntentListenerWithContext function. AFAIK it wasn't implemented yet - please correct me if I'm wrong there @kemerava

toolbox/fdc3-workbench/src/store/IntentStore.ts
```diff
@@ -39,6 +39,7 @@ class IntentStore {
 
   async addIntentListener(
```
kriswest, line 40: This was working nicely when Rob and I tested it - however you don't get any visual indication in Workbench's SystemLog which context types each listener is register for (now there are multiple).

kriswest (commented):
Needs a little more work to add the checks needed in agent-proxy to prevent conflicting listener registrations + a visual indication in workbench of what a intent listener is filtering for in terms of context type.

Currently, If you register multiple conflicting intentlisteners then the resolver pops up with multiple options (one per listener) but then can't figure out how to delicer to the one you selected (as they are all the same app. We tested an unfiltered listener vs. a filtered one and used the filtered context type - so it could have fired either intent listener.

kriswest:
Other than the abovve this is pretty close @Roaders 👏

website/docs/api/ref/DesktopAgent.md
Roaders: I think that this is covered in the next paragraph.

toolbox/fdc3-workbench/src/store/IntentStore.ts
Roaders: added a visual indicator to indicate which context was used when creating an intent listener. Also clear the context when an intent listener is added

packages/fdc3-agent-proxy/src/intents/DefaultIntentSupport.ts
Roaders: added this along with conformance tests for it

website/docs/api/conformance/Intents-Tests.md
```diff
@@ -335,10 +335,57 @@ As the methods of resolving ambiguous intents are often user interactive, it is
...
+Listeners for **different** intents MUST be allowed, regardless of whether they are filtered or unfiltered:
+
+| App  | Step                      | Details |
+|------|---------------------------|---------|
+| Test | 1. Add Intent Listener | App performs `fdc3.addIntentListener("aTestingIntent1")`. |
+| Test | 2. Add Intent Listener for a different intent | App performs `fdc3.addIntentListener("aTestingIntent12")`. |
+
+- `AddingIntentListenersForDifferentIntentsIsAllowed` ![3.0+](https://img.shields.io/badge/FDC3-3.0+-purple): Perform above steps which should successfully add both listeners (no error thrown).
+
+### Re-adding an Intent Listener after Unsubscribe
```
Roaders, line 381: @kriswest I updated these docs and to me they now seem correct. Please shout if they are not.

packages/fdc3-standard/src/api/DesktopAgent.ts
```diff
@@ -383,6 +383,35 @@ export interface DesktopAgent {
...
 
+  /**
+   * Adds a listener for incoming intents raised by other applications, via calls to `fdc3.raiseIntent` or `fdc3.raiseIntentForContext`, but filtered to one or more context types. The handler will only be invoked when the incoming intent's context `type` matches one of the supplied `contextType` values.
+   *
+   * The `contextType` parameter MAY be either a single context type string or an array of context type strings. See `addIntentListener` for details and restrictions for both usage and implementation that also apply to this method.
+   *
+   * ```js
+   * //Handle a raised intent filtered to a single context type
+   * const listener = await fdc3.addIntentListenerWithContext('StartChat', 'fdc3.contact', context => {
+   *   // start chat has been requested by another application
```
Roaders, line 394: @kriswest I think that this is the section that we were discussing in the meeting today but on a second look this seems fine to me. This commented section is within a triple backticks sections. Please correct me if I am wrong.

Roaders:
@kriswest I looked at the 2 areas we discussed and updated one. I think the other one is actually fine. Let mw know if not. Hopefully this can be merged now.

packages/fdc3-standard/src/api/DesktopAgent.ts
kriswest: yep, looks correct to me

website/docs/api/conformance/Intents-Tests.md
kriswest: Yep, looks correct now. The references between tests are something we'll have to fix one day - but its the current style of a great many other tests in this set!

kriswest (approved):
LGTM!

Issue #1290, opened 2024-07-24
Add property to support analytics across apps


julianna-ciq:
There's a growing interest in tracking workflow and data usages, with tools like DataDog, OpenTelemetry, Sumo, or Zipkin. While Desktop Agents already have a lot of the information they need to provide these integrations, there's no reliable way for a DA to identify when multiple API calls are related. Proposed below is an optional addition to context payloads, which will provide hooks for telemetry tracking tools.

This proposal follows the same approach that HTTP used for their [traceparent headers](https://www.w3.org/TR/trace-context/#examples-of-http-traceparent-headers), as well as the approach used in the [open source OpenTelemetry standard](https://opentelemetry.io/docs/concepts/signals/traces/).

Supercedes the previously submitted issue https://github.com/finos/FDC3/issues/1142

## Proposal

While a full integration with a telemetry tool would need to include information about <ins>which app</ins> shared <ins>which context</ins> <ins>when</ins>, none of those features would require a change to the Desktop Agent standard. The DA already knows all of those pieces. The only thing the DA doesn't already know, per the standard, is if an FDC3 context payload is shared *in response* to another FDC3 message.

This proposal is for a single, optional property change that enables the Desktop Agent and individual Apps to communicate this narrow detail. 

### Desktop Agent
Desktop Agents *SHOULD* include a `traceId` (a properly random UUID). If an intent/context is raised/broadcast with a `traceId`, the desktop agent should do nothing. If there is no `traceId`, the deskotp agent *SHOULD* add one.

### Individual App
If an app receives a context payload, and wants to do something *in response*, the app should forward the received `traceId`. If the app is not responding to anything, it can safely ignore the `traceId`.

### Example

Price App broadcasts this context.
```json
{
  "type": "fdc3.instrument",
  "id": {
    "ticker": "MSFT"
  }
}
```

Since there's no `traceId`, the desktop agent will inject one. Order App receives this context payload:
```json
{
  "type": "fdc3.instrument",
  "id": {
    "ticker": "MSFT"
  },
  "traceId": "699381f3-284f-4014-a476-e114ef951cef"
}
```

The Order App will then send off price information about MSFT. Since it's *responding* to a context, it will include the `traceId` of the context it's responding to.
```json
{
  "type": "fdc3.valuation",
  "price": 20,
  "value": 100,
  "CURRENCY_ISOCODE": "USD",
  "traceId": "699381f3-284f-4014-a476-e114ef951cef"
}
```

[...]

Issue #1546, opened 2025-03-20
Support the ability to specify where I would like a new app to open


Roaders:
### Use Case:
When opening a new app it is likely that as a developer I might want to be able to specify where the app should be opened.

If I have a container that has 4 slots for applications when I open a new chart I might want to specify that it opens top left in an empty spot or possible load over an existing app in the bottom left corner.

### Workflow Examples
I would imagine that the best way to handle this would be to add an additional property to `agent.open()` and to `agent.raiseIntent()` that the developer could pass any additional context that they would like:

```ts
open(app: AppIdentifier, context?: Context, openParams?: any): Promise<AppIdentifier>;
```

```ts
raiseIntent(intent: string, context: Context, app?: AppIdentifier, openParams?: any): Promise<IntentResolution>;
```

(I am definitely open to a better name for the parameter)

the location that the new app should be opened in could then be passed:

```ts
agent.open("my-new-app", my-context, { targetIframe: "top-left" });
```

the `openParams` (which I am sure could do with a better name) would be passed to the DesktopAgent that would be responsible for opening the new app and it could do whatever it wanted with it to open the apps in the correct way.

robmoffat:
can we enumerate the open params?  If this is being added to the standard, wouldn't they need to be standardised too?

kriswest:
> can we enumerate the open params? If this is being added to the standard, wouldn't they need to be standardised too?

We should yes, however, that's where things will get much more complex... We've got to consider: 

- Web browsers
  - with iframes
  - with multiple widgets in a page
  - with layout management of windows (see https://developer.mozilla.org/en-US/docs/Web/API/Window_Management_API)
  - golden layout or swimlane style layout management
- Containers
  - variable coordinate systems (ask an Electron developer about display scaling and DPI vs. DIP points and rounding errors sometime...)
  - window grouping and layout systems

Getting down to a standard set of options will, I suspect, prove difficult

Roaders:
As @novavi suggests it might be very hard for an app that is requesting another app to open to know what options are available to it for open options.

I think that the simplest scenario might be something like:

App A is a webpage that has 4 iframes within it arranged in a grid. The user drags an instrument onto the placeholder for the top left iframe. App A then requests the App B opens the dragged instrument in the top left iframe.

This is a very simple example as App A owns the space that the app will be opened in. I suspect that the payload in this case might look something like:

```TS
{
    type: "open-app-context",
    targetApp: "AppA@myhost.com",
[...]

PR #1595, opened 2025-05-21
[WIP] Expand broadcast, raise*, and open methods to include metadata

julianna-ciq:
## Describe your change

Adds the `metadata` property at the end of the `.broadcast()`, `.raiseIntent()`, `.raiseIntentForContext()`, and `.open()` methods.
Includes schema updates, doc updates, and implementation updates.

Provides placeholder space for `signature` (see #1571) and `hostParams` (#1546).

### Related Issue
Resolves https://github.com/finos/FDC3/issues/1290

## Review Checklist

  - [x] **Docs & Sources**: If yes, were both documentation (/docs) and sources updated?<br/>
  - [x] **Schemas**: If yes, were changes applied to the Bridging and FDC3 for Web protocol schemas?<br/>
    - [x] If yes, was code generation (`npm run build`) run and the results checked in?<br/>
- [x] **Context types**: Were new Context type schemas created or modified in this PR?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18982,8 +18982,8 @@
       }
     },
     "node_modules/vite": {
-      "version": "6.3.6",
-      "integrity": "sha512-0msEVHJEScQbhkbVTb/4iHZdJ6SXp/AvxL2sjwYQFfBqleHtnCqv1J3sa9zbWz/6kW1m9Tfzn92vW+kZ1WV6QA==",
+      "version": "6.4.1",
+      "integrity": "sha512-+Oxm7q9hDoLMyJOYfUYBuHQo+dkAloi33apOPP56pzj+vsdJDzr+j1NISE5pyaAuKL4A3UD34qd0lx5+kfKp2g==",
       "dev": true,
       "license": "MIT",
       "dependencies": {
@@ -20262,7 +20262,7 @@
         "nodemon": "^3.1.7",
         "rimraf": "^6.0.1",
         "typescript-eslint": "^8.17.0",
-        "vite": "^6.3.6"
+        "vite": "^6.4.1"
       }
     },
     "toolbox/fdc3-for-web/demo/node_modules/typescript": {
@@ -20347,7 +20347,7 @@
         "globals": "^15.14.0",
         "typescript": "^5.6.3",
         "typescript-eslint": "^8.17.0",
-        "vite": "^6.3.6"
+        "vite": "^6.4.1"
       }
     },
     "toolbox/fdc3-for-web/reference-ui/node_modules/typescript": {
@@ -20401,7 +20401,7 @@
         "terser": "^5.34.1",
         "typescript": "^5.6.3",
         "uuid": "^9.0.1",
-        "vite": "^6.3.6",
+        "vite": "^6.4.1",
         "vite-plugin-node-polyfills": "^0.23.0",
         "web-vitals": "^1.1.2"
       }
--- a/toolbox/fdc3-for-web/demo/package.json
+++ b/toolbox/fdc3-for-web/demo/package.json
@@ -19,7 +19,7 @@
     "nodemon": "^3.1.7",
     "rimraf": "^6.0.1",
     "typescript-eslint": "^8.17.0",
-    "vite": "^6.3.6"
+    "vite": "^6.4.1"
   },
   "dependencies": {
     "@finos/fdc3": "2.2.0",
--- a/toolbox/fdc3-for-web/reference-ui/package.json
+++ b/toolbox/fdc3-for-web/reference-ui/package.json
@@ -20,7 +20,7 @@
     "globals": "^15.14.0",
     "typescript": "^5.6.3",
     "typescript-eslint": "^8.17.0",
-    "vite": "^6.3.6"
+    "vite": "^6.4.1"
   },
   "dependencies": {
     "@finos/fdc3": "2.2.0"
--- a/toolbox/fdc3-workbench/package.json
+++ b/toolbox/fdc3-workbench/package.json
@@ -68,7 +68,7 @@
     "terser": "^5.34.1",
     "typescript": "^5.6.3",
     "uuid": "^9.0.1",
-    "vite": "^6.3.6",
+    "vite": "^6.4.1",
     "vite-plugin-node-polyfills": "^0.23.0",
     "web-vitals": "^1.1.2"
   }


packages/fdc3-agent-proxy/src/apps/DefaultAppSupport.ts
```diff
@@ -78,6 +89,7 @@ export class DefaultAppSupport implements AppSupport {
         },
         context,
       },
+      metadata,
```
kriswest: In all other cases, parameters to FDC3 functions go into the payload element... I think we should stick to that pattern

packages/fdc3-agent-proxy/src/intents/DefaultIntentSupport.ts
```diff
@@ -130,16 +132,25 @@ export class DefaultIntentSupport implements IntentSupport {
...
     const request: RaiseIntentRequest = {
       type: 'raiseIntentRequest',
       payload: {
         intent,
         context,
-        app: app,
+        app,
       },
-      meta: meta,
+      metadata: {
```
kriswest: See previous comment about keeping this in the payload

packages/fdc3-standard/src/api/ContextMetadata.ts
```diff
@@ -18,3 +18,15 @@ export interface ContextMetadata {
    */
   readonly source: AppIdentifier;
 }
+
+export interface AppProvidableContextMetadata {
```
kriswest: Should we shorten this to `AppProvidableMetadata` or `AppContextMetadata`?

packages/fdc3-standard/src/api/ContextMetadata.ts
```diff
@@ -18,3 +18,15 @@ export interface ContextMetadata {
...
+  signature?: string;
+  traceId?: string;
+}
+
+export interface DesktopAgentProvidableContextMetadata {
```
kriswest: Given that most of these fields are forwarded on from app provided metadata, we should could/should just integrate it into ContextMetadata... I was expecting to see a relationship to that type as that is the one we currently return, containing `source`

kriswest: ContextMetadata appears defunct after this change, so it either needs to be removed or these changes integrated into it. I prefer the latter (put the new fileds into that type and keep it)

website/docs/api/spec.md
```diff
@@ -840,6 +840,61 @@ The [Context specification](../context/spec#assumptions) recommends that complex
...
+### Metadata
 
-Optional metadata about each context message received, including the app that originated the message, SHOULD be provided by the desktop agent implementation to registered context handlers on all types of channel. As this metadata is optional, apps making use of it MUST handle cases where it is not provided.
+See [[#Context_Metadata]]
+
+## Context Metadata
+
+Optional metadata about each context or intent message received SHOULD be provided by the desktop agent implementation to registered intent handlers. As this metadata is optional, apps making use of it MUST handle cases where it is not provided.
+
+Registered listeners SHOULD receive the following properties:
```
kriswest: YOu SHOULD be able to pass these - in which case it SHOULD or MUST be passed on. But it needs to be conditional on an app setting it (apart from source which DA MSUT provide and MUST NOT forward if passed by app to avoid spoofing).

website/docs/api/spec.md
```diff
@@ -840,6 +840,61 @@ The [Context specification](../context/spec#assumptions) recommends that complex
...
+})
+```
+
+### Timestamp
+
+The Desktop Agent MAY provide timestamp information to intent handlers. 
+
+Apps SHOULD NOT provide timestamp information. If an app provides a timestamp, the Desktop Agent MUST ignore the timestamp.
+
+### Source
```
kriswest: I'd use the original label (Originating App Metadata) in the content somewhere to retain the link to the use case.

website/docs/api/spec.md
```diff
@@ -535,9 +535,9 @@ When an instance of an application is launched, it is expected to add an [`Inten
 
 Intent handlers SHOULD be registered via [`fdc3.addIntentListener`](ref/DesktopAgent#addintentlistener) within 15 seconds of the application launch (the minimum timeout Desktop Agents are required to provide) in order to be widely compatible with Desktop Agent implementations. Individual Desktop Agent implementations MAY support longer timeouts or configuration to control or extend timeouts.
 
-### Originating App Metadata
+### Metadata
```
kriswest: I'd add a brief textual description saying additional metadata may be passed ...

packages/fdc3-standard/src/api/ContextMetadata.ts
```diff
@@ -18,3 +18,15 @@ export interface ContextMetadata {
    */
   readonly source: AppIdentifier;
 }
+
+export interface AppProvidableContextMetadata {
+  signature?: string;
```
robmoffat, line 23: why are apps limited on the metadata they can send?  Shouldn't they be allowed to put anything in here?

website/docs/api/spec.md
```diff
@@ -840,6 +840,61 @@ The [Context specification](../context/spec#assumptions) recommends that complex
...
+Optional metadata about each context or intent message received SHOULD be provided by the desktop agent implementation to registered intent handlers. As this metadata is optional, apps making use of it MUST handle cases where it is not provided.
+
+Registered listeners SHOULD receive the following properties:
+* traceId
+* timestamp
+* source
+
+### Trace information
+
+The Desktop Agent SHOULD provide a `traceId` to intent handlers. If the originating app provides `traceId` infomration, the Desktop Agent SHOULD use provide the `traceId` from the app. If the originating app does not provide `traceId` information, the Desktop Agent SHOULD generate a uuid.,
```
robmoffat, line 858: spelling of "information"

website/docs/api/spec.md
```diff
@@ -840,6 +840,61 @@ The [Context specification](../context/spec#assumptions) recommends that complex
...
+  }
+  */
+
+ fdc3.raiseIntent("Start Call", contactContext, {
+  traceId: metadata.traceId
+ })
+})
+```
+
+### Timestamp
```
robmoffat, line 890: do we need this?  We already have a timestamp in the message meta I think

kriswest: I agree, tinestamp could/should be DA provided rather than app provided - even if it's the client code generating it

packages/fdc3-standard/src/api/ContextMetadata.ts
julianna-ciq: The discussions around #1290 indicated that we should limit the metadata. Let's discuss this at the next Web meeting.

website/docs/api/spec.md
julianna-ciq: It was in my notes as one of the few properties we should support. Let's discuss what metadata properties should be included at the next Web meeting.

packages/fdc3-schema/schemas/api/openRequest.schema.json
```diff
@@ -17,6 +17,9 @@
 				"payload": {
 					"$ref": "#/$defs/OpenRequestPayload"
 				},
+				"metadata": {
```
robmoffat: @kriswest @julianna-ciq

I've been looking at this.  It doesn't seem like this gets generated into `OpenRequest` object in BrowserTypes.  I'm not enough of an expert on JSON schema to really say why, except that I can add the metadata into the base type, AppRequest and it seems to push through. 

However, I think this is the wrong place for this:  I think the context metadata belongs in the payload, along with the context.   After all, the message already has a `meta` field (which is the meta of the message) so it seems like we need to have meta inside the payload (the meta of the payload).

kriswest: I agree that we should create fields in the payload, rather than as a sibling of it. That will move it further from the `meta` element that relates to the "envelope" or message format and more clearly tie it to a particular API call.

The reason that its not generating out in the current location is that the extra property is not allowed in the AppRequest schema, so when OpenRequest has AppRequest composed with it (as all requests do) the field gets dropped - that is part of the role of AppRequest: to govern the shape of those things derived from it and prevent [additionalProperties](https://github.com/finos/FDC3/blob/d0fb59aeb3086a3ad6adda414068ccbbb5360700/packages/fdc3-schema/schemas/api/appRequest.schema.json#L65) that should not be there.

I'd say just go ahead and move it into the payload for every API message that should transmit either the app provided or super-set of app and DA-provided metadata (i.e. requests include appProvided, responses include both app and DA provided).

packages/fdc3-standard/src/api/DesktopAgent.ts
```diff
@@ -271,7 +272,12 @@ export interface DesktopAgent {
    * }
    * ```
    */
-  raiseIntent(intent: Intent, context: Context, app?: AppIdentifier): Promise<IntentResolution>;
+  raiseIntent(
+    intent: Intent,
+    context: Context,
+    app?: AppIdentifier,
+    metadata?: AppProvidableContextMetadata
+  ): Promise<IntentResolution>;
```
robmoffat: This allows metadata to go from the app raising the intent to the app receiving it.  However, it doesn't on its own allow metadata to come back from the receiver back to the raiser.  We might need to extend the IntentResolution interface to add `getResultMetadata` perhaps?

kriswest: Good point! Changing the return type of `IntentResolution.getResult()` would not be ideal (breaking).

However, we should also conider the fact that `IntentResolution` is itself a metadata object and returns the source for the resolution at least: 
- what app did the DA deliver it to..., 
- what was the intent (could have been chosen by the user or DA if there was only one)

But it is returned before the result (should come back at the same time that intent is delivered - so before the result). Hence, before we pick a solution here, what additional metadata might someone want?  I'd argue timestamp isn't as interesting in intents (as you can't `getCurrentContext()` and pick it up later). Signatures are interesting (and would have to come from the app, rather than the DA). Anything else? Signature is probably enough of a use case...

packages/fdc3-standard/src/api/ContextMetadata.ts
kriswest: Meh, changed my mind

packages/fdc3-standard/src/api/ContextMetadata.ts
kriswest: it needs to be limited to promote standardisation of the content (so its interoperable between firms/implementations), however adding a less controlled custom element was discussed to contain properties that are/will always be specific to a firm/app - and will also allow prototyping new use cases.

Future PRs would extend the standardised properties as we come up with them.

packages/fdc3-standard/src/api/DesktopAgent.ts
kriswest: Spent some time with @julianna-ciq talking about this - its one of the harder parts to solve in the design because its difficult for `IntentHandler` functions to return metadata, currently they can return `Context`, `Channel` or `void`, but we'd need a compound type with the metadata to return.

Further, `IntentResolution` is itself a metadata object that overlaps with the ContextMetadata some already, in that it contains the `source` field. A `trace_id` is less useful as it should be the same value used in the original call if provided and its still connected to that call. `timestamp` is less interesting as we don't have a call like `getCurrentCOntext` to worry about where it will be received later.

The one case I thought it mattered on is `signature`, there is a workaround where an app could return a channel and deliver signed results over that instead.
 
Hence, with added complexity and reduced utility, we think this should just be left out for now, and revisited if anyone comes up with a solid us-case for needing it.

packages/fdc3-standard/src/api/ContextMetadata.ts
robmoffat: Just in the https://github.com/finos/FDC3/issues/1644 meeting, we are discussing the CDM use case.  In this case, a CDM message may be broken down into a number of different FDC3 contexts.  e.g. a priceQuantity message contains, date, instrument, amount, exchange, each of which might be a different context broadcast.

In this case, to re-assemble the message, it would be nice to use a common traceId across them all, but also have a count metadata item, so you know how many broadcasts should be bundled together.

kriswest:
Wouldn't you also have the single compound message with them all in? What is suggested in the current docs is that you compose one large context made up of smaller ones, broadcast all the smaller ones with the big one last. Apps can listen for whatever they support.

Sent from Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

robmoffat:
> Wouldn't you also have the single compound message with them all in? What is suggested in the current docs is that you compose one large context made up of smaller ones, broadcast all the smaller ones with the big one last. Apps can listen for whatever they support.

Yeah makes sense.

PR #1619, opened 2025-06-21
feat: adds node implementation for app directory

### LINKED ISSUE #563: Revive and update or replace AppD reference implementation
## Enhancement Request

It is advisable that any standard provide a reference implementation that may be used to confirm or compare the expected behavior of an implementation of the standard and/or be used as a base for other implementations. The AppD is one such standard and it is likely that the provision of a usable reference implementation will help drive its uptake throughout the financial services desktop software community.

In the past, a java implementation of the AppD was provided at:  https://github.com/FDC3/appd-poc. This is not currently maintained, nor included in the FDC3 repository. Reviving this implementation, or providing a new one (ideally based on Node.js) is desirable.

### Use Case:
- Reference implementations may be used:  
  - directly, saving a firm from implementing their own AppD
  - as a starting point by firm implementing their own AppD
  - as an example of expected behavior from AppD.

Yousuf-Ejaz:
## Describe your change

resolves #563 

### Related Issue
Raising the PR according to our discussion at [Github](https://github.com/finos/FDC3/issues/563#issuecomment-2983341261)

## Review Checklist


toolbox/fdc3-appd-node/src/keys/private.key
```diff
@@ -0,0 +1,28 @@
+-----BEGIN PRIVATE KEY-----
```
robmoffat, line 1: public and private keys shouldn't be checked into the codebase.

robmoffat (commented):
This is a great effort for a hackathon, well done.

I'm not convinced this should make it into the main FDC3 repo, though:

1.  It's highly specific, and requires mongo or another database to be configured to hold user details.    This might be useful in production environments, but probably isn't helpful from the point of view of demonstrating FDC3 to people downloading the repo.

2.  If we did want to include it, we should integrate with the demo FDC3 server in `toolbox/fdc3-for-web/demo` and get it to use that.   At the moment, that just picks up a bunch of AppD records from random URLs.   I can see an argument for using this instead, but it would require a bit of work to simplify.

3.  Weirdly, (and this is not a compliant for the author, but for the AppD Spec) I've not really used the AppD endpoints either in Sail or in the above demo DA.   Are they actually useful?  It seems to me that the specification that really matters is the AppD record, not the container they come in.

kriswest:
@robmoffat having implemented support for an appD in a desktop agent, I can confirm that the rest endpoints are essential, because you need to standardised response format when querying multiple appDs for their. Yes, 95% of that definition is the record format, but if it's wrapped in rest response in different ways, it's not a lot of use to you...

However, we could split out the record format definition into its own schema, that the appd Open API spec then references, without making any change to that response format I believe. We could then use that schema separately and generate typescript types from it - Which would better serve the other ways that the appd record format is used. WDYT @robmoffat  - that could go on an issue. 

Also many thanks for reviewing and responding to this and the other at the implementation issue :star:

PR #1695, opened 2025-11-03
1693 web reference impl to use messageport

### LINKED ISSUE #1693: FDC3 for Web reference implementation always uses a websocket
The FDC3 for Web reference implementation demonstrates two different approaches to communicating with a web-based desktop agent: direct comms with a parent window or frame via a MessagePort, or the load of an adaptor into an iframe and then communication with that via the MessagePort while the adaptor handles onwards communication with the desktop agent, typically over a websocket. 

However, a shortcut was taken in the implementation in that the direct communication version is also hooked into the web socket unnecessarily at:
https://github.com/finos/FDC3/blob/1300f2b13ff0aebac47196f98ac5254b2715d939/toolbox/fdc3-for-web/demo/src/client/da/dummy-desktop-agent.ts#L194-L196

This slightly simplifies the code for posting responses - but at the cost of making the examples harder to understand (i.e. its not good in a reference implementation). This should be updated to provide an example that does not involve the use of the websocket and consideration given to splitting into separate examples for the two different cases (instead of a switch) as that would further simplify the code, making comprehension much easier

kriswest:
## Describe your change

Updates the FDC3 for Web reference implementation (/toolbox/fdc3-for-web/demo) to use direct MessagePort comms with a child window or frame when that option is selected, without involving the websocket used in the iframe adaptors comms. This makes the example slightly easier to follow. A future update should further separate these cases for easier comprehension.

Includes a correction to the eslint config for those projects (setting the tsconfigRootDir).

### Related Issue

resolves #1693

## Review Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 
-* Corrected /toolbox/fdc3-for-web/demo to only use MessagePort communication when 'Parent Post-Message' selected in the demo. ([#1694](https://github.com/finos/FDC3/pull/1694))
+* Corrected /toolbox/fdc3-for-web/demo to only use MessagePort communication when 'Parent Post-Message' selected in the demo. ([#1695](https://github.com/finos/FDC3/pull/1695))
 * Corrected the property set in WCP1Hello by getAgent that indicates whether an intent resolver is needed. ([#1684](https://github.com/finos/FDC3/issues/1684))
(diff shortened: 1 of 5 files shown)


Roaders (dismissed):

kriswest:
@Roaders could you re-approve, had to adjust the eslint config again for CI to pass

robmoffat:
I'm a bit confused about this.  Having re-read the description again it seems that this isn't about changing the way anything works, but just making It easier to understand?   Is that right?

Roaders (dismissed):

Roaders (approved):

PR #1696, opened 2025-11-05
Server Context Refactor

robmoffat:
## Describe your change

- Simplified the relationships between the `MessageHandlers`, `ServerContext` and `FDC3Server` by combining `ServerContext` and `FDC3Server` into `FDC3ServerInstance`, with an abstract implementation re-used in `fdc3-get-agent` and `fdc3-web-impl` tests and the `demo` desktop agent itself.
- This removes a cyclic relationship between these three classes which existed before. 
- Moved all server state into the `FDC3ServerInstance` class, giving a better separation of concerns with `MessageHandler`s: these only manage conversion to and from the DACP / WCP format.
- Moved heartbeat state logic out of `HeartbeatHandler` and into `FDC3ServerInstance` via the `heartbeatActivity` method.  This means that desktop agent implementations extending `FDC3ServerInstance` are able to customise the connection / disconnection process themselves if they think they have a better idea of the state of applications.
- Conformance tests all passing.

### Related PR

This is based on #1586 which should be merged and released first as it contains fixes for `fdc3-agent-proxy`.

## Review Checklist

### DIFF
--- a/packages/fdc3-testing/README.md
+++ b/packages/fdc3-testing/README.md
@@ -1,0 +1,220 @@
+# @finos/fdc3-testing
+
+A shared testing infrastructure for FDC3 Cucumber-based test suites. This module provides a **domain-specific language (DSL)** for testing FDC3 APIs in a readable, declarative way.
+
+## Purpose
+
+The testing module provides:
+
+- **Reusable step definitions** for common testing patterns
+- **A `PropsWorld` class** that holds test state in a key-value store
+- **Variable resolution** using `{variable}` syntax to reference stored values
+- **JSONPath matching** for validating complex objects
+- **Schema validation** to verify messages conform to FDC3 schemas
+
+## Installation
+
+```json
+{
+  "devDependencies": {
+    "@finos/fdc3-testing": "2.2.1-beta.4"
+  }
+}
+```
+
+## The Testing DSL
+
+The DSL allows tests to be written in natural language while still being precise.
+
+### Variable References
+
+Variables are referenced using curly braces: `{variableName}`. Special values include:
+
+| Syntax | Meaning |
+|--------|---------|
+| `{null}` | Represents null |
+| `{true}` / `{false}` | Boolean values |
+| `{123}` | Numeric values (any number) |
+| `{result}` | The result of the last operation |
+| `{api}` | Typically the Desktop Agent instance |
+| `{myVar}` | Any variable stored in the test world |
+
+### Core Step Definitions
+
+#### Calling Methods
+
+```gherkin
+When I call "{api}" with "getOrCreateChannel" with parameter "channel-name"
+When I call "{api}" with "raiseIntent" with parameters "Buy" and "{instrumentContext}"
+When I call "{api}" with "open" with parameters "appId" and "{context}" and "{options}"
+```
+
+#### Storing References
+
+```gherkin
+When I refer to "{result}" as "channel1"
+```
+
+#### Promise Resolution
+
+```gherkin
+Then the promise "{myPromise}" should resolve
+Then the promise "{myPromise}" should resolve within 10 seconds
+```
+
+#### Assertions
+
+```gherkin
+Then "{result}" is null
+Then "{result}" is not null
+Then "{result}" is true
+Then "{result}" is false
+Then "{result}" is undefined
+Then "{result}" is empty
+Then "{result}" is "expected value"
+Then "{result}" is an error
+Then "{result}" is an error with message "UserCancelledResolution"
+```
+
+#### Object Matching
+
+```gherkin
+Then "{result}" is an object with the following contents
+  | source.appId | source.instanceId |
+  | bank         | b1                |
+```
+
+#### Array Validation
+
+```gherkin
+Then "{result}" is an array of objects with the following contents
+  | type | id.channelId |
+  | user | one          |
+  | user | two          |
+
+Then "{result}" is an array of objects with length "{expectedLength}"
+
+Then "{result}" is an array of strings with the following values
+  | value   |
+  | string1 |
+  | string2 |
+```
+
+#### Schema Validation
+
+The `matches_type` column validates that data conforms to an FDC3 schema:
+
+```gherkin
+Then messaging will have posts
+  | payload.intent | payload.context.type | matches_type       |
+  | OrderFood      | fdc3.instrument      | raiseIntentRequest |
+```
+
+#### Utilities
+
+```gherkin
+Given schemas loaded
+Given we wait for a period of "500" ms
+Given "{handler}" is a invocation counter into "{count}"
+Given "{fn}" is a function which returns a promise of "{value}"
+```
+
+### JSONPath Support
+
+Table columns use JSONPath expressions to match nested object properties:
+
+| Expression | Matches |
+|------------|---------|
+| `source.appId` | `result.source.appId` |
+| `payload.context.id.ticker` | Deeply nested values |
+| `payload.app.instanceId` | Can match `{null}` for undefined |
+
+## Usage
+
+### Basic Setup
+
+Import and call `setupGenericSteps()` in your step definitions:
+
+```typescript
+import { setupGenericSteps } from '@finos/fdc3-testing';
+
+setupGenericSteps();
+```
+
+### Extending PropsWorld
+
+Extend `PropsWorld` if you need additional test state:
+
+```typescript
+import { PropsWorld } from '@finos/fdc3-testing';
+
+export class MyWorld extends PropsWorld {
+  myCustomState: SomeType;
+}
+```
+
+### Example Feature File
+
+```gherkin
+Feature: Basic Intents Support
+
+  Background: Desktop Agent API
+    Given A Desktop Agent in "api"
+    And schemas loaded
+    And app "chipShop/c1" resolves intent "OrderFood" with result type "void"
+    And app "bank/b1" resolves intent "Buy" with context "fdc3.instrument" and result type "fdc3.order"
+    And "instrumentContext" is a "fdc3.instrument" context
+
+  Scenario: Raising an intent and invoking the intent resolver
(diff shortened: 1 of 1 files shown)


robmoffat:
@seewhatson. @patrocinio

toolbox/fdc3-for-web/fdc3-web-impl/src/FDC3ServerInstance.ts
```diff
@@ -0,0 +1,312 @@
...
+  instanceId: string;
+  intentName: string;
+  listenerUUID: string;
+};
+
+/**
+ * Handles messaging to apps and opening apps for ONE FDC3 environment.
+ * Stores all state for MessageHandlers to use.
+ */
+export interface FDC3ServerInstance {
```
kriswest: FDC3ServerInstance suggests a concrete implementation rather than an interface... Perhaps better to stick to FDC3Server

toolbox/fdc3-for-web/fdc3-web-impl/src/FDC3ServerFactory.ts
```diff
@@ -0,0 +1,55 @@
...
+   */
+  shutdownEverything(): Promise<void>;
+}
+
+/**
+ * Handles setting up the handlers with the right timeouts.
+ * Subclass this and implement the createInstance method with your
+ * implementation of FDC3ServerInstance.
+ */
+export abstract class AbstractFDC3ServerFactory {
```
kriswest: What advantage is there to having this factory pattern, separate from an Implementation of an FDC3ServerInstance? The constructor it provides is opinionated about what it sets up (such as the handlers that would be used in instances). Why don't we just have an FDC3Server interface and a default implementation of it that sets up the handlers (which again conform to an interface)?

The example in demo put both the factory and instance implementations in the same file and the factory is just passing on args to the instances' constructor. Why not just construct instances directly, or even just have a static createInstance function on an FDC3Server implementation (if you regularly need to create new instances - which I don't actually think is the case.  That way the implementation is on control of what handlers etc. that it sets up - that is exactly the sort of thing that is likely to vary between implementations (e.g. a firm using this needs to provide a custom IntentHandler in order to tie it into some other subsystem).

toolbox/fdc3-for-web/fdc3-web-impl/src/FDC3ServerInstance.ts
```diff
@@ -0,0 +1,312 @@
+import { AppIdentifier, AppIntent, DisplayMetadata, PrivateChannelEventTypes } from '@finos/fdc3-standard';
+import { Context } from '@finos/fdc3-context';
+import { AppRegistration, InstanceID, State, ReceivableMessage } from './AppRegistration';
+import { PendingApp } from './PendingApp';
+import { Directory } from './directory/DirectoryInterface';
+
+export enum HeartbeatActivityEvent {
```
kriswest: might be worth extracting these types to a new file and consolidating with AppRegistration and any others

toolbox/fdc3-for-web/demo/src/client/da/DemoFDC3ServerInstance.ts
```diff
@@ -48,19 +51,23 @@ function isRunningAppRegistration(
   return !!(details as RunningAppRegistration).window;
 }
```
kriswest, line 53: Extracting the types and typeguards above into a separate file would make this easier to read

toolbox/fdc3-for-web/demo/src/client/da/DemoFDC3ServerInstance.ts
```diff
@@ -70,27 +77,30 @@ export class DemoServerContext implements ServerContext<DemoAppRegistration> {
   /**
    * Sets the appId, url, state and either the window or a Promise<Window> for a given connection UUID.
    */
-  setInstanceDetails(uuid: InstanceID, meta: RunningAppRegistration | LaunchingAppRegistration): void {
+  setInstanceDetails(uuid: InstanceID, meta: AppRegistration): void {
+    // Type assertion to handle our extended type
+    const demoMeta = meta as DemoAppRegistration;
```
kriswest, line 83: In the previous pattern a generic was used in the ServerContext to allow implementations to customise the registration - that felt better than having overrides.

toolbox/fdc3-for-web/demo/src/client/da/DemoFDC3ServerInstance.ts
```diff
@@ -219,7 +229,7 @@ export class DemoServerContext implements ServerContext<DemoAppRegistration> {
           state: State.Pending,
         };
 
-        this.setInstanceDetails(instanceId, metadata);
+        this.setInstanceDetails(instanceId, metadata as AppRegistration);
```
kriswest, line 245: Type safety vs. overrides again

toolbox/fdc3-for-web/demo/src/client/da/DemoFDC3ServerInstance.ts
```diff
@@ -280,3 +288,27 @@ export class DemoServerContext implements ServerContext<DemoAppRegistration> {
...
+    heartbeats: boolean = true,
+    intentTimeoutMs: number = 20000,
+    openHandlerTimeoutMs: number = 10000
+  ) {
+    super(directory, channels, heartbeats, intentTimeoutMs, openHandlerTimeoutMs);
+    this.socket = socket;
+  }
+
+  createInstance(): FDC3ServerInstance {
+    return new DemoFDC3ServerInstance(this.socket, this.directory, this.handlers, this.channels);
```
kriswest, line 325: Example of this not providing any tangible benefit - its just separating default values and behaviours over 3 files making them harder to interpret. I'm not seeing an advantage, just extra complexity.

toolbox/fdc3-for-web/fdc3-web-impl/src/FDC3ServerInstance.ts
kriswest: I appreciate that the demo server can start multiple instances, which recycle the same socket.io server, with different instance IDs, but that feels like an oddity rather than a pattern; it probably doesn't make sense for end-users to have multiple different instances of a DA that are each an island of interop. I imagine most implementors will find a way to have multiple windows talk to the same Desktop Agent.

Server/Client terminology is perhaps also not helping here (too many servers... and this one is actually running in a browser window). I wonder if something like DesktopAgentCore or DesktopAgentBackend would be easier to understand/intuit?

toolbox/fdc3-for-web/fdc3-web-impl/src/FDC3ServerInstance.ts
```diff
@@ -0,0 +1,312 @@
...
+
+export type ContextListenerRegistration = {
+  appId: string;
+  instanceId: string;
+  listenerUuid: string;
+  channelId: string | null;
+  contextType: string | null;
+};
+
+export type PrivateChannelEventListener = {
```
kriswest: To be consistent with the others, this type and DesktopAgentEventListener should end in Registration

```suggestion
export type PrivateChannelEventListenerRegistration = {
```

toolbox/fdc3-for-web/fdc3-web-impl/src/FDC3ServerInstance.ts
```diff
@@ -0,0 +1,312 @@
...
+   * Returns the list of apps open and connected to FDC3 at the current time.
+   * Note, it is the implementor's job to ensure this list is
+   * up-to-date in the event of app crashes or disconnections.
+   */
+  getConnectedApps(): Promise<AppRegistration[]>;
+
+  /**
+   * Return the list of all apps that have ever been registered with the ServerContext.
+   */
+  getAllApps(): Promise<AppRegistration[]>;
```
kriswest: To avoid confusion with retrieving app details bring 'Instance' or 'Registration' into the name:

```suggestion
  getAllAppRegistrations(): Promise<AppRegistration[]>;
```

toolbox/fdc3-for-web/fdc3-web-impl/src/FDC3ServerInstance.ts
```diff
@@ -0,0 +1,312 @@
...
+
+  /**
+   * Returns the list of apps open and connected to FDC3 at the current time.
+   * Note, it is the implementor's job to ensure this list is
+   * up-to-date in the event of app crashes or disconnections.
+   */
+  getConnectedApps(): Promise<AppRegistration[]>;
+
+  /**
+   * Return the list of all apps that have ever been registered with the ServerContext.
```
kriswest: I'd also consider updating this comment to remove 'ever been' as a long running implementation might drop older registrations, or ones it knows are gone forever (say a window closed so there no way its coming back as the same instance ever again - the instance details in SessionStorage dies with the window). Also clarify that its instances/registrations rather than app definitions.
```suggestion
   * Return the list of all app instances that have been registered with the ServerContext.
```

toolbox/fdc3-for-web/fdc3-web-impl/src/FDC3ServerInstance.ts
```diff
@@ -0,0 +1,312 @@
...
+   * The server instance can decide how to respond to different heartbeat events.
+   */
+  heartbeatActivity(instanceId: InstanceID, event: HeartbeatActivityEvent): Promise<void>;
+
+  /**
+   * Returns the list of apps open and connected to FDC3 at the current time.
+   * Note, it is the implementor's job to ensure this list is
+   * up-to-date in the event of app crashes or disconnections.
+   */
+  getConnectedApps(): Promise<AppRegistration[]>;
```
kriswest:
```suggestion
  getConnectedAppRegistrations(): Promise<AppRegistration[]>;
```

See later comment `getAllApps()`, rename will help comprehension. Mmatters less here because of the 'connected', but we should be consistent.

kriswest (commented):
Apparently I abandoned a review last year - closing it to drop comments I'd left you @robmoffat

PR #1717, opened 2025-12-10
Configure environment variables for CVE scanning

TheJuanAndOnly99:
Added environment variables for OSS_INDEX_USERNAME and OSS_INDEX_TOKEN in CVE scanning workflow. The secrets have also been added to the repo.

### DIFF
--- a/.github/workflows/cve-scanning.yml
+++ b/.github/workflows/cve-scanning.yml
@@ -44,8 +44,14 @@ jobs:
         working-directory: website
 
       - run: npx --yes auditjs ossi --whitelist allow-list.json
+        env:
+          OSS_INDEX_USERNAME: ${{ secrets.OSS_INDEX_USERNAME }}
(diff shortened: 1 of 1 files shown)


kriswest:
> Hi @kriswest yes. The problem is that forks cannot access secrets.

So it had to be merged before it would work in the scheduled scans - next one happens overnight. Will it be able to scan PRs (so we can use it as a check)?

TheJuanAndOnly99:
@kriswest I sent you another PR that fixes it https://github.com/finos/FDC3/pull/1721 (it fails on CVEs found). Unfortunately it will still not work to scan PRs coming from forks as they can't access the secrets. Not sure how we can work around that securely yet.

thomassalmon422-a11y:
[CCE_Licensing_Agreement_Hardened.pdf](https://github.com/user-attachments/files/24535822/CCE_Licensing_Agreement_Hardened.pdf)

PR #1742, opened 2026-01-29
Update channelChangedEvent.schema.json for backwards compatibility

kriswest:
## Describe your change

Example schema change to support either the old property or the one standardised to match the FDC3 Standard event (so it doesn't have to be translated).

Types still need to be generated
### Related Issue

#1585

### DIFF
--- a/packages/fdc3-schema/schemas/api/channelChangedEvent.schema.json
+++ b/packages/fdc3-schema/schemas/api/channelChangedEvent.schema.json
@@ -30,24 +30,49 @@
 		"ChannelChangedEventPayload": {
 			"title": "channelChanged Event Payload",
 			"type": "object",
-			"properties": {
-				"newChannelId": {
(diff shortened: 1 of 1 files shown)


kriswest:
> @kriswest I rebuilt but the ts is still wrong:

Sorry I hadn't touched that. Was on a train when I raised this to show what I thought the schema needed. `Thanks for moving it onwards.

I think the failing test was something else:

```
> @finos/fdc3-agent-proxy@2.2.1-beta.4 test
> tsc && nyc --reporter=lcov --reporter=text  --reporter json  cucumber-js

Error: src/listeners/DesktopAgentEventListener.ts(18,11): error TS2322: Type 'string | null | undefined' is not assignable to type 'string | null'.
  Type 'undefined' is not assignable to type 'string | null'.
```

Thats usually something to do with an optional property in a type (which changes the type to include undefined)

PR #1748, opened 2026-01-30
fix the progress reporting in conformance tests

kriswest:
Fixes the progress reporting in conformance tests by adjusting how the test suites are assembled.

The channel tests were being assembled asynchronously, but not being awaited properly leading to their assembly happening after the runner had started, causing it to miscalculate the progress %

After this PR the reference implementation will produce:
image


kriswest:
@robmoffat The problem with the progress calculation is that a couple of the test suite's setup functions (specifically app channels and user channels) were changed to be async (some time before migration into the main FDC3 repo), these were not awaited/awaitable because of the incorrect typing. Theres also a small race there (that I think we always win as far as execution goes, but always lose as far as progress calculation goes) as the tests are still being defined when they start running.

Fixing the issue lead to changing the type for the test suites and making all the suite functions async and awaiting them all. That ensures that they are fully created before test execution begins and voila, correct progress calculation. I also separated those two suites back out from each other and matched them up to the style of the others for cleanliness.

An alternative solution is to make the test creation synchronous by fixing those two test suites. I'm not sure if you can achieve that as ChannelControl takes a reference to the desktop agent retrieved with getAgent... perhaps by moving the getAgent call into Channel Control? If you want to take on doing that instead, be my guest - this PR can be considered research into the cause. However, also happy to see this just merged and the issue resolved as I think its a valid solution.

PR #1810, opened 2026-03-27
fdc3.addContextListener() accepts array type

### LINKED ISSUE #1646: Suggestion: fdc3.addContextListener() should accept an array of contextTypes to listen for
## Enhancement Request

`fdc3.addContextListener(type, handler)` method should offer an option to pass an array of types the client is interested in.

### Use Case:
A lot of our apps at LSEG are able to handle both `fdc3.instrument` and `fdc3.instrumentList`. As such, our apps add a handler for `null` to handle any of the two  incoming context types. But given the example below, sometimes it is not possible to determine what is the latest between the two when an app is joining a channel.

LSEG is moving forward with the approach of offering to run our apps in our customers' platforms / Desktop Agents. 
Therefore we need to have a standardized way of ensuring we get the latest context between a subset of types when the app joins a channel to ensure consistency with the rest of the apps in the platform. 

For example: 

```txt
11:00 App A and App B both join channel green
11.01 App A broadcasts an instrument
11.02 App A broadcasts an instrumentList
11.03 App B broadcasts a contact
11.04 App C joins channel green
```

App C only supports instrument and instrumentList - but cotnact being the latest,  how can App C determine that instrumentList was sent after instrument in this case?

### Workflow Description
App C would add a context listener such as:

```ts
    fdc3.addContextListener(['fdc3.instrument','fdc3.instrumentList'], (ctx) => { console.log(ctx); });
```

And, given the above example, when App C joins the channel, it would log `instrumentList` since it is the latest between 2 types in the channel.

kemerava:
## Describe your change

Handing the addContextListener with array type: if an array is passed, then the handler will be call for all the types in the array.

Seeking opinion: I have added logic such that if there is a null passed to the array, then it is behaving the same way as passing null into addContextListner(null) -> listen to all. But if an empty array is passed in, then nothing is listened on. Also the implementation of the agent proxy made me think about some potential questions, like would it be possible to unsubscribe somehow just from one of the context types?

### Related Issue
resolves #1646 

## Review Checklist


kriswest:
Needs the conformance tests implementing

packages/fdc3-agent-proxy/src/channels/ChannelSupport.ts
```diff
@@ -7,7 +7,7 @@ export interface ChannelSupport {
   createPrivateChannel(): Promise<PrivateChannel>;
   leaveUserChannel(): Promise<void>;
   joinUserChannel(id: string): Promise<void>;
-  addContextListener(handler: ContextHandler, type: string | null): Promise<Listener>;
+  addContextListener(handler: ContextHandler, type: string | null | (string | null)[]): Promise<Listener>;
```
kriswest: I think this should be:

```suggestion
  addContextListener(handler: ContextHandler, type: string | string[] | null): Promise<Listener>;
```

if you are registering a null listener do you need to register others? Unlike intent listeners multiple context listeners can fire for a single message - but is there any point in allowing that here as they'll all have the same handler so it will always end up being being double processing with the same handler...

kriswest: Commenting on first occurrance only, but needs fixing throughout

packages/fdc3-schema/schemas/api/addContextListenerRequest.schema.json
```diff
@@ -54,11 +54,35 @@
...
+					"description": "Array of context types to listen for. May contain `null` to listen to all context types in addition to specific types. When `null` is present, other context types are ignored as the listener will receive all context types.",
+					"type": "array",
+					"items": {
+						"oneOf": [
+							{
+								"type": "string"
+							},
+							{
+								"type": "null"
+							}
```
kriswest: reduce to just array of string (as per earlier comment about not needing to mix null with an array of strings

kriswest (changes requested):
Drop null from the array - see comment inline for why it doesn't make sense.

kriswest:
Has picked up some small conflicts and needs the above issue resolving to merge. A separate issue can be raised for implementing the conformance tests

packages/fdc3-agent-proxy/src/channels/ChannelSupport.ts
kemerava: Resolved, thank you!

packages/fdc3-schema/schemas/api/addContextListenerRequest.schema.json
kemerava: Done, thank you!

kemerava:
> Has picked up some small conflicts and needs the above issue resolving to merge. A separate issue can be raised for implementing the conformance tests

Thank you for the review, @kriswest  - I have addressed your comments and resolved merge conflicts, please review! Thank you!!

packages/fdc3-agent-proxy/src/apps/DefaultAppSupport.ts
```diff
@@ -140,6 +140,7 @@ export class DefaultAppSupport implements AppSupport {
         provider: 'unknown',
         appMetadata: { appId: 'unknown', instanceId: 'unknown' },
         optionalFeatures: {
+          OriginatingAppMetadata: false,
```
kriswest, line 143: This should not be restored by the PR as its now a requirement (in the metadata object). PR may need a rebase?

packages/fdc3-agent-proxy/src/channels/DefaultChannel.ts
```diff
@@ -121,17 +121,49 @@ export class DefaultChannel implements Channel {
...
+      theContextType = contextTypeOrHandler;
+      theHandler = handler;
+    } else if (Array.isArray(contextTypeOrHandler) && typeof handler === 'function') {
+      // Handle array-based context types
+      if (contextTypeOrHandler.length === 0) {
+        // Empty array
+        return {
+          unsubscribe: () => Promise.resolve(),
+        };
+      }
```
kriswest, line 144: I wonder if we'd be better off returning `ChannelError.InvalidArguments` here. Do we need to support listening for nothing?

packages/fdc3-agent-proxy/src/channels/DefaultChannel.ts
```diff
@@ -121,17 +121,49 @@ export class DefaultChannel implements Channel {
...
+      theContextType = contextTypeOrHandler; // Pass the array directly
+      theHandler = handler;
+    } else if (typeof contextTypeOrHandler === 'function') {
+      // deprecated one-arg version
+      theContextType = null;
+      theHandler = contextTypeOrHandler as ContextHandler;
+    } else {
+      //invalid call
+      // TODO: Replace with Standardized error when #1490 is resolved
       throw new Error('Invalid arguments passed to addContextListener!');
```
kriswest: Not something you changed, but needs doing:

```suggestion
      throw new Error(ChannelError.InvalidArguments);
```

packages/fdc3-agent-proxy/src/channels/DefaultChannel.ts
```diff
@@ -121,17 +121,49 @@ export class DefaultChannel implements Channel {
...
+        return {
+          unsubscribe: () => Promise.resolve(),
+        };
+      }
+      theContextType = contextTypeOrHandler; // Pass the array directly
+      theHandler = handler;
+    } else if (typeof contextTypeOrHandler === 'function') {
+      // deprecated one-arg version
+      theContextType = null;
+      theHandler = contextTypeOrHandler as ContextHandler;
```
kriswest, line 150: This was removed already in an earlier PR and shouldn't be restored I think

packages/fdc3-agent-proxy/src/channels/DefaultChannel.ts
```diff
@@ -121,17 +121,49 @@ export class DefaultChannel implements Channel {
     return null;
   }
 
-  async addContextListener(contextType: string | null, handler: ContextHandler): Promise<Listener> {
-    if (typeof contextType !== 'string' && contextType !== null) {
+  async addContextListener(
+    contextTypeOrHandler: string | null | ContextHandler | string[],
+    handler?: ContextHandler
+  ): Promise<Listener> {
```
kriswest, line 127: I think because this PR was started before other changes were made for 3.0 its ended up as a mishmash of the old handling code and new... I think this another look as it should be much shorter than this,.

kriswest: I.e. contextTypeOrHandler is no longer a thing and after validating args you can go straight to addContextListenerInner.

packages/fdc3-agent-proxy/src/channels/DefaultChannel.ts
```diff
@@ -121,17 +121,49 @@ export class DefaultChannel implements Channel {
...
+      // TODO: Replace with Standardized error when #1490 is resolved
       throw new Error('Invalid arguments passed to addContextListener!');
     }
-    if (typeof handler !== 'function') {
-      throw new Error('Invalid arguments passed to addContextListener!');
+
+    // Handle array case by creating individual listeners for each type
+    if (Array.isArray(theContextType)) {
+      // Pass the array directly to DefaultContextListener which will handle multiple listeners
+      return await this.addContextListenerInner(theContextType, theHandler);
```
kriswest, line 160: Doesn't need it own clause any more - unless you move validation that at least one contextType is listened for here.

packages/fdc3-agent-proxy/src/listeners/DefaultContextListener.ts
```diff
@@ -10,37 +10,65 @@ export class DefaultContextListener
...
-      'addContextListenerResponse',
-      'contextListenerUnsubscribeRequest',
-      'contextListenerUnsubscribeResponse'
-    );
+    // For arrays, use the contextTypes field in the payload
+    if (Array.isArray(contextType)) {
+      // Handle empty array case - will be caught upstream and return dummy listener
+      if (contextType.length === 0) {
+        throw new Error('Empty arrays should be handled upstream');
+      }
```
kriswest, line 28: As per previous comment, unless theres a use case I'm not considering I think we can eliminate this case by rejecting it as an invalid argument.

packages/fdc3-agent-proxy/src/listeners/DefaultContextListener.ts
```diff
@@ -10,37 +10,65 @@ export class DefaultContextListener
...
-      m.payload.channelId == this.channelId &&
-      (m.payload.context?.type == this.contextType || this.contextType == null)
-    );
+    // Handle array context types in filtering
+    let contextTypeMatch = false;
+    if (Array.isArray(this.contextType)) {
+      // For arrays, match if any type in the array matches
+      contextTypeMatch = this.contextType.includes(m.payload.context?.type ?? '');
+    } else {
+      // Single context type - use original logic
```
kriswest, line 67:
```suggestion
      // Single context type
```

packages/fdc3-agent-proxy/src/listeners/DefaultContextListener.ts
```diff
@@ -10,37 +10,65 @@ export class DefaultContextListener
...
+        messageExchangeTimeout,
+        { channelId, contextType: undefined, contextTypes: contextType },
+        handler,
+        'addContextListenerRequest',
+        'addContextListenerResponse',
+        'contextListenerUnsubscribeRequest',
+        'contextListenerUnsubscribeResponse'
+      );
+    } else {
+      // Single context type - use original logic
```
kriswest, line 42:
```suggestion
      // Single context type
```

packages/fdc3-agent-proxy/src/DesktopAgentProxy.ts
```diff
@@ -91,14 +92,40 @@ export class DesktopAgentProxy implements DesktopAgent, Connectable {
     }
   }
 
-  addContextListener(contextType: string | null, handler: ContextHandler): Promise<Listener> {
-    if (typeof contextType !== 'string' && contextType !== null) {
-      throw new Error('Invalid arguments passed to addContextListener!');
-    }
-    if (typeof handler !== 'function') {
+  addContextListener(
+    contextTypeOrHandler: ContextHandler | ContextType | null | ContextType[],
```
kriswest, line 96:
```suggestion
    contextTypeOrHandler: ContextHandler | ContextType | ContextType[] | null,
```

packages/fdc3-agent-proxy/src/DesktopAgentProxy.ts
```diff
@@ -91,14 +92,40 @@ export class DesktopAgentProxy implements DesktopAgent, Connectable {
...
+      theContextType = contextTypeOrHandler;
+      theHandler = handler;
+    } else if (Array.isArray(contextTypeOrHandler) && typeof handler === 'function') {
+      // Handle array-based context types
+      if (contextTypeOrHandler.length === 0) {
+        // Empty array
+        return Promise.resolve({
+          unsubscribe: () => Promise.resolve(),
+        });
+      }
```
kriswest, line 115: as per previous comment we could prevent this and simply return `ChannelError.InvalidArguments`

packages/fdc3-agent-proxy/src/DesktopAgentProxy.ts
```diff
@@ -91,14 +92,40 @@ export class DesktopAgentProxy implements DesktopAgent, Connectable {
...
+      theContextType = contextTypeOrHandler; // Pass the array directly
+      theHandler = handler;
+    } else if (typeof contextTypeOrHandler === 'function') {
+      // deprecated one-arg version
+      theContextType = null;
+      theHandler = contextTypeOrHandler as ContextHandler;
```
kriswest, line 121: This was removed previously and should not come back

packages/fdc3-agent-proxy/src/DesktopAgentProxy.ts
```diff
@@ -91,14 +92,40 @@ export class DesktopAgentProxy implements DesktopAgent, Connectable {
     }
   }
 
-  addContextListener(contextType: string | null, handler: ContextHandler): Promise<Listener> {
-    if (typeof contextType !== 'string' && contextType !== null) {
-      throw new Error('Invalid arguments passed to addContextListener!');
-    }
-    if (typeof handler !== 'function') {
+  addContextListener(
```
kriswest, line 95: This whole function has not merged well with other changes and can/should be simplified

website/docs/api/ref/DesktopAgent.md
```diff
@@ -180,6 +184,8 @@ func (desktopAgent *DesktopAgent) AddContextListener(contextType string, handler
 
 Adds a listener for incoming context broadcasts from the Desktop Agent (via a User channel or [`fdc3.open`](#open) API call). If the consumer is only interested in a context of a particular type, they can specify that type. If the consumer is able to receive context of any type or will inspect types received, then they can pass `null` as the `contextType` parameter to receive all context types.
 
+Alternatively, you can pass an array of context types to listen for multiple specific types at once. If the array contains `null`, it will be treated as if `null` was passed directly (meaning listen to all context types), so any other context types are ignored as the listener will receive all context types. Empty arrays will be ignored.
```
kriswest, line 187: We dropped the nulls in the array - and may change to throwing an error on empty arrays, which will need updating here.

website/docs/api/ref/Channel.md
```diff
@@ -234,6 +239,8 @@ func (ch *Channel) AddContextListener(contextType string, handler ContextHandler
 
 Adds a listener for incoming contexts of the specified _context type_ whenever a broadcast happens on this channel.
 
+Alternatively, you can pass an array of context types to listen for multiple specific types at once. If the array contains `null`, it will be treated as if `null` was passed directly (meaning listen to all context types), so any other context types are ignored as the listener will receive all context types. Empty arrays will be ignored.
```
kriswest, line 242: We dropped the nulls in the array - and may change to throwing an error on empty arrays, which will need updating here.

kriswest (changes requested):
There are some merge errors here in that the handling for addContextListener has reverted to what it was when this PR was started - later changes simplified it massively after the deprecated form was removed for 3.0. Hence, DesktopAgent.addContextListener and Channel.addContextListener need another look. Also docs are out of sync with implementation.

Have not yet had a look at tests (defs or implementation).

I think we should throw a ChannelError.InvalidArguments if someone passes an empty array as an arg or anything in the array thats not a string. What do others think @finos/fdc3-maintainers ?

packages/fdc3-agent-proxy/src/channels/DefaultChannel.ts
```diff
@@ -121,17 +121,49 @@ export class DefaultChannel implements Channel {
...
+      //invalid call
+      // TODO: Replace with Standardized error when #1490 is resolved
       throw new Error('Invalid arguments passed to addContextListener!');
     }
-    if (typeof handler !== 'function') {
-      throw new Error('Invalid arguments passed to addContextListener!');
+
+    // Handle array case by creating individual listeners for each type
+    if (Array.isArray(theContextType)) {
+      // Pass the array directly to DefaultContextListener which will handle multiple listeners
```
kriswest, line 159: Could validate all elements are strings and throw an error if not.

` throw new Error(ChannelError.InvalidArguments);`

packages/fdc3-agent-proxy/src/DesktopAgentProxy.ts
```diff
@@ -91,14 +92,40 @@ export class DesktopAgentProxy implements DesktopAgent, Connectable {
...
+      theHandler = handler;
+    } else if (Array.isArray(contextTypeOrHandler) && typeof handler === 'function') {
+      // Handle array-based context types
+      if (contextTypeOrHandler.length === 0) {
+        // Empty array
+        return Promise.resolve({
+          unsubscribe: () => Promise.resolve(),
+        });
+      }
+      theContextType = contextTypeOrHandler; // Pass the array directly
```
kriswest, line 116: Could validate all elements are strings and throw an error if not.

` throw new Error(ChannelError.InvalidArguments);`

packages/fdc3-agent-proxy/src/DesktopAgentProxy.ts
```diff
@@ -91,14 +92,40 @@ export class DesktopAgentProxy implements DesktopAgent, Connectable {
...
+      theHandler = handler;
+    } else if (typeof contextTypeOrHandler === 'function') {
+      // deprecated one-arg version
+      theContextType = null;
+      theHandler = contextTypeOrHandler as ContextHandler;
+    } else {
+      //invalid call
+      // TODO: Replace with Standardized error when #1490 is resolved
       throw new Error('Invalid arguments passed to addContextListener!');
```
kriswest:
```suggestion
      throw new Error(ChannelError.InvalidArguments);
```

kriswest:
@kemerava this is failing CI tests but I can't actually see a test failure... Re-running in the hopes that it clears...

packages/fdc3-agent-proxy/src/apps/DefaultAppSupport.ts
kriswest: This still needs dealing with

kriswest:
Review comments still need dealing with from the earlier bad merge/rebase

kriswest:
Getting some honest test failures this time:

> ⎯⎯⎯⎯⎯⎯⎯ Failed Tests 1 ⎯⎯⎯⎯⎯⎯⎯
> 
>  FAIL  test/features/app-channels.feature:177:3 > Feature: Channel Listeners Support > Scenario: Adding a context listener with an empty array returns a dummy listener
> AssertionError: Scenario finished with 1 errors:
> 
> "{result}" is undefined (#156)
> expected TypeError: Cannot read properties of unde… to be undefined
> 
> - Expected:
> undefined
> 
> + Received:
> TypeError {
>   "message": "Cannot read properties of undefined (reading 'call')",
> }
> 
>  ❯ test/features/app-channels.feature:156:1
>     154|     Then "{contexts}" is empty
>     155|     And I call "{theListener}" with "unsubscribe"
>     156|     Then "{result}" is undefined
>        | ^
>     157|
> 
> ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯[1/1]⎯

And some fails in fdc3-security and fdc3-web-impl relating to the new function signatures that need looking at @kemerava

kemerava:
@kriswest  please review - the build is passing now too

packages/fdc3-agent-proxy/src/channels/DefaultChannel.ts
```diff
@@ -124,17 +124,28 @@ export class DefaultChannel implements Channel {
...
 
-  async addContextListener(contextType: string | null, handler: ContextHandler): Promise<Listener> {
-    if (typeof contextType !== 'string' && contextType !== null) {
-      throw new Error('Invalid arguments passed to addContextListener!');
-    }
+  addContextListener(contextType: string | null, handler: ContextHandler): Promise<Listener>;
+  addContextListener(contextTypes: string[], handler: ContextHandler): Promise<Listener>;
+  async addContextListener(typeOrTypes: string | string[] | null, handler: ContextHandler): Promise<Listener> {
     if (typeof handler !== 'function') {
       throw new Error('Invalid arguments passed to addContextListener!');
```
kriswest:
```suggestion
      throw new Error(ChannelError.InvalidArguments);
```

packages/fdc3-agent-proxy/src/channels/DefaultChannel.ts
```diff
@@ -124,17 +124,28 @@ export class DefaultChannel implements Channel {
...
+      }
+      return await this.addContextListenerInner(typeOrTypes, handler);
     }
-    return await this.addContextListenerInner(contextType, handler);
+
+    if (typeof typeOrTypes === 'string' || typeOrTypes === null) {
+      return await this.addContextListenerInner(typeOrTypes as string | null, handler);
+    }
+
+    throw new Error('Invalid arguments passed to addContextListener!');
```
kriswest:
```suggestion
    throw new Error(ChannelError.InvalidArguments);
```

packages/fdc3-agent-proxy/src/channels/DefaultChannelSupport.ts
```diff
@@ -322,7 +323,17 @@ export class DefaultChannelSupport implements ChannelSupport, Connectable {
...
+          if (Array.isArray(this.contextType)) {
+            // Don't filter by type - let the individual listeners handle their own filtering
+            contextTypeParam = undefined;
+          } else if (this.contextType != null) {
+            contextTypeParam = this.contextType;
+          }
+
+          const result = await channel.getCurrentContextWithMetadata(contextTypeParam);
           if (result) {
             this.handler(result.context, result.metadata);
```
Roaders: Cached-context replay bypasses the array filter here. With a listener for `['fdc3.instrument']`, registering on or joining a channel whose latest context is `fdc3.contact` calls the handler with that contact: the lookup is unfiltered and the handler is invoked directly, without going through `filter()`. It also misses matching older cached context. Could the replay lookup respect the requested types? I reproduced the incorrect contact delivery using this method.

packages/fdc3-agent-proxy/src/channels/DefaultChannel.ts
```diff
@@ -124,17 +124,28 @@ export class DefaultChannel implements Channel {
...
     if (typeof handler !== 'function') {
-      throw new Error('Invalid arguments passed to addContextListener!');
+      throw new Error(ChannelError.InvalidArguments);
+    }
+
+    if (Array.isArray(typeOrTypes)) {
+      if (typeOrTypes.length === 0) {
+        return { unsubscribe: () => Promise.resolve() };
+      }
+      return await this.addContextListenerInner(typeOrTypes, handler);
```
Roaders, line 146: Arrays containing `null` are not normalized here as they are in `DesktopAgentProxy`. On app/private channels, `['fdc3.instrument', null]` reaches `DefaultContextListener`, whose `.includes(context.type)` filter rejects contacts rather than receiving all types as documented. I reproduced this filtering failure. The request schema also permits only strings in `contextTypes`. Could arrays containing `null` be normalized to an unfiltered subscription before constructing the listener?

toolbox/fdc3-for-web/fdc3-web-impl/src/handlers/BroadcastHandler.ts
```diff
@@ -483,7 +483,7 @@ export class BroadcastHandler implements MessageHandler {
       instanceId: from.instanceId ?? 'no-instance-id',
       channelId: channelId,
       listenerUuid: sc.createUUID(),
-      contextType: arg0.payload.contextType,
+      contextType: arg0.payload.contextType ?? null,
```
Roaders: Array requests populate `payload.contextTypes` and omit `contextType`, so this records every array subscription as unfiltered. That causes unrelated broadcasts to be sent to the client and loses the requested types in private-channel add-listener/unsubscribe events, which producers use to manage subscriptions. Client-side filtering hides the broadcast issue from ordinary handlers. `ClientSideHandlersImpl.handleAddContextListener` similarly ignores `contextTypes` and subscribes with `null`. Could both handlers preserve the array subscription semantics, including private-channel notifications?

packages/fdc3-standard/src/api/DesktopAgent.ts
```diff
@@ -480,6 +480,16 @@ export interface DesktopAgent {
...
 
+  /**
+   * Adds a listener for incoming contexts using an array of context types.
+   *
+   * Pass multiple context types to listen for several types with one listener.
+   *
+   * See the single `contextType` overload above for full behavior details.
+   *
+   */
+  addContextListener(contextTypes: ContextType[], handler: ContextHandler): Promise<Listener>;
```
Roaders, line 493: The documented `addContextListener(['fdc3.instrument', null], handler)` call is rejected by this public overload under strict null checking because `ContextType[]` excludes `null`. The new `Channel` overload has the same problem with `string[]`. The conformance tests work around it with `as unknown as string[]`, concealing the public API mismatch. Could both overloads accept nullable array elements and the tests use them without casts?

Roaders (commented):
Reviewed the array-based `addContextListener` changes. Four issues are noted inline.

Additional test gap: `ACArrayContextListeners1` and `UCArrayContextListeners1` use broadcast sequences containing only the two matching types. They therefore cannot verify the documented rejection of an unrelated type such as `fdc3.portfolio`; please include a non-matching broadcast.

Validation: source review and focused reproductions of the cached-context replay and null-array filtering issues; the full suite was not run.

toolbox/fdc3-for-web/fdc3-web-impl/src/handlers/BroadcastHandler.ts
kriswest: This is not handling the 'contextTypes' array argument in teh payload that you added in the addContextListenerRequest schema... which also means this can't be covered correctly by a unit test

packages/fdc3-security/src/secure-boundary/ClientSideHandlersImpl.ts
```diff
@@ -143,7 +143,7 @@ export class ClientSideHandlersImpl implements FDC3Handlers {
     const id = this.messaging.createUUID();
     if (channel) {
       const cl = await channel.addContextListener(
-        acl.payload.contextType,
+        acl.payload.contextType ?? null,
```
kriswest: need to handle the contextTypes (plural) payload field here as well

website/docs/api/ref/Channel.md
kriswest: This still needs resolving

kriswest (changes requested):
@kemerava the conflicts look fixed - many thanks.

I think theres still a gap in the reference implementation where the array type is not being handled and is probably being interpreted as a null (all types listener) as it didn't check the plural version of teh argument. If true that means theres a unit test in fdc3-web-impl that will need to catch the difference between an array lisener and an all listener.

also theres a documentation fix previously identified that still needs updating.

At today's meeting we identified that this could be bumped to 3.1 if its not ready soon as its an additive change. I'd rather get it closed and merged - but the decision on that is yours @kemerava as you are holding the pen/keyboard ;-)

packages/fdc3-agent-proxy/src/channels/DefaultChannel.ts
kemerava: Removed allowing null in the array

packages/fdc3-standard/src/api/DesktopAgent.ts
kemerava: Updated and fixed the tests

toolbox/fdc3-for-web/fdc3-web-impl/src/handlers/BroadcastHandler.ts
kemerava: Updated here and the private channels

packages/fdc3-agent-proxy/src/channels/DefaultChannelSupport.ts
kemerava: Fixed the replay issue, should not receive the `fdc3.contact` if a listener is for `['fdc3.instrument']`. Thank you

toolbox/fdc3-for-web/fdc3-web-impl/src/handlers/BroadcastHandler.ts
kemerava: Updated, some of the tests were outdated as you said

packages/fdc3-security/src/secure-boundary/ClientSideHandlersImpl.ts
kemerava: Done

kemerava:
@Roaders and @kriswest - thank you for the review, please take a look now

website/docs/api/ref/DesktopAgent.md
kemerava: I added the error for an empty array

kriswest (dismissed):
LGTM after teh null handling change @kemerava - is this now ready to merge in your opinion (i.e. are you done? asking as updated 1 min ago)

Roaders (dismissed):

kemerava:
> LGTM after teh null handling change @kemerava - is this now ready to merge in your opinion (i.e. are you done? asking as updated 1 min ago)

Thanks for the review @kriswest - just was fixing some of the tests, all is finished now

kriswest (approved):
LGTM

PR #1817, opened 2026-04-01
Add basic conformance tests for addEventListener

### LINKED ISSUE #1774: Create Basic conformance test for addEventListener
There is currently no conformance test that ensures that a given DesktopAgent implements the addEventListener function. This is defined within the required section of the [spec](https://fdc3.finos.org/docs/api/spec#desktop-agent-api-standard-compliance).

- Ensure that the function exists and is callable
- Ensure that a `Listener` object is returned

Maximiliano-Zeballos:
## Describe your change

Add basic conformance tests for `addEventListener` as described in #1774.

Two tests following the existing pattern used for `addContextListener` (BasicCL1/BasicCL2):

- **BasicAEL1**: Verifies `addEventListener` is callable with a specific event type (`userChannelChanged`) and returns a valid `Listener` object with an `unsubscribe` function.
- **BasicAEL2**: Verifies the same behavior when passing `null` as the event type.

Also adds the `addEventListener` documentation link to `apiDocuments.ts`.

### Related Issue

Resolves #1774

## Review Checklist

- [ ] **API changes**: Does this PR include changes to any of the FDC3 APIs? — *No, only conformance tests.*

### DIFF
--- a/toolbox/fdc3-conformance/src/test/basic/fdc3.basic.ts
+++ b/toolbox/fdc3-conformance/src/test/basic/fdc3.basic.ts
@@ -1,4 +1,4 @@
-import { Context, DesktopAgent } from '@finos/fdc3';
+import { Context, DesktopAgent, FDC3Event } from '@finos/fdc3';
 
 import { APIDocumentation } from '../support/apiDocuments';
 import { ContextType, Intent } from '../support/intent-support';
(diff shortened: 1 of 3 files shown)


kriswest (changes requested):
The test implementations look fine to me - but we need the definitions also added to the standard's docs at https://github.com/Maximiliano-Zeballos/FDC3/blob/feature/add-event-listener-conformance-test/website/docs/api/conformance/Basic-Tests.md

kriswest (changes requested):
Hi @Maximiliano-Zeballos, many thanks for your contribution. Please note that we'll need you to:

- rebase against main
- Add a description of the new case in the docs at: https://github.com/Maximiliano-Zeballos/FDC3/blob/feature/add-event-listener-conformance-test/website/docs/api/conformance/Basic-Tests.md
- add an entry in CHANGELOG.md describing the change

toolbox/fdc3-conformance/src/test/basic/fdc3.basic.ts
```diff
@@ -82,6 +82,41 @@ const basicGI1 = (fdc3: DesktopAgent, documentation: string) => {
   });
 };
 
+const basicAEL1 = (fdc3: DesktopAgent, documentation: string) => {
+  it('(BasicAEL1) Method is callable', async () => {
```
Roaders, line 86: perhaps I am misreading this but this doesn't seem like a good test description. Surely it should be `addEventListener Method is callable`. I suspect that there is probably the equivalent of a describe block that this test will appear within so it will be ok but right now I can't see that.

kriswest:
This could do with a rebase and the checks should start passing. We also need to ensure that the test definitions in the docs at https://github.com/finos/FDC3/blob/main/website/docs/api/conformance/Basic-Tests.md are updated so they stay in sync with the implementation.

Deal with that and @Roaders' comment and this should be mergeable! Many thanks for taking it on

PR #1823, opened 2026-04-02
Fix linting configs, fdc3-context and fdc3-conformance builds and reduce lint

### LINKED ISSUE #1828: Joining a user channel then adding a context listener should replay current context on the channel
There is a race condition in the fdc3 workbench, which is receiving context from the channel (when it already exists and the app is already joined to the channel before the listener is added) before it has updated its internal state for which listeners have been added.

Original description:
> The FDC3 reference implementation can be demonstrated as not replaying the current context (not calling the registered context handler) if you join a channel and then add the context listener, but will do if you add the listener and then join the channel, or if you switch channels after adding a listener. 
>
> This is not correct and should already be covered in the conformance tests under `UCBasicUsage4` 
(see: https://fdc3.finos.org/docs/api/conformance/User-Channel-Tests)
>
> However, despite the reference implementation demonstrably failing this test, it is not being picked up by the conformance testing suite, where it passes all tests.
>
> Investigate why this test is not being failed and correct the conformance testing suite, then fix the reference implementation to pass the corrected test.
>
> This should be corrected for the 2.2.2 release.

kriswest:
## Describe your change

- Modernises eslint configs for several packages 
- Sets missing tsconfigRootDir through eslint configs
- Reduces lint in fdc3-workbench
- Fixes a race condition in workbench's context store which was causing the context listeners on user channels to miss replay of current context on the channel in some cases

In the course of testing this I discovered issues in fdc3-context (missing module type causing it to output a commonjs module) and fdc3-conformance (missing extension on a webpack plugin). These are also fixed in this PR and necessary for the FDC3 2.2.2 release.

### Related Issue

N/A

## Review Checklist

- [ ] **Issue**: If a change was made to the FDC3 Standard, was an issue linked above?


packages/fdc3-context/package.json
```diff
@@ -11,6 +11,7 @@
     "tag": "latest"
   },
   "license": "Apache-2.0",
+  "type": "module",
```
Roaders, line 14: I must have missed this one. Sorry about that!

Roaders (dismissed):
LGTM

kriswest:
@Roaders could you take another look at this please - I've contacted FINOS about the netlify build failures, but its otherwise ready to go

robmoffat:
For this I get:

```

> @finos/fdc3-get-agent@2.2.2-beta.1 test
> tsc && vitest run

node:internal/modules/package_json_reader:268
  throw new ERR_MODULE_NOT_FOUND(packageName, fileURLToPath(base), null);
        ^

Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'strip-literal' imported from /Users/rob/Documents/finos/fdc3-general/FDC3/packages/fdc3-get-agent/node_modules/@vitest/runner/dist/utils.js
    at Object.getPackageJSONURL (node:internal/modules/package_json_reader:268:9)
    at packageResolve (node:internal/modules/esm/resolve:768:81)
    at moduleResolve (node:internal/modules/esm/resolve:858:18)
    at defaultResolve (node:internal/modules/esm/resolve:990:11)
    at #cachedDefaultResolve (node:internal/modules/esm/loader:749:20)
    at ModuleLoader.resolve (node:internal/modules/esm/loader:726:38)
    at ModuleLoader.getModuleJobForImport (node:internal/modules/esm/loader:312:38)
    at #link (node:internal/modules/esm/module_job:208:49) {
  code: 'ERR_MODULE_NOT_FOUND'
}

```

kemerava (approved):
LGTM

PR #1858, opened 2026-04-16
Ensure current user channel is resolved before DesktopAgentProxy is used

### LINKED ISSUE #1857: V2.2.2-beta.1 regression: If the desktop agent changes the current user channel the agent-proxy is not aware of the change
The agent-proxy maintains a state variable with the current channel and move user channel context listeners when the channel changes. However, the handling for those changes is implemented in the joinUserChannel function of the agent-proxy:

https://github.com/finos/FDC3/blob/0cf94c30c808bbbee36530b55dbc7e555c4a384e/packages/fdc3-agent-proxy/src/channels/DefaultChannelSupport.ts#L229-L252
(lines 243-251)

This works fine if an app makes a joinUserChannel API call, or if an injected channel selector UI is used, but will fail if the desktop agent moves the app to a different channel. As that is a supported use case and worked in 2.2.0 this is a regression and needs to be resolved.

Desktop Agent implementations might use the a `joinUserChannelRequest` message to trigger the channel change and we could listen, outside of the `joinUserChannel` for it and make the changes, or we might listen for a  `channelChangedEvent` in order to trigger the change - the latter should work for all cases.

kriswest:
## Describe your change

Resolves a regression for web-based Desktop Agents that implement their own user channel management (where an app may start already joined to a channel) by:

- Retrieving the current channel from the Desktop Agent on startup. In prior releases this happened on calling addContextListener, but was removed since.
- Preventing a race condition with addContextListener by making DefaultChannelSupport implement Connectable — moving the getUserChannel() and addEventListener initialization calls from fire-and-forget promises in the constructor into a connect() method that is properly awaited via the DesktopAgentProxy's existing connectables lifecycle.
- Reducing redundant requests for the current channel and user channel list by ensuring the caches for each are always used when populated.
- Adding tests for ChannelSupport cases where the Desktop Agent has preset a user channel

### Related Issue

resolves #1857 

## Review Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,6 +67,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 * Stopped fdc3-workbench flagging FDC3 version 2.2 as unsupported. ([#1841](https://github.com/finos/FDC3/pull/1841))
 * Resolved vulnerable dependencies (esbuild, serialize-javascript, elliptic) and consolidated shared devDependencies to simplify future maintenance. ([#1841](https://github.com/finos/FDC3/pull/1841))
 * Fixed the lack of handling of WCP6Disconnect messages in MessagePort example in the FDC3 Web reference implementation. ([#1854](https://github.com/finos/FDC3/pull/1854))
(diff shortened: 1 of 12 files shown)


kriswest:
@Roaders @julianna-ciq I think this resolves the regression we discussed and fixes an existing race condition. I'll push a v2.2.2-beta.2 with this in to facilitate easier testing.

Roaders (approved):
LGTM

PR #1864, opened 2026-04-23
Improving release workflow to add version sync guardrail and updating instructions

kriswest:
## Describe your change

A recent release (2.2.2) was accidentally tagged as a pre-release in NPM due to the root package still having a beta release label.

This PR updates the release workflow with a new version check script to make sure version numbers are in sync (including root package)

### Related Issue

n/a

## Review Checklist

### DIFF
--- a/scripts/version-check.js
+++ b/scripts/version-check.js
@@ -1,0 +1,21 @@
+const fs = require('fs');
+
+const root = JSON.parse(fs.readFileSync('package.json', 'utf8'));
+const workspaces = root.workspaces || [];
+
+const mismatched = workspaces
+  .map(dir => ({
+    dir,
+    ...JSON.parse(fs.readFileSync(`${dir}/package.json`, 'utf8')),
+  }))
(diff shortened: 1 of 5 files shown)


scripts/version-check.js
```diff
@@ -0,0 +1,21 @@
+const fs = require('fs');
+
+const root = JSON.parse(fs.readFileSync('package.json', 'utf8'));
+const workspaces = root.workspaces || [];
+
+const mismatched = workspaces
+  .map(dir => ({
+    dir,
+    ...JSON.parse(fs.readFileSync(`${dir}/package.json`, 'utf8')),
```
Roaders, line 9: this is fine for our current package.json as we list each workspace individually. If the workspace list was updated to incluse wildcards such as `/projects/*` then this will fail.
I think it's fine for now though. I imagine this script will fail if we use wildcards which is good.

Roaders (approved):

Roaders:
LGTM made a minor comment but I don't think that should stop merging.

PR #1870, opened 2026-04-30
[Bugfix] channel changed event

### LINKED ISSUE #1869: `addEventListener` is invoked with the wrong event type when leaving a channel in 2.2.2
## Minor Issue

Found while testing with `@finos/fdc3` version 2.2.2 (latest published version).
 
When subscribing via `fdc3.addEventListener`, leaving a user channel results in a callback with an incorrect event type (`channelChangedEvent` instead of the expected `userChannelChanged`). Those are the repro steps of my tests:
 
```ts
// Add event listener and log received events. It fails in both cases when I do and do not pass listener type as first arg
fdc3.addEventListener(null, console.warn);

// Join a channel
await fdc3.joinUserChannel("fdc3.channel.1");

/* The handler is invoked with:
{ 
  type: "userChannelChanged",
  details: {
    currentChannelId: "fdc3.channel.1"
  }
}
*/

// Leave the current channel
await fdc3.leaveCurrentChannel();

// The handler is invoked with:
/*
{
  type: "channelChangedEvent",
  newChannelId: null
}
*/
```
 
The issue appears to originate from [this](https://github.com/finos/FDC3/blob/main/packages/fdc3-agent-proxy/src/listeners/DesktopAgentEventListener.ts#L14) code of the handler wrapper. When leaving a channel, the payload contains `newChannelId: null` (since there is no active channel). Because of this, the logic falls in the `else` block, which forwards message coming from the DACP right to the FDC3 API which is not the one [described](https://fdc3.finos.org/docs/api/ref/Events#fdc3channelchangedevent) in the standard.
 
Also, `newChannelId` is deprecated in favor of `currentChannelId` . Both should be taken into consideration but the newer `currentChannelId` field is not being used in [this](https://github.com/finos/FDC3/blob/main/packages/fdc3-agent-proxy/src/listeners/DesktopAgentEventListener.ts#L14) logic.
 
 
@ksgeorgieva suggested the following fix of the `wrapHandler` function. She has already tested this suggestion and found that it works as expected:

```ts
function wrapHandler(handler: EventHandler): (msg: AgentEventMessage) => void {
  return (m: AgentEventMessage) => {
    if (m.type !== "channelChangedEvent") {
      // Ignore unsupported event types (or throw if appropriate)
      return;
    }
    // 'currentChannelId' takes precedence over deprecated 'newChannelId' if it's set.
    const currentChannelId =
      m.payload.currentChannelId ?? m.payload.newChannelId ?? null;
    // Creating correct channel changed even which follows the typings of the standard
    const channelChangedEvent: FDC3ChannelChangedEvent = {
      type: "userChannelChanged",
      details: { currentChannelId }
    };
    handler(channelChangedEvent);
  };
}
```

### Area of Issue

- [ ] App Directory
- [x] API
- [ ] Context Data
- [ ] Intents
- [ ] Desktop Agent Bridging
- [ ] Use Cases
- [ ] Other

ksgeorgieva:
## Describe your change

The `addEventListener` callback was invoked with the wrong event object when a user left a channel. Also, the `currentChannelId` payload field was not used properly, since `newChannelId` is deprecated.

### Related Issue

Resolves #1869 

## Review Checklist

### DIFF
--- a/packages/fdc3-agent-proxy/src/listeners/DesktopAgentEventListener.ts
+++ b/packages/fdc3-agent-proxy/src/listeners/DesktopAgentEventListener.ts
@@ -8,22 +8,30 @@ import {
 import { Messaging } from '../Messaging.js';
 import { AbstractListener } from './AbstractListener.js';
 
+const handleChannelChangedEvent = (handler: EventHandler, m: ChannelChangedEvent) => {
+  const currentChannelId = m.payload.currentChannelId ?? m.payload.newChannelId ?? null;
+
+  const channelChangedEvent: FDC3ChannelChangedEvent = {
+    type: 'userChannelChanged',
+    details: {
+      currentChannelId,
+    },
+  };
+
+  handler(channelChangedEvent);
+};
+
 function wrapHandler(handler: EventHandler): (msg: AgentEventMessage) => void {
   return (m: AgentEventMessage) => {
-    if (m.type !== 'channelChangedEvent') {
-      return;
+    if (m.type === 'channelChangedEvent') {
+      return handleChannelChangedEvent(handler, m);
     }
 
-    const currentChannelId = m.payload.currentChannelId ?? m.payload.newChannelId ?? null;
-
-    const channelChangedEvent: FDC3ChannelChangedEvent = {
-      type: 'userChannelChanged',
-      details: {
-        currentChannelId,
-      },
-    };
-
-    handler(channelChangedEvent);
+    //forward other events
+    handler({
+      type: m.type,
+      details: m.payload,
+    });
   };
 }
 
--- a/packages/fdc3-agent-proxy/src/listeners/DesktopAgentEventListener.ts
+++ b/packages/fdc3-agent-proxy/src/listeners/DesktopAgentEventListener.ts
@@ -10,23 +10,20 @@ import { AbstractListener } from './AbstractListener.js';
 
 function wrapHandler(handler: EventHandler): (msg: AgentEventMessage) => void {
   return (m: AgentEventMessage) => {
-    //backwards compatibility, see issue #1585
-    if (m.type === 'channelChangedEvent' && m.payload.newChannelId) {
-      const restructured: FDC3ChannelChangedEvent = {
-        type: 'userChannelChanged',
-        details: {
-          currentChannelId: m.payload.newChannelId,
-        },
-      };
-
-      handler(restructured);
-    } else {
-      // currently unused but meeting issue #1585
-      handler({
-        type: m.type,
-        details: m.payload,
-      });
+    if (m.type !== 'channelChangedEvent') {
+      return;
     }
+
+    const currentChannelId = m.payload.currentChannelId ?? m.payload.newChannelId ?? null;
+
+    const channelChangedEvent: FDC3ChannelChangedEvent = {
+      type: 'userChannelChanged',
+      details: {
+        currentChannelId,
+      },
+    };
+
+    handler(channelChangedEvent);
   };
 }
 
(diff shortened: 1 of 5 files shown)


packages/fdc3-agent-proxy/src/listeners/DesktopAgentEventListener.ts
```diff
@@ -10,23 +10,20 @@ import { AbstractListener } from './AbstractListener.js';
...
+
+    const channelChangedEvent: FDC3ChannelChangedEvent = {
+      type: 'userChannelChanged',
+      details: {
+        currentChannelId,
+      },
+    };
+
+    handler(channelChangedEvent);
   };
```
kriswest: I think the if statement at the beginning may block the contextCleared event from being returned in FDC3 3.0. I think it needs changing back so that we retain the else clause that was forwarding other events as received, but retaining your fix to the currentChannelId / newChannelId handling. Doesn't matter in 2.2 (as theres only `channelChangedEvent` event), but it'll be easier to merge into main as well as the release as below:

```suggestion
    if (m.type == 'channelChangedEvent') {
      //backwards compatibility, see issue #1585
      const currentChannelId = m.payload.currentChannelId ?? m.payload.newChannelId ?? null;

      const channelChangedEvent: FDC3ChannelChangedEvent = {
        type: 'userChannelChanged',
        details: {
          currentChannelId,
        },
      };
  
      handler(channelChangedEvent);
    } else {
      //forward other events
      handler({
        type: m.type,
        details: m.payload,
      });
    }
};
```

kriswest (changes requested):
Thanks for contributing this @ksgeorgieva. Suggesting a tweak that will make this easier to merge into both main and 2.2.* release line. Could you check and merge the suggestion if acceptable then I think this is good to go and I can get on with release prep.

packages/fdc3-agent-proxy/src/listeners/DesktopAgentEventListener.ts
ksgeorgieva: I've refactored it a bit further by extracting the `channelChangedEvent` handling into a dedicated helper.
I think this makes `wrapHandler` easier to read and maintain, while also keeping the generic event forwarding path very clear. It will also scale better for FDC3 3.0 — if additional events require custom handling in the future, we can follow the same pattern and keep each event-specific transformation isolated in its own handler function.

packages/fdc3-agent-proxy/test/features/user-channels.feature
```diff
@@ -267,6 +267,52 @@ Feature: Basic User Channels Support
...
+    Given "typesHandler" pipes events to "types"
+    And "bothFieldsMessage" is a channelChangedEvent message with currentChannelId "modern" and newChannelId "deprecated"
+    When I call "{api}" with "addEventListener" with parameters "userChannelChanged" and "{typesHandler}"
+    And I refer to "{result}" as "theListener"
+    And messaging receives "{bothFieldsMessage}"
+    Then "{types}" is an array of objects with the following contents
+      | currentChannelId |
+      | modern           |
+
+  Scenario: Wildcard event listener does not fire for non-channelChangedEvent messages
```
kriswest: This test needs to be removed or adjusted after the refactor as other events can now fire again.

kriswest: You can find the failure at: https://github.com/finos/FDC3/actions/runs/25853635398/job/75965958257?pr=1870#step:6:4994

kriswest (changes requested):
Looks great @ksgeorgieva. However, one of the tests needs to be rethought or removed as its failing after we restored forwarding of other events.

packages/fdc3-agent-proxy/test/features/user-channels.feature
ksgeorgieva: Updated

kriswest (approved):
Great job, particularly on the coverage. Merging and will start prepping a 2.2.3 release with this

PR #1887, opened 2026-05-18
test: add getInfo optional features conformance check

### LINKED ISSUE #1776: Create basic conformance test to ensure that Optional Features are returned as part of ImplementationMetadata
Basic conformance test should ensure that the DesktopAgent returns a valid optionalFeatures object. This can be done as part of the `getInfo` test

nanookclaw:
## Summary

Adds a basic conformance test for `fdc3.getInfo()` to require `ImplementationMetadata.optionalFeatures`, covering the two required boolean members: `UserChannelMembershipAPIs` and `DesktopAgentBridging`.

The existing advanced metadata validator already checked those members, so this extracts that portion into `validateOptionalFeatures()` and reuses it from both the full metadata validation path and the new `BasicGI2` test.

Closes #1776.

## Verification

- `git diff --check`
- `npx tsc -b packages/fdc3-context packages/fdc3-schema packages/fdc3-standard packages/fdc3-agent-proxy packages/fdc3-get-agent packages/fdc3`
- `npm run build --workspace fdc3-conformance`


toolbox/fdc3-conformance/src/test/basic/fdc3.basic.ts
```diff
@@ -81,6 +84,17 @@ const basicGI1 = (fdc3: DesktopAgent, documentation: string) => {
   });
 };
 
+const basicGI2 = (fdc3: DesktopAgent, documentation: string) => {
+  it('(BasicGI2) ImplementationMetadata contains optionalFeatures with required boolean members', async () => {
```
Roaders, line 88: It doesn't look like this is actually validating the boolean members. It looks like it's just checking that the `optionalFeatures` property exists

nanookclaw: Good catch to call out. The boolean-member assertions are in the shared metadata validator this test calls: `toolbox/fdc3-conformance/src/test/support/metadata-support.ts` now asserts that `optionalFeatures` exists, that both spec-required members exist, and that each member's `typeof` is `boolean`.

`BasicGI2` calls that helper, so it should fail if either required member is missing or non-boolean. The current API interface defines those two required keys in `packages/fdc3-standard/src/api/ImplementationMetadata.ts` (`UserChannelMembershipAPIs` and `DesktopAgentBridging`).

If you'd prefer the assertions inline in `BasicGI2` for readability, I can move them there; I kept them in `MetadataValidator` because the existing metadata validation path uses the same rules.

kriswest:
closed in favour of #1888 many thanks for your contribution!

PR #1919, opened 2026-06-16
Security documentation kw code edits

kriswest:
## Describe your change

Changes that I worked out while testing the web reference demo. Not all the demos are consistently styled, but the older ones are much closer after this and less, well, awful :-p 

- Removing the messaging property from IntentResolution as it shouldn't be there and isn't used (noticed it while testing demo apps)
- Adding styling to the old front-end demo apps and fixing content rendering - makes them more useful
- Removing the 'old' app - I just don't think we need it and it will never work in the web demo
- Changing categories on the demo apps to better organise them in the launcher view
  - However, we are abusing the categories field to do this - we're not using the Standard's tags


robmoffat (approved):
The trading view apps look a bit weird now but everything else is good

Image

Yannick-Malins (approved):
reviewed and tested locally with Rob. All looks good

PR #1926, opened 2026-06-17
CodeGen of AppD record, adding minimumAgentVersion 

### LINKED ISSUE #1382: Question: Documentation or Types for Desktop Agent implementation of App Directory
### Question Area
 - [x] App Directory
 - [ ] API
 - [ ] Context Data
 - [ ] Intents
 - [ ] Use Cases
 - [ ] Other

### Question

We are close to implementing the app directory support for our Desktop Agent. I have been reading the docs but there does not seem to be any clear indication about how a Desktop Agent should handle app directory records - I am thinking specifically about how to open an app.

The only docs that I have found is the [swagger](https://fdc3.finos.org/schemas/2.1/app-directory.html#tag/Application/paths/~1v2~1apps~1%7BappId%7D/get). This is of some use but I was hoping for at least some Typescript `d.ts` files.

Is it specified anywhere how a DA should open an app?  I have made some assumptions:

- the DA should look through the hostManifests until it finds one it recognises and then use that to open the app.
- the host manifests should be listed in order of priority (if the DA knows about the first one it will open that and ignore the rest)
- ideally a web manifest should always be included at the end of the list so that a DA can always open an app even if know nothing about the other manifests
- the `url` supplied in the `details` section is not used for opening the app but is just used for app identity

Are these assumptions correct?
Thanks

### LINKED ISSUE #1871: App Directory does not contain information about the version of FDC3 that the application requires
At the moment the Finos app directory listed at https://directory.fdc3.finos.org/v2/apps/ only lists (as far as I can tell) FDC3 2.0 apps that will not work with a FDC3 2.2 web desktop agent as they do not use `getAgent()`.

It would probably be a good idea if the app directory specified which version of FDC3 an app was written against. Possibly we need a supported feature list as well.

For example. If I show a list of applications to open or that can handle an intent I want to filter out all the applications that do not use `getAgent()` as they will not work with my desktop agent.

robmoffat:
## Describe your change

resolves #1871
resolves #1382 

Migrates the FDC3 AppD schema into the fdc3-schema module and adds generation of typescript types for working with the directory API and individual appD records, to aid Desktop Agent developers.

Also adds a new `minimumAgentVersion` field to the app directory record to indicate the minimum version of FDC3 necessary to run the application.

## Review Checklist

- [x] **API changes**: AppD record
  - [x] **Docs & Sources**: If yes, were both documentation (/docs) and sources updated?<br/>


packages/fdc3-schema/generated/app-directory/AppDirectoryTypes.ts
```diff
@@ -0,0 +1,877 @@
...
+   */
+  title: string;
+  type: Type;
+  details: LaunchDetails;
+  /**
+   * An array of string categories that describe the application. These are meant as a hint to
+   * catalogs or stores listing FDC3-enabled apps and it is expected that these will make a
+   * best effort to find appropriate categories (or category) under which to list the app.
+   * AppD record authors are encouraged to use lower-case and, where possible, to select
+   * categories from the following list:
```
Roaders, line 48: if only certain types are allowed should this be an enum or a string union to only allow these?

packages/fdc3-schema/generated/app-directory/AppDirectoryTypes.ts
```diff
@@ -0,0 +1,877 @@
...
+   */
+  contactEmail?: string;
+  /**
+   * An optional set of name value pairs that can be used to deliver custom data from an App
+   * Directory to a launcher. Deprecated due to a lack of a standard means of retrieval via
+   * the Desktop Agent API. To be replaced in a future version with an `applicationConfig`
+   * element and standard API to retrieve it. See issue
+   * [#1006](https://github.com/finos/FDC3/issues/1006) for details.
+   */
+  customConfig?: Array<any[] | boolean | number | number | null | NameValuePairObject | string>;
```
Roaders, line 96: the comment says name value pairs but this is an array that accepts a lot more than just name value pairs.

packages/fdc3-schema/generated/app-directory/AppDirectoryTypes.ts
```diff
@@ -0,0 +1,877 @@
...
+   */
+  broadcasts?: string[];
+  /**
+   * Context type names that the application listens for.
+   */
+  listensFor?: string[];
+  [property: string]: any;
+}
+
+export interface BaseApplicationObject {
```
Roaders, line 362: There is a lot of duplication in here. Should `Application` extend this?

website/docs/app-directory/spec.md
```diff
@@ -49,3 +49,27 @@ An FDC3 Standard compliant App Directory implementation **MAY**:
...
+This field is **distinct from** the existing `version` field, which describes the application's own software release (for example `"1.0.0"`).
+
+When `minimumAgentVersion` is **`2.2` or higher**, the application is assumed to require the Web Connection Protocol and **`getAgent()`** (see [Supported Platforms](../api/supported-platforms.md)).
+
+When `minimumAgentVersion` is **absent**, Desktop Agents MUST NOT exclude the application solely on compatibility grounds — the record is treated as legacy or unknown.
+
+Desktop Agents **SHOULD** exclude applications from open and intent resolution candidate lists when:
+
+- `minimumAgentVersion` is present and the Desktop Agent's `fdc3Version` (from `getInfo()`) is lower than the required version, or
+- `minimumAgentVersion` is `2.2` or higher and the Desktop Agent does not support the Web Connection Protocol / `getAgent()`.
```
Roaders, line 66: Is this true? If I have an existing 2.0 DA and application and I want to update to FDC3 2.2 I have to re-write all of the comms to use the WCP? I can't carry on using my existing communication method?

I guess in this case the Desktop Agent vendor would ignore the `SHOULD` here and include the app anyway.

kriswest: It's not true as these records also apply to non-web applications where WCP and getAgent don't apply - the AppD standard also shouldn't be trying to govern something from the API Part.

PR #1930, opened 2026-06-18
Add setInstanceMetadata API to allow apps to update instance-specific metadata (#1846)

### LINKED ISSUE #1846: No mechanism exists for identifying different instances of the same app.
### Use Case

FDC3 allows a user to open multiple instances of the same application. If you have an application that displays a stock price chart then you could have an instance that displays prices for Apple, Google and Meta. If you raise an intent and want to handle that intent in one of those applications there is currently no way of telling these 3 instances apart and they will all appear the same in the app resolver UI.

The closest that we currently have is the `name` and `title` field in the `AppMetadata` object. This object can be returned by the `getAppMetadata` function where it should obviously return data for the queried app but when returned from the `findIntents` or `findIntentsByContext` functions it would be possible to modify this object on a per instance basis. It would also be possible to modify the result from `getAppMetadata` to return instance specific details if an instanceId was included in the AppIdentifier - but I do no think that this would be a good idea.

There is also no defined way that an application is able to update it's title so that if a user navigates from Apple to Google within the app the Desktop Agent is able to correctly display the updated title the next time that the app resolver ui is displayed.

Questions:

- should we add a `getInstanceMetadata` function that would return an `InstanceMetadata` object? (this object may not be required, we could probably just use the AppMetadata object).
- should we define a way that applications are able to inform the desktop agent what their title is? It would currently be possible to generate a title for new instance when it is opened based on the context that is sent. For example the name of a contact or the id of a stock ticker but there is no standardized way of updating this title. It is obviously possible to do this but with no standardised way of doing it apps from different sources will not be able to do it in Desktop Agents that they do not know about.

I thought it was worth discussing this before I implement a home brew way of handling this in our own agent.

Roaders:
## Describe your change

Add InstanceMetadata type and setInstanceMetadata() method to DesktopAgent, allowing apps to update their instance-specific metadata (e.g. title) so Desktop Agents can display meaningful information in resolver UIs to disambiguate multiple instances of the same app.

- Add InstanceMetadata interface with title field and index signature
- Update AppMetadata.instanceMetadata to use InstanceMetadata type
- Add setInstanceMetadataRequest/Response DACP message schemas
- Implement client-side in fdc3-agent-proxy (DefaultAppSupport)
- Implement server-side in fdc3-web-impl (OpenHandler)
- Store instanceMetadata on AppRegistration and return in getAppMetadata
- Add Cucumber tests in both proxy and web-impl
- Regenerate BrowserTypes.ts and BridgingTypes.ts

### Related Issue
fixes #1846

## Review Checklist

---


packages/fdc3-standard/src/api/InstanceMetadata.ts
```diff
@@ -0,0 +1,26 @@
...
+ * allowing Desktop Agents to display meaningful, instance-specific information
+ * (e.g. the name of the contact or instrument currently being viewed).
+ *
+ * Must only be set on an `AppMetadata` object if `instanceId` is also set.
+ */
+export interface InstanceMetadata {
+  /** A user-friendly title for this specific instance that can be used to render UI elements,
+   *  such as a resolver or intent picker, to help distinguish it from other instances of the same application.
+   *  For example, an app displaying a stock chart might set this to the name of the instrument being displayed. */
+  readonly title?: string;
```
robmoffat, line 20: The only problem with this being here is that AppMetadata already holds the title https://fdc3.finos.org/docs/api/ref/Metadata

packages/fdc3-standard/src/api/InstanceMetadata.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  /** Additional custom metadata fields that can be used to further disambiguate instances.
+   *  These are implementation-specific and may include information such as screen position,
+   *  workspace assignment, or other instance-specific details. */
+  readonly [key: string]: any;
```
robmoffat, line 25: Also, are there no other properties we can enumerate?  This is a whole lot of stuff for just setting a title.  

It seems like the main place you want this is to disambiguate windows in the intent resolver (I think).  So are there not further things we could set to help with that?

packages/fdc3-agent-proxy/src/apps/PageTitleSupport.ts
```diff
@@ -0,0 +1,116 @@
+/**
```
robmoffat, line 1: I feel like all this is overkill. We should simplify this by just leaving it up to the app to call the method on the DesktopAgent when it wants to, rather than trying to bind these things together programmatically.  Just my 10c

kriswest:
@Roaders I will circle around to your reviews in the next day or two, trying to strike while the iron is hot on a few new contributors with MRs for the 3.0 scope as there are some solid contributions there! + a little cleardown on the dependency updates needed (if you get a chance to approve my two PRs for those please do)

packages/fdc3-standard/src/api/InstanceMetadata.ts
kriswest: We did discuss this a little already, it does seem reasonable for an instance to be able to pick up a title that differentiates it from other instances. However, it could be more intuitive by using a name like `instanceTitle`?

Kudos for already adding usage to the intent resolvers. However, it might be worth us demonstrating the use of both titles if available? Rather than substituting the instance title? If we don't recommend using both we should be explicit about _recommending_ that the app title is embedded in instance title (SHOULD rather than MUST).

packages/fdc3-standard/src/api/InstanceMetadata.ts
kriswest: we talked a little about a placement description? I think we'd need to stay well away from standardizing that for now, but allow for an additional descriptive string to help with disambiguation. However... that seems like a job of the DA now that I think about it, so already possible in a custom intent resolver (but not an injected one as we'd need another field to be transmitted).

PR #1965, opened 2026-07-08
Add optional FDC3 version property to AppD record

### LINKED ISSUE #1871: App Directory does not contain information about the version of FDC3 that the application requires

julianna-ciq:
## Summary
- add AppD `fdc3Version` metadata as an npm-style semver range
- filter web Desktop Agent AppD records by compatibility with the agent FDC3 version
- add focused coverage for compatible, incompatible, invalid, and unversioned AppD records

Fixes #1871

## Verification
- `npm test --workspace toolbox/fdc3-for-web/fdc3-web-impl` (124 tests passed)
- `tsc -p toolbox/fdc3-for-web/demo/tsconfig.json --noEmit`
- Prettier check on changed files


kriswest:
picked up some conflicts and is failing out of the website build for some reason - needs investigation

julianna-ciq:
@kriswest I think Netlify is failing because `.npmrc` has a min-release-age = 7 [rule](https://github.com/finos/FDC3/blob/main/.npmrc#L14), but dependency `fast-uri@3.1.4` is [only 3 days old](https://www.npmjs.com/package/fast-uri?activeTab=versions). I think we'll need to downgrade fast-uri to 3.1.2 if we want to merge anything in the next 4 days

kriswest:
Hmm, dependabot is supposed to have the same setting: 
https://github.com/finos/FDC3/blob/31f79d4faf566eb77657fbbd1d72f0d7b7fc56a2/.github/dependabot.yml#L13

and it proposed that update. I guess its not applying that config to the website folder?? Would also explain why its not grouping updates for the website.

I'll raise something to try and do that downgrade and fix the dependabot config.

kriswest:
@julianna-ciq This should fix it, could you take a look: https://github.com/finos/FDC3/pull/2007
If happy go ahead and merge it, then rebase here

packages/fdc3-standard/src/app-directory/specification/appd.schema.json
```diff
@@ -471,6 +471,11 @@
 			  "type": "string",
 			  "description": "Version of the application. This allows multiple app versions to be defined using the same app name. This can be a triplet but can also include things like 1.2.5 (BETA)"
 			},
+			"fdc3Version": {
+			  "type": "string",
+			  "format": "semver-range",
```
kriswest: Do you know what version of the json schema specification this is supported in? Can we get that added to the description.

If its not been adopted we can do this with a regex over a string type. Theres a regex for semver in here somewhere: https://github.com/semver/semver.org/issues/431

kriswest: I don't see it in 2020-12: https://www.learnjsonschema.com/2020-12/format-annotation/format/

maybe a hallucination?

kriswest:
Is this still a draft? can you change title if not and I'll review

packages/fdc3-standard/src/app-directory/specification/appd.schema.json
```diff
@@ -953,6 +958,11 @@
 			  "type": "string",
 			  "description": "Version of the application. This allows multiple app versions to be defined using the same app name. This can be a triplet but can also include things like 1.2.5 (BETA)"
 			},
+			"fdc3Version": {
```
kriswest: This shouldn't be getting added to the v1 app spec. However, we could also remove that entirely as its overdue to go along with the deprecated routes.

toolbox/fdc3-for-web/fdc3-web-impl/src/directory/BasicDirectory.ts
```diff
@@ -14,14 +15,35 @@ export function genericResultTypeSame(real: string | undefined, required: string
...
 }
 
+export function appSupportsFdc3Version(app: DirectoryApp, fdc3Version: string): boolean {
+  const appFdc3Version = app.fdc3Version;
+  if (appFdc3Version == null) {
+    return true;
+  }
+
+  const range = semver.validRange(appFdc3Version);
+  return range != null && semver.satisfies(fdc3Version, range);
```
kriswest: This isn't using the regex defined in the App Directory schema and probably should be...

kriswest: Can we validate using teh schema OR at least the regex from the schema

kriswest (dismissed):
The regex for validation needs testing and needs a few tweaks - we should lose the v1 appD spec at the same time

toolbox/fdc3-for-web/fdc3-web-impl/src/directory/BasicDirectory.ts
```diff
@@ -1,5 +1,10 @@
+import semver from 'semver';
 import { Directory, DirectoryApp, DirectoryIntent } from './DirectoryInterface.js';
 
+// Keep synchronized with BaseApplication.properties.fdc3Version.pattern in appd.schema.json.
+export const FDC3_VERSION_RANGE_PATTERN =
```
Roaders, line 5: is this the best way of doing this? Can't we use the semver library to verify the versions?
If we do want to keep this it's probably a good idea to have some comments here explaining with examples of what are valid versions.

toolbox/fdc3-for-web/fdc3-web-impl/src/directory/BasicDirectory.ts
```diff
@@ -14,14 +19,40 @@ export function genericResultTypeSame(real: string | undefined, required: string
   }
 }
 
+export function appSupportsFdc3Version(app: DirectoryApp, fdc3Version: string): boolean {
```
Roaders, line 22: I assume `fdc3Version` here is the version of the desktop agent and not a range?

kriswest:
@julianna-ciq just a reminder to complete the changes on this one so we can merge

toolbox/fdc3-for-web/fdc3-web-impl/src/directory/BasicDirectory.ts
julianna-ciq: Updated the comment to explain why both checks are needed: semver validates and evaluates the npm-style range, while the schema regex enforces FDC3's two-component version form and rejects semver-valid triplets such as 1.2.3. The comment now includes valid and invalid examples.

toolbox/fdc3-for-web/fdc3-web-impl/src/directory/BasicDirectory.ts
julianna-ciq: Yes, this is the exact Desktop Agent version, while app.fdc3Version is the application's supported range. I renamed the function argument, constructor argument, and stored property to desktopAgentFdc3Version to make that distinction explicit.

julianna-ciq:
@kriswest The outstanding review items are now addressed: latest main is merged, the regex comment explains why the schema restriction is needed in addition to semver (including valid/invalid examples), and the Desktop Agent version parameter/property has been renamed to desktopAgentFdc3Version for clarity. Roaders' two threads are resolved, and all fresh CI/CD checks are passing. Could you please re-review?

Roaders (dismissed):
lgtm

toolbox/fdc3-for-web/demo/src/client/da/FDC3_2_1_JSONDirectory.ts
```diff
@@ -18,19 +18,12 @@ const convertToDirectoryList = (data: any): DirectoryApp[] => {
 };
 
 export class FDC3_2_1_JSONDirectory extends BasicDirectory {
```
kriswest: Update filename and classname to FDC3_JSONDirectory so that we don't have to keep bumping this - as aa reference implementation it should be maintained against teh current standard

package-lock.json
```diff
@@ -15800,14 +15806,16 @@
       "version": "3.0.0-alpha.2",
       "license": "Apache-2.0",
       "dependencies": {
-        "@finos/fdc3-standard": "3.0.0-alpha.2"
+        "@finos/fdc3-standard": "3.0.0-alpha.2",
+        "semver": "^7.5.3"
```
kriswest: needs bumping to 7.8.* to match the types imported and to avoid an immediate dependabot PR

kriswest (commented):
Update semver version to 7.8.*, otherwise all LGTM

toolbox/fdc3-for-web/demo/src/client/da/FDC3_2_1_JSONDirectory.ts
julianna-ciq: Renamed the demo file and class to `FDC3_JSONDirectory` and updated its import and usage in the dummy Desktop Agent.

package-lock.json
julianna-ciq: Updated `semver` to the latest release, 7.8.5. `@types/semver` was already at the latest release, 7.8.0, and the lockfile now resolves both versions.

Roaders (approved):

PR #1977, opened 2026-07-09
fix(get-agent): reduce normal discovery log noise

### LINKED ISSUE #1902: `getAgent` discovery emits warnings/errors for normal scenarios, creating misleading console noise
### Use Case

`getAgent` produces console warnings and errors that aren't actually problems. In applications with existing non-FDC3 postMessage traffic, or applications designed to run both with and without a Desktop Agent, messages not related to the connection workflow produce misleading warnings — a consumer reading the console might think something is broken when it isn't.

Adjusting the `AbstractLogger`'s `logLevel` to prevent these from being written might be too heavy handed and result in genuinely useful logs being hidden as well.

#### Examples

**1. `HelloHandler` warns on every unrelated postMessage**

During connection/discovery, `HelloHandler` logs a warning for any received message whose `connectionAttemptUuid` doesn't match the expected one:

```
FDC3 getAgent:  HelloHandler: Ignoring message with invalid connectionAttemptUuid.
Expected 9e489a1f-..., received: undefined { id: "fdsMetadata", type: "READY_CONFIRMATION", name: "containerChain" }
```

In applications already built around postMessage for non-FDC3 purposes (host harnesses, framework messaging, custom RPC, etc.), every one of those unrelated messages produces an FDC3 warning. The warnings describe normal traffic the handler is correctly choosing to ignore — not a misconfiguration or a protocol violation. It's debug-level information at best.

**2. "Desktop agent not found" is logged as an error**

```
FDC3 getAgent:  Desktop agent not found. No error reported during discovery.
```

For applications designed to run both inside and outside an FDC3 container — adapting their messaging when an agent is present, working normally otherwise — the absence of an agent is an expected runtime state, not an error. Reporting it at error level gives consumers the impression that something failed, and risks the symptom being mis-attributed to FDC3 when investigating unrelated issues.

#### Suggested change

- Demote `HelloHandler`'s "Ignoring message with invalid connectionAttemptUuid" to `debug`, or drop it. It only describes traffic the handler is intentionally filtering.
- Demote "Desktop agent not found. No error reported during discovery." to `info` or `debug`. Reserve `error` for actual failures during discovery (a discovery strategy threw, a handshake started and failed, etc.).

The general principle: only emit `warn`/`error` for conditions a consumer should investigate. Routine filtering of unrelated messages and the legitimate "no agent in this environment" outcome are neither.

LarryHu0217:
## Describe your change

Reduces misleading console noise during normal `getAgent` discovery. Unrelated `postMessage` traffic is ignored without warning, while valid WCP2/WCP3 messages with the wrong connection-attempt UUID still warn. An expected agent-not-found timeout is reported as a warning instead of an error. Feature tests capture console output and verify all three behaviors.

Resolves #1902.

## Validation

- Prettier check on the changed TypeScript test file
- `npm --workspace packages/fdc3-get-agent run lint`
- `npm --workspace packages/fdc3-get-agent run test` (38 passing)
- `git diff --check origin/main...HEAD`

### DIFF
--- a/packages/fdc3-get-agent/test/features/desktop-agent-strategy.feature
+++ b/packages/fdc3-get-agent/test/features/desktop-agent-strategy.feature
@@ -402,6 +402,12 @@ Scenario: Latch to Desktop Agent Preload via SessionStorage which has gone away
     When a HelloHandler for connection attempt "expected-connection-uuid" receives unrelated postMessage traffic
     Then captured console warn output should not contain "invalid connectionAttemptUuid"
 
+  Scenario: WCP messages for another connection attempt still produce warnings
+    Given console output is captured
+    When a HelloHandler for connection attempt "expected-connection-uuid" receives WCP messages for other connection attempts
+    Then captured console warn output should contain "received: wcp2-other-connection-uuid"
+    And captured console warn output should contain "received: wcp3-other-connection-uuid"
+
   Scenario: Someone calls getAgent twice
     Given Parent Window desktop "da" listens for postMessage events in "{parentWin}", returns direct message response
     And we wait for a period of "200" ms
--- a/packages/fdc3-get-agent/test/features/desktop-agent-strategy.feature
+++ b/packages/fdc3-get-agent/test/features/desktop-agent-strategy.feature
@@ -390,7 +390,7 @@ Scenario: Latch to Desktop Agent Preload via SessionStorage which has gone away
     Given console output is captured
     When I call getAgent for a promise result with the following options
       | dontSetWindowFdc3 | timeoutMs | intentResolver | channelSelector |
-      | true              |       100 | false          | false           |
+      | true              |      4000 | false          | false           |
     And I refer to "{result}" as "theAPIPromise"
     Then the promise "{theAPIPromise}" should resolve
     And "{result}" is an error with message "AgentNotFound"
@@ -400,7 +400,7 @@ Scenario: Latch to Desktop Agent Preload via SessionStorage which has gone away
   Scenario: Unrelated postMessage traffic is ignored without warning
     Given console output is captured
     When a HelloHandler for connection attempt "expected-connection-uuid" receives unrelated postMessage traffic
-    Then captured console warn output should not contain "invalid connectionAttemptUuid"
+    Then captured console debug output should contain "Ignoring unexpected message in HelloHandler"
 
   Scenario: WCP messages for another connection attempt still produce warnings
     Given console output is captured
--- a/packages/fdc3-get-agent/test/features/desktop-agent-strategy.feature
+++ b/packages/fdc3-get-agent/test/features/desktop-agent-strategy.feature
@@ -387,12 +387,20 @@ Scenario: Latch to Desktop Agent Preload via SessionStorage which has gone away
       |         2.0 | cucumber-app      | preload-provider |
 
   Scenario: Nothing works and we timeout
+    Given console output is captured
     When I call getAgent for a promise result with the following options
       | dontSetWindowFdc3 | timeoutMs | intentResolver | channelSelector |
-      | true              |      4000 | false          | false           |
+      | true              |       100 | false          | false           |
     And I refer to "{result}" as "theAPIPromise"
     Then the promise "{theAPIPromise}" should resolve
     And "{result}" is an error with message "AgentNotFound"
+    And captured console warn output should contain "Desktop agent not found. No error reported during discovery."
+    And captured console error output should not contain "Desktop agent not found. No error reported during discovery."
+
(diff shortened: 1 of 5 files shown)


packages/fdc3-get-agent/test/features/desktop-agent-strategy.feature
```diff
@@ -387,12 +387,26 @@ Scenario: Latch to Desktop Agent Preload via SessionStorage which has gone away
       |         2.0 | cucumber-app      | preload-provider |
 
   Scenario: Nothing works and we timeout
+    Given console output is captured
     When I call getAgent for a promise result with the following options
       | dontSetWindowFdc3 | timeoutMs | intentResolver | channelSelector |
-      | true              |      4000 | false          | false           |
+      | true              |       100 | false          | false           |
```
kriswest: not sure why this timeout was shortened (or why it was originally 4 seconds TBH - however execution of tests can sometimes be slow on CI so I'm wary of shortening it without an explanation)

packages/fdc3-get-agent/test/features/desktop-agent-strategy.feature
```diff
@@ -387,12 +387,26 @@ Scenario: Latch to Desktop Agent Preload via SessionStorage which has gone away
...
     And I refer to "{result}" as "theAPIPromise"
     Then the promise "{theAPIPromise}" should resolve
     And "{result}" is an error with message "AgentNotFound"
+    And captured console warn output should contain "Desktop agent not found. No error reported during discovery."
+    And captured console error output should not contain "Desktop agent not found. No error reported during discovery."
+
+  Scenario: Unrelated postMessage traffic is ignored without warning
+    Given console output is captured
+    When a HelloHandler for connection attempt "expected-connection-uuid" receives unrelated postMessage traffic
+    Then captured console warn output should not contain "invalid connectionAttemptUuid"
```
kriswest: I don't think we need to be testing that the warning that was removed is NOT displayed. However, you could check for the debug output that replaced it. Either remove or replace this check

kriswest (commented):
This mostly LGTM (thanks for adding the console capture and testing) - a couple of small tweaks requested.

packages/fdc3-get-agent/test/features/desktop-agent-strategy.feature
```diff
@@ -387,12 +387,24 @@ Scenario: Latch to Desktop Agent Preload via SessionStorage which has gone away
...
     And I refer to "{result}" as "theAPIPromise"
     Then the promise "{theAPIPromise}" should resolve
     And "{result}" is an error with message "AgentNotFound"
+    And captured console warn output should contain "Desktop agent not found. No error reported during discovery."
+    And captured console error output should not contain "Desktop agent not found. No error reported during discovery."
+
+  Scenario: Unrelated postMessage traffic is ignored without warning
+    When a HelloHandler for connection attempt "expected-connection-uuid" receives unrelated postMessage traffic
+
+  Scenario: WCP messages for another connection attempt still produce warnings
```
kriswest:
```suggestion
  Scenario: WCP messages for another connection attempt produce warnings
```

packages/fdc3-get-agent/test/features/desktop-agent-strategy.feature
```diff
@@ -387,12 +387,24 @@ Scenario: Latch to Desktop Agent Preload via SessionStorage which has gone away
...
       | dontSetWindowFdc3 | timeoutMs | intentResolver | channelSelector |
       | true              |      4000 | false          | false           |
     And I refer to "{result}" as "theAPIPromise"
     Then the promise "{theAPIPromise}" should resolve
     And "{result}" is an error with message "AgentNotFound"
+    And captured console warn output should contain "Desktop agent not found. No error reported during discovery."
+    And captured console error output should not contain "Desktop agent not found. No error reported during discovery."
+
+  Scenario: Unrelated postMessage traffic is ignored without warning
+    When a HelloHandler for connection attempt "expected-connection-uuid" receives unrelated postMessage traffic
```
kriswest: Needs a then clause for there to be any point to this test (its doing something but not checking it). I know I asked for the check on the old warning messages to be removed, but i think it may need to check that the output produced is now at the debug level? Its that or drop this test entirely.

kriswest (commented):
Slight adjustment to oone of the scenario names + one of the tests is not currently a test and either needs a then clause or removal.

LarryHu0217:
Addressed the follow-up in `abe0ff30`: the unrelated-message scenario now enables DEBUG logging and asserts the debug output, and the WCP warning scenario uses the requested name. The full get-agent suite passes (38 tests), and GitHub CI is green.

kriswest (approved):
LGTM!

Issue #1852, opened 2026-04-15
Automate maintenance of ContextTypes and IntentTypes enums and type in fdc3-standard


kriswest:
## Enhancement Request

fdc3-standard contains: 
- a `ContextType` type and deprecated `ContextTypes` enum (composed from A `StandardContextTypes` and `ExperimentalContextTypes` enums) at https://github.com/finos/FDC3/blob/cdc754e507e81c0eb45cd47c59b11a949d77e8e3/packages/fdc3-standard/src/context/ContextType.ts
- A `StandardIntent` type and deprecated `Intents` enum  at https://github.com/finos/FDC3/blob/main/packages/fdc3-standard/src/intents/Intents.ts

These are hand maintained and the ContextType type was recently found to be out of sync with the contexts in the fdc3-contexts package (now fixed). The types and enums should move to the fdc3-context and be automatically maintained against the schemas (as we have the goal that adding a context is simple as adding the schema to fdc3-context) and moved to fdc3-context to facilitate that (export from there and re-export from fdc3-standard/fdc3 to avoid a breaking change).

A simple script could extract the context types (standard and experimental) from the available schemas and populate the enum(s). The enum should be un-deprecated, as it is useful for developers and can be used to define the type (to avoid duplication):

```ts
export type ContextType= `${ContextTypes}`;
``` 

Some docs for these types and enums should be provided at:
- https://fdc3.finos.org/docs/context/spec#standard-context-types
- https://fdc3.finos.org/docs/intents/spec#standard-intents

LarryHu0217:
I took a quick look at the current repo shape. `ContextTypes` / `StandardContextType` are still hand-maintained in `packages/fdc3-standard/src/context/ContextType.ts`, while `@finos/fdc3-context` already generates context interfaces from the schemas and `fdc3-standard` has `ContextTypeSync.test` as a guard.

If useful, I can start with a narrow first PR that focuses only on the context side: generate/export the context type values from `fdc3-context`, keep the existing `fdc3-standard` re-export/backcompat surface, and leave intents/docs as follow-ups. Would maintainers prefer that split, or should this wait for one PR covering contexts + intents + docs together?

Issue #1713, opened 2025-12-04
Enhance FDC3 standards for LLMs


rmcneely-interopio:
## Enhancement Request

There has been a lot of discussion lately around LLMS and how they intereact with FDC3 and what can be done improve these workflows. This issue is meant to be an broad overview of how we can work to improve the FDC3 standard to make working with LLMs better.  

### Use Case

Generally speaking as LLMs work their way into developer tools there is a natural curiosity to see how they can be used with in context space of working on FDC3 desktop agent applications. There appear to have been multiple independent attempts at working with LLMs and the appD directory including but limited to interop.io's work with ioInsights, Derek Novavi's work with an early MCP-FDC3 library, the discussions around #1699, and various acknowledgements in other discussions.

For example one space that seems to have come up has been using an LLM to pick a published intent as a tool for the desktop agent to proceed with an action. In this scenario a user would like to perform an action (typically while in a chat window) and kick off some interop enabled workflow.

### Proposals

The intent of this issue is to find out if there is enough consensus on these problems to merit raising specific enhancements on how to improve LLM integration. 

To date it seems like a lot of the early attempts at utilizing LLMs with in a desktop agent are tied to trying to mine the appDirectory for appropriate information. Multiple people have mentioned this as a starting point. Our initial goals for working with LLMs should focus on how to make this easier. 

Talking to the team working on integrating LLMs with an FDC3 desktop agent there was one response that I overwhelming received: there is not enough context available in the appDirectory to feed the LLM. Currently the available information inside an apps `intents` property is not rich enough.

The other use case that has been discussed previously has been working with MCP UI or another similar framework on how to produce UI elements in a chat. After speaking with the internal team at interio.io there was concern that this may not be the appropriate place for FDC3 to get involved since very little of the spec is prescriptive of how a desktop agent should handle UI elements.

robmoffat:
@julianna-ciq  - decide if this stays in scope

Issue #719, opened 2022-05-16
Add /intents and /contexts endpoints to the AppD API


kriswest:
## Enhancement Request

When working with an app directory it would be useful to be able to request the list intents and contexts used by apps listed in the app directory. This information is already retrievable, to a certain extent, from the individual records, however, doing so requires interesting all the records and collecting up the information. 

Adding intent and context endpoints to the API would allow: 
- Pre-population of search/browsing interfaces for filtering apps by supported intents and contexts
  - Note the underlying data for such searches may be significantly improved by #247
- The use of these endpoints to provide display data (e.g. displayName)
  - There is a long-standing issue with intents in that each app may configure a different display name for the same intent: #312 

Proposed endpoints:
| route | REST verb | Description |
|---------------------------------------|------|--------------|
| /v2/intents                                      | GET | Retrieve a list intent supported in the app directory instance |
| /v2/intents/{intent name}               | GET | Retrieve detail on a particular intent   |
| /v2/contexts                                   | GET | Retrieve a list of contexts referenced in the app directory instance |
| /v2/contexts/{context type name} | GET | Retrieve detail on a particular context  |

The detail endpoints for an individual intent and context types could:
- Be replaced by returning all the detail under the main endpoints
- Provide a list of the apps that reference it
- Provide a schema or similar detail for the context (however this could/should be optional as not every implementation will want to provide references for standardized contexts)

A number of AppD implementations already provide these endpoints, including electron-fdc3 and the AppD browser demoed to the group by Citi.

It has also been proposed in the past that it is necessary to provide tools to govern internal/proprietary intents and contexts and that this has been implemented in the past through an extension to an app directory implementation. These would be natural extension to the proposed details endpoints.

kriswest:
@nkolba it'd be interesting to hear what you use these endpoints for and the functionality they have in electron-fdc3
@Qiana-Citi @RandallDang-Citi likewise it'd be really interesting to hear more about your use of these endpoints at Citi.

RandallDang-Citi:
> @nkolba it'd be interesting to hear what you use these endpoints for and the functionality they have in electron-fdc3 @Qiana-Citi @RandallDang-Citi likewise it'd be really interesting to hear more about your use of these endpoints at Citi.

@kriswest Thanks for the invitation, Qiana will join.

pierreneu:
@kriswest We are looking into this to be able to dynamically show to the user the locally available intents/apps on a fdc3 object.
[...]

PR #1980, opened 2026-07-13
feat(context): add programmatic context schema discovery

### LINKED ISSUE #1993: Provide programmatic (runtime) access to standardized context type schemas
### Use Case

Consumers of FDC3 increasingly need to reason about the **shape** of standardized context types at runtime, not just their names. Today the JSON Schemas are the single source of truth for context definitions, but they are only used at build time (to generate `ContextTypes.ts`) and are not exposed programmatically. As a result, tooling has to hard-code or re-bundle copies of the schemas.

Concrete end-user problems this would support:

- **AI agents / protocol bridges** (e.g. exposing FDC3 as [WebMCP](https://github.com/webmachinelearning/webmcp)/MCP tools): an agent needs to discover which context types exist and what fields each requires so it can construct valid contexts (e.g. build a valid `fdc3.instrument`) without a human hard-coding every type.
- **Validators**: an app wants to validate a context object it received or is about to broadcast against the standard schema.
- **Dynamic UI / form generation**: a "context inspector" or resolver-style UI wants to render inputs for a chosen context type from its schema (titles, descriptions, examples).
- **Documentation / developer tooling**: enumerate the standardized types and surface their metadata (title, description, examples) without scraping the docs site.

### Workflow Description

Provide a small, read-only, runtime-accessible API (in `@finos/fdc3-context`, re-exported from `@finos/fdc3`) that lets any consumer:

1. Enumerate the standardized context type identifiers.
2. Retrieve the JSON Schema for a given context type.
3. Retrieve lightweight metadata (title, description, `$id`, examples) for a type.

This should require no Desktop Agent connection (the schemas are static, published artifacts) and should stay automatically in sync with the schema files (the source of truth).

### Workflow Examples

```javascript
import { getContextTypes, getContextSchema, getContextSchemaMetadata } from '@finos/fdc3-context';

// 1. Discover the available standardized context types
getContextTypes(); // ["fdc3.action", "fdc3.chart", "fdc3.contact", ...]

// 2. Retrieve a schema and validate/produce a context object (e.g. with Ajv)
const schema = getContextSchema('fdc3.instrument');
const validate = ajv.compile(schema);
validate({ type: 'fdc3.instrument', id: { ticker: 'AAPL' } }); // true

// 3. Drive a UI or agent tool description from schema metadata
getContextSchemaMetadata('fdc3.instrument');
// { type: 'fdc3.instrument', title: 'Instrument', description: '...', id: '...', examples: [...] }
```

### Additional Information

- A draft implementation is proposed in PR #1980 (this issue is being raised per maintainer request to capture the use cases and support milestone planning).
- Per review feedback, the implementation should avoid generating a duplicate copy of every schema and instead load the already-bundled schema files, and should align with the generation work desired in #1852 (auto-maintained `ContextTypes` enum / `ContextType` union derived from the schemas, owned by `fdc3-context` and re-exported from `fdc3-standard`). The `getContextTypes()` capability overlaps with the enum/union from #1852 and can share a single generation step.

nchalaisRocket:
## Describe your change

This PR adds **programmatic context schema discovery** to `@finos/fdc3-context` (and, by re-export, `@finos/fdc3`).

Today the standardized context JSON Schemas are the single source of truth for context definitions, but they are only used at build time (to generate `ContextTypes.ts`) and are not available to consumers at runtime. This makes it hard for tooling — resolvers, validators, form/UI generators, documentation, and agents bridging FDC3 onto other protocols (e.g. MCP) — to reason about the set of available context types and their shapes without hard-coding or re-bundling the schemas.

This change exposes the schemas at runtime via a small accessor API:

| Function | Description |
|----------|-------------|
| `getContextTypes(): string[]` | Sorted list of standardized context type ids that have a published JSON Schema. |
| `getContextSchema(type): ContextSchema \| undefined` | The JSON Schema for a context type, or `undefined`. |
| `getAllContextSchemas(): Record<string, ContextSchema>` | Map of every context type id to its JSON Schema. |
| `getContextSchemaMetadata(type): ContextSchemaMetadata \| undefined` | Lightweight metadata (`title`, `description`, `$id`, `examples`). |
| `isStandardContextType(type): boolean` | Whether an id corresponds to a standardized context type. |

Implementation notes:

- The registry is **generated** from the existing `schemas/context/*.schema.json` files via a new `schemagen` build step (`generateContextSchemas.cjs`), wired into `npm run generate`. It therefore stays automatically in sync with the source-of-truth schemas — no hand-maintained duplication.
- The abstract base context type (`fdc3.context`) is intentionally excluded, as it is not a concrete, broadcastable context type.
- All returned values are defensive copies.
- No changes to the Desktop Agent API, DACP wire protocol, or context type schemas themselves — this is purely additive, read-only, and requires no Desktop Agent connection.
- Added unit tests, a spec section (`website/docs/context/spec.md` → "Programmatic Schema Discovery"), package README docs, and a CHANGELOG entry. Full monorepo `npm run build` and the `@finos/fdc3-context` test suite pass.

### Related Issue

No existing issue — raising this to propose and discuss the feature. Happy to open a tracking issue if the maintainers prefer. This is intended as a starting point for discussion.

## Review Checklist

- [ ] **Issue**: If a change was made to the FDC3 Standard, was an issue linked above? *(feature raised for discussion; no issue yet)*
- [x] **API changes**: Does this PR include changes to any of the FDC3 APIs (`DesktopAgent`, `Channel`, `PrivateChannel`, `Listener`, `Bridging`)? — **No.** This adds library accessor functions to `@finos/fdc3-context`; the Desktop Agent API and wire protocols are unchanged.
- [x] **Context types**: Were new Context type schemas created or modified in this PR? — **No.** Existing schemas are read and exposed unchanged; code generation (`npm run build`) was run and the generated registry checked in.


kriswest (commented):
Please do raise an issue for this proposed change describing the use cases - we use issues for release milestone planning so all PRs should (for anything other than project governance/infrastructure maintenance) should have an associated issue.

Generating a copy of every schema as a javascript object to facilitate this access at runtime seems wasteful when the schema files themselves are bundled (https://github.com/finos/FDC3/blob/15931acc6f150e19fbd71fcd77d314cf49fee65a/packages/fdc3-context/tsconfig.json#L13) and could be loaded instead... There would be no need for a generation step reducing complexity.

However, one generation step is desired already for #1852, where an enum and union of the known types is desired (similar to your getContextTypes function). That could be dealt with here in fdc3-context (where should have been moved to previously) and then re-exported by fdc3-standard.

nchalaisRocket:
Thanks for the review @kriswest!

- **Issue**: raised #1993 describing the use cases (agents/protocol bridges, validation, dynamic UI/forms, docs tooling). Happy to refine it.
- **Avoiding a generated copy of the schemas**: agreed — that's wasteful given the schema files are already bundled/emitted to `dist/schemas` (and `resolveJsonModule` is enabled). I'll rework `getContextSchema`/`getAllContextSchemas` to load the bundled schemas instead of shipping a duplicate JS object, and drop the standalone generation step for the copy.
- **Aligning with #1852**: makes sense to fold this in. I'll move the auto-maintained `ContextTypes` enum / `ContextType` union generation into `fdc3-context` (derived from the schemas), re-export from `fdc3-standard`/`fdc3` to avoid a breaking change, and have `getContextTypes()` build on that single generation step rather than a separate one.

I'll push a revised implementation along those lines. Let me know if you'd prefer #1852 handled as a separate PR that this one builds on, or all together here.

kriswest:
> Let me know if you'd prefer https://github.com/finos/FDC3/issues/1852 handled as a separate PR that this one builds on, or all together here.

Personally, I'm happy to deal with both issues in one PR/review. I support the schema access use case and am very happy to see a new contributor taking it on, alongside overlapping work we've already committed to!

kriswest:
n.b. `isStandardContextType(type): boolean` is already a function in fdc3-standard that could be maintained in fdc3-context and reexported also

thorsent:
"Today the standardized context JSON Schemas are the single source of truth for context definitions" - this is not true. The schemas must represent the actual apps which are available in any given system (which will rarely conform to the standard definitions, and certainly wouldn't support all of them in any given configuration).

Agreed that such a feature would allow agents to interoperate with FDC3 in a very powerful way but it involves multiple changes to the specification, not a library commit that reads from hardcoded files:

1 - an expansion of appd to support context schemas (which would be loaded by desktop agents)
2 - specification of a standardized function where a desktop agent can be queried for available schemas

kriswest:
@thorsent makes a good point - this will of course only enable retrieval of the standard schemas. Apps can define their own context types and there would end up being no way to retrieve those schemas (a context does not link to its schema). There is a very old issue open for the AppD to be able to do that: https://github.com/finos/FDC3/issues/719 then as @thorsent indicates a desktop agent could make all schemas from all appDs it is aware of available.

However, I don't think that negates the value of improving access to the standard's own schemas. But, depending on your use case, further work my be needed!

thorsent:
@kriswest we wouldn't want to expose the standard schemas in totum (as implemented here) because it would imply that the running instance contains apps which support those contexts.

A coding agent might desire a complete set, which it could already obtain from the source code. But a runtime agent really needs to know what is actually available at runtime. Even if the hardcoded list were filtered by availability the net result would be that the agent has access to partial information at best, which would result (I think) in confused developers and agents.

I don't think it's that much extra specification to implement a more complete solution. The question is whether there's an appetite to expand the spec in this direction.

kriswest:
This proposal doesn't expose anything through the Desktop Agent API, only from the fdc3-context and fdc3 libs. I read this as relating to cases where you already know its one of the standardized types that you are working with and want to gather context (about the the context type) to generate a valid instance of a context type - rather than discovering what's in use through the Desktop Agent or AppD.. On that level it seems reasonable - but definitely does not solve a wider problem where a firm defines their own context - in our case we provide our own library for that includes our custom schemas and types (+the FDC3 standard ones for convenience).

Hence a question to @nchalaisRocket: is the problem you want to solve tightly scoped to the standard types. Or are you interested in a general solution? As discussed here a general solution is possible but would require us to evolve the AppD servers and Desktop Agent API to manage and expose additional data, which could then include custom/proprietary types. We've had a related issue open for many years, but the use cases have grown due to AI. There is another issue open that relates to that:
- https://github.com/finos/FDC3/issues/1713.

nchalaisRocket:
Thanks @kriswest and @thorsent, good points from both of you.

On scope: this PR is small on purpose. It just lets you read the *standard* context schemas straight from the library. The idea is simple. Say you already know you're working with an `fdc3.instrument` and you want its schema so you can build or check a valid one. It doesn't touch the Desktop Agent API or AppD, and it doesn't try to say anything about what a running system actually supports.

Tom, fair point on the "single source of truth" wording. I only meant that the schema files are the official definition of the *standard* types, not what's available at runtime. I'll fix that and make the scope clear.

Runtime discovery of what a DA or AppD actually offers, custom types included (#719 / #1713), is clearly the bigger goal, but that's a real spec change and out of scope here. I'd keep it separate. This PR just gives it a solid starting point for the standard types.
Based on your feedback, here's what I'll change:
- Drop the copied schemas and load the bundled files instead, so nothing is duplicated.
- Fold in #1852: generate the `ContextTypes` enum and the `StandardContextType` / `ExperimentalContextType` / `ContextType` types in `fdc3-context`, and re-export them from `fdc3-standard` / `fdc3` (no breaking change, and the enum is no longer deprecated).
- Move `isStandardContextType` into `fdc3-context` and re-export it too, good call Tom.
- Tidy up the docs and rebase to clear the conflict with `main`.

Does that sound reasonable? Standard schemas here, runtime discovery as its own thing.

thanks a lot for your kind reviews

kriswest:
This did sound reasonable to me - please do update it as described when you have a minute and we'll try and get it merged.

PR #1984, opened 2026-07-15
Add getAgent opt-in flags for own broadcasts and self intent resolution

### LINKED ISSUE #1751: Allow apps to decide if they receive their own broadcast messages
## Enhancement Request

Remove the recommendation that Desktop Agents don't deliver messages back to the sender and let the app implementor decide if they wish to receive them or not.

### Use Case

The FDC3 Standard _recommends_ (SHOULD) that DAs don't deliver broadcast messages back to the sender, as it forces some apps to examine context sharing messages before applying them, just to check if they came from themselves. This can / was observed  to cause flickering issues in a variety of apps when they changed context, broadcast that context, and immediately received a message back and reapplied the same context.

> DesktopAgent implementations SHOULD ensure that context messages broadcast to a channel by an application joined to it are not delivered back to that same application.

See: 
- https://fdc3.finos.org/docs/api/ref/DesktopAgent#broadcast
- https://fdc3.finos.org/docs/api/spec#desktop-agent-api-standard-compliance

Since then use cases have popped up for apps, composed of multiple widgets in the same page, to receive their own messages so that they can communicate with each other in the same way that they do other apps. This proposal doesn't solve routing of intents to handlers in the same app, but would make context sharing easier.

### Additional Information

Back when this was added to FDC3 (in FDC3 1.2, see #285) the only way to determine if you sent a message, was to do a deep comparison of the context message + assume no one else would send an identical one (there are few small use cases where it makes to resent a message - such as to reset a view).

With the metadata changes proposed for FDC3 3.0 (#1728) that comparison is no longer necessary, as we can simply check if the source of the broadcast was this app instance (compare instanceId). That would allow an argument to `getAgent` to control whether an app receives its own messages (for FDC3 for web implementation, and potentially preload implementation, either by patching the addContextListener function or passing the argument to the DA somehow).

### LINKED ISSUE #1860: Question: Resolving raised intents to the app that raised them
### Question Area

- [ ] App Directory
- [x] API
- [ ] Context Data
- [ ] Intents
- [ ] Use Cases
- [ ] Other

### Question

The spec states that the Agent SHOULD:

> Ensure that context messages broadcast by an application on a channel are not delivered back to that same application if they are joined to the channel.

But it doesn't mention anything similar for raising and handling intents.

Just to clarify: There is nothing in the spec that states that intent resolution should not select the app that raised the intent?

Roaders:
Implements #1751 and #1860 by adding two optional booleans to GetAgentParams:

- receiveOwnBroadcasts: when set, the Desktop Agent delivers an app's own broadcasts back to it. Defaults to false (current behaviour).
- resolveOwnIntents: when set, the raising app's own instance is considered when resolving an intent it raised. Defaults to false (current behaviour); other instances of the same app remain eligible either way.

The flags travel on the WCP4ValidateAppIdentity handshake message and are stored per-instance by the reference Desktop Agent (each instance of an app may connect with different getAgent options). BroadcastHandler makes its self-delivery filter conditional on receiveOwnBroadcasts; IntentHandler excludes the raising instance (by instanceId) unless resolveOwnIntents is set, dropping any AppIntent left with no resolvers so a self-only intent yields NoAppsFound.

Updates the API spec, getAgent, broadcast, Channel and raiseIntent/ raiseIntentForContext documentation (including net-new intent self-resolution wording), regenerates schema types, and adds conformance/BDD tests.

### Related Issue
fixes #1751 
fixes #1860 

## Review Checklist


toolbox/fdc3-for-web/fdc3-web-impl/src/handlers/IntentHandler.ts
```diff
@@ -439,29 +446,36 @@ export class IntentHandler implements MessageHandler {
...
+    const appIntents: AppIntent[] = uniqueIntentNames
+      .map(i => {
+        const directoryAppsWithIntent = matchingIntents.filter(mi => mi.intentName == i).map(mi => mi.appId);
+        const runningDirectoryApps = connectedApps.filter(ca => directoryAppsWithIntent.includes(ca.appId));
+        const appRegistrations = matchingRegistrations
+          .filter(registration => registration.intentName === i) // filter registrations for the current intent
+          .map(listener => ({ appId: listener.appId, instanceId: listener.instanceId, state: State.Connected }))
+          .filter(appRegistration => allIntents.every(intent => intent.appId !== appRegistration.appId)); // filter out apps that have intents registered in the directory
+
+        const runningApps: AppRegistration[] = [...runningDirectoryApps, ...appRegistrations].filter(
```
Roaders, line 458: bug confirmed in existing reference implementation. The app that raises the intent is considered as a target for the intent:

Image

Roaders: should we add a conformance test that ensures that an app will return a result rather than select itself as an intent target?

PR #2009, opened 2026-07-23
fix conformance tests

Roaders:
There seem to be some conformance test failures on main. This is my attempt at fixing them but there is still one test failing that I can't track down. Despite the error count being 1 none of the tests themselves seem to have a failure status or message....

I currently end up with this

image

with no signs of any errors when I scroll through the tests. That app that's open never closes.

### DIFF
--- a/packages/fdc3-agent-proxy/src/channels/DefaultChannel.ts
+++ b/packages/fdc3-agent-proxy/src/channels/DefaultChannel.ts
@@ -7,6 +7,7 @@ import {
   Channel,
   EventHandler,
   AppProvidableContextMetadata,
+  ChannelError,
 } from '@finos/fdc3-standard';
 import { Context } from '@finos/fdc3-context';
 import { Messaging } from '../Messaging.js';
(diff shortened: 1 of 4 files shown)


kriswest (commented):
Aha so the conformance test was correct (although there may be a different issue causing the progress to be out of whack) but the agent-proxy was not. Thanks for taking on the fix.

packages/fdc3-agent-proxy/src/channels/DefaultChannel.ts
```diff
@@ -154,7 +160,8 @@ export class DefaultChannel implements Channel {
     let listener: RegisterableListener;
     switch (type) {
       case 'contextCleared':
```
kriswest, line 162: I know its redundant but I prefer this so its ready for when we expand the list:

```suggestion
      case 'contextCleared':
        listener = new EventListener(this.messaging, 'contextCleared', handler);
        break;
```

packages/fdc3-agent-proxy/src/channels/DefaultChannelSupport.ts
```diff
@@ -193,6 +193,10 @@ export class DefaultChannelSupport implements ChannelSupport, Connectable {
   }
 
   async getOrCreate(id: string): Promise<Channel> {
+    if (typeof id !== 'string') {
```
kriswest: How about banning blank strings as channel names?
```suggestion
    if (typeof id !== 'string' || id.trim() === "") {
```

packages/fdc3-agent-proxy/src/listeners/AbstractListener.ts
```diff
@@ -76,6 +76,9 @@ export abstract class AbstractListener<X, Y extends SubscriptionRequest> impleme
     this.subscribeResponseType = subscribeResponseType;
     this.unsubscribeRequestType = unsubscribeRequestType;
     this.unsubscribeResponseType = unsubscribeResponseType;
+
+    //bind to allow destructuring
+    this.unsubscribe = this.unsubscribe.bind(this);
```
kriswest, line 81: good catch

kriswest (dismissed):
Lots of good catches, a couple of small suggestions

Roaders:
thanks @kriswest . suggestions applied

kriswest (approved):
LGTM

PR #2020, opened 2026-07-23
Add ContextWithMetadata to IntentResult

julianna-ciq:
## What changed

- Added `ContextWithMetadata` to the TypeScript `IntentResult` union.
- Updated the matching API reference documentation.

## Why

The `IntentResult` documentation explains that intent handlers may return `ContextWithMetadata`, but the documented and exported TypeScript union omitted it. This made the type declaration inconsistent with the supported handler result described immediately below it.

## Impact

TypeScript consumers can now use `ContextWithMetadata` as an `IntentResult`, and the published API reference accurately reflects the union.

## Validation

- `npm test --workspace packages/fdc3-standard` — 5 test files passed, 14 tests passed.
- `git diff --check`


kriswest (changes requested):
Apparently fixing this causing an example app to blow up as its expecting a type property on a non-void result (both channel and context have type). So a legitimate build fail apparently.

See the netlify build log: server-apps/security-demo-app1/src/app1.ts

kriswest:
Apologies this has picked up a couple of  conflicts in need of manual resolution

kriswest:
RE-review after rebase so we can read the docs

kriswest:
Need to address go and .NET type defintions to deal with IntentResolutionResult. Their definitions will be different...
Lets Build the IntentResolutionResult type rather than using a typescript utility type to do it, as that will better match up across languages....

kriswest (approved):
LGTM, but doesn't address the .NET and go binding updates fully. We can deal with that during the final pass for .NET and Go to sync them with FDC3 3.0 (@kemerava @bingenito)

PR #2043, opened 2026-07-30
Add channel ID to context cleared event

### LINKED ISSUE #2042: Add `channelId` to `FDC3ContextClearedEvent`
## Problem

When an application registers a listener using either:

```ts
desktopAgent.addEventListener("contextCleared", handler);
```

or:
```ts
channel.addEventListener("contextCleared", handler);
```

the resulting `FDC3ContextClearedEvent` only identifies the type of context that was cleared:

```ts
export interface FDC3ContextClearedEvent extends FDC3Event {
  readonly type: "contextCleared";
  readonly details: {
    contextType: string | null;
  };
}
```

For a listener registered through `DesktopAgent.addEventListener`, this does not tell the application which channel was cleared. An application may interact with its current User Channel and multiple App Channels, so `contextType` alone is insufficient to identify the affected channel.

Although a listener registered on a Channel object has an implicit association with that channel, using a different event shape would be inconsistent and would make events harder to handle uniformly.

Note that the Desktop Agent Communication Protocol already includes `channelId` in the `contextClearedEvent` payload:

```json
{
  "channelId": "channel-id",
  "contextType": "fdc3.instrument"
}
```

So, `channelId` is not currently part of the standard `FDC3ContextClearedEvent` interface, but is needed even in the DACP implementation. 

## Proposed change

Add `channelId` to `FDC3ContextClearedEvent.details`:

```ts
export interface FDC3ContextClearedEvent extends FDC3Event {
  readonly type: "contextCleared";
  readonly details: {
    channelId: string;
    contextType: string | null;
  };
}
```

For the sake of consistency, I would also recommend using the same event shape for both of these:

```ts
desktopAgent.addEventListener("contextCleared", handler);
channel.addEventListener("contextCleared", handler);
```

This means `channelId` should still be included when the listener was registered on a specific `Channel`, even though the channel is implicit from the object on which the listener was registered. Keeping the shapes identical:
- allows shared event handlers
- avoids API-specific variants of FDC3ContextClearedEvent
- aligns the standard API with the existing DACP event
- makes the source channel explicit when events are logged, forwarded, or processed asynchronously

### Area of Issue

- API
- Context Data

julianna-ciq:
Closes #2042

## What changed

- Added `channelId` to `FDC3ContextClearedEvent.details` in the standard TypeScript API.
- Updated the TypeScript and .NET API reference shapes.
- Required `contextCleared` events emitted through both Desktop Agent and Channel listeners to identify the cleared channel.
- Updated the API compliance and context-clearing documentation.
- Added a changelog entry.

## Why

A `DesktopAgent.addEventListener("contextCleared", handler)` listener cannot otherwise determine which channel was
cleared. The DACP `contextClearedEvent` payload already contains `channelId`, so the public standard API was out of
alignment with the wire protocol.

Using the same event shape for Desktop Agent and Channel listeners also allows applications to share handlers and makes
logged or asynchronously processed events self-describing.

## Validation

- `npm test --workspace packages/fdc3-standard`
- 5 test files passed; 14 tests passed
- `git diff --check`


website/docs/api/ref/Channel.md
```diff
@@ -760,7 +760,9 @@ Task ClearContext(string? contextType);
 </Tabs>
 
 Used to clear the specified context type if provided, otherwise, clear all context types present in the channel. The Desktop Agent MUST update its internal representation of the context in the channel and ensure that subsequent calls to [`getCurrentContext`](#getcurrentcontext) and any new joiners to that channel (through [`joinUserChannel`](DesktopAgent#joinUserChannel) or [`addContextListener`](DesktopAgent#addContextListener)) will not receive anything for either the specified context type or the most recent context until new context has been broadcast to the channel. 
-Desktop Agents MUST also immediately notify the apps that are listening to `contextCleared` event for this channel. If a `contextType` parameter was provided, then the `contextType` field will be set to that type, otherwise, it is omitted. 
+Desktop Agents MUST also immediately notify the apps that are listening to the `contextCleared` event for this
+channel. The event's `channelId` field MUST identify this channel. If a `contextType` parameter was provided, then the
```
julianna-ciq:
```suggestion
Desktop Agents MUST also immediately notify the apps that are listening to the `contextCleared` event for this channel. The event's `channelId` field MUST identify this channel. If a `contextType` parameter was provided, then the
```

website/docs/api/ref/Events.md
```diff
@@ -284,7 +288,12 @@ public class Fdc3ContextClearedEvent : Fdc3Event
 
 Type representing the format of `contextCleared`  events.
 
-The specific type of context is defined in the contextType field, which can be empty if we are clearing all the contexts on the channel.
+The channel on which context was cleared is identified by `details.channelId`. This field is included for events
+received through both `DesktopAgent.addEventListener` and `Channel.addEventListener`, allowing the same event handler
+to be used with either API.
```
julianna-ciq:
```suggestion
The channel on which context was cleared is identified by `details.channelId`. This field is included for events received through both `DesktopAgent.addEventListener` and `Channel.addEventListener`, allowing the same event handler to be used with either API.
```

website/docs/api/ref/Events.md
```diff
@@ -284,7 +288,12 @@ public class Fdc3ContextClearedEvent : Fdc3Event
...
+
+The specific type of context is defined in `details.contextType`, which will be `null` if all contexts on the channel
+were cleared.
```
julianna-ciq:
```suggestion
The specific type of context is defined in `details.contextType`, which will be `null` if all contexts on the channel were cleared.
```

website/docs/api/ref/Channel.md
```diff
@@ -760,7 +760,10 @@ Task ClearContext(string? contextType);
...
-Desktop Agents MUST also immediately notify the apps that are listening to `contextCleared` event for this channel. If a `contextType` parameter was provided, then the `contextType` field will be set to that type, otherwise, it is omitted. 
+Desktop Agents MUST also immediately notify the apps that are listening to the `contextCleared` event for this channel. The event's `channelId` field MUST identify this channel. If a `contextType` parameter was provided, then the
+`contextType` field will be set to that type; otherwise, it will be `null`. A Desktop Agent MAY deliver only events
+for the individual channel on which the listener was registered, or it MAY deliver events for all public channels. An
```
kriswest: I don't think we should allow that type of variation in behaviour. We should only deliver the events that you registered for and not others - delivering the events on channels you didn't register for them on would be confusing and I don't see a use case. 

```suggestion
`contextType` field will be set to that type; otherwise, it will be `null`. An
```

website/docs/api/ref/Events.md
```diff
@@ -284,7 +288,13 @@ public class Fdc3ContextClearedEvent : Fdc3Event
 
 Type representing the format of `contextCleared`  events.
 
-The specific type of context is defined in the contextType field, which can be empty if we are clearing all the contexts on the channel.
+The channel on which context was cleared is identified by `details.channelId`. This field is included for events received through both `DesktopAgent.addEventListener` and `Channel.addEventListener`, allowing the same event handler to be used with either API.
+
+Desktop Agents are compliant if they deliver only `contextCleared` events for the individual channel on which a
+listener was registered, or if they deliver events for all public channels. Applications that need to distinguish
+between channels must inspect `details.channelId`.
+
```
kriswest: As per previous comment, I think we should drop the below. Its promoting a weird variation in behaviour between agents and is not backed up by a use case.


```suggestion
```

website/docs/api/spec.md
```diff
@@ -171,7 +171,7 @@ A Desktop Agent's [`Channel`](ref/Channel) implementation **MUST**:
...
 - Return the most recent context (and associated metadata for [`getCurrentContextWithMetadata`](ref/Channel#getcurrentcontextwithmetadata)) matching the requested type, or `null` if no matching context is found.
 - Return `null` from [`getCurrentContext`](ref/Channel#getcurrentcontext) and [`getCurrentContextWithMetadata`](ref/Channel#getcurrentcontextwithmetadata) for context types that have been cleared via [`clearContext`](ref/Channel#clearcontext), until new context of that type is broadcast.
-- Notify listeners registered for the `contextCleared` event (via [`addEventListener`](ref/Channel#addeventlistener)) when [`clearContext`](ref/Channel#clearcontext) is called, including the `contextType` in the event if one was specified.
+- Notify listeners registered for the `contextCleared` event (via [`addEventListener`](ref/Channel#addeventlistener)) when [`clearContext`](ref/Channel#clearcontext) is called, including the channel's `channelId` and the `contextType` that was cleared, or `null` if all context types were cleared. Desktop Agents MAY deliver only events for the individual channel on which the listener was registered, or MAY deliver events for all public channels.
```
kriswest:
```suggestion
- Notify listeners registered for the `contextCleared` event (via [`addEventListener`](ref/Channel#addeventlistener)) when [`clearContext`](ref/Channel#clearcontext) is called, including the channel's `channelId` and the `contextType` that was cleared, or `null` if all context types were cleared. Desktop Agents MUST deliver only events for the individual channel on which the listener was registered.
```

kriswest (changes requested):
I think the option for DAs to deliver a wider scope of events than were registered for should be dropped.  Event listeners are registered on specific channels and delivering them for all channels would be confusing and invites variation between Desktop Agents making code that works consistently in all agents more complex to write. I don't see a use case for it etiher - and were there one I would suggest we provide a different way to register for events on all channels. Finally, it would violate principal that you need to know the channel name in order to interact with it that runs through the rest of the FDC3 Channel design - I may not be a huge fan of that design, but note that this would be inconsistent with it.

kriswest:
Small conflict to fix

kriswest (dismissed):
LGTM

kriswest (approved):
LGTM, but doesn't update the go definitions (@kemereva). We can deal with that in the final pass to sync the go definitions with all changes in FDC3 3.0

PR #2052, opened 2026-08-05
Clarify channel event types

julianna-ciq:
## Describe your change

Define a new `ChannelEventTypes` type containing the standardized `"contextCleared"` event and use it for `Channel.addEventListener` in both the public API and the agent proxy implementation.

This replaces the arbitrary `string` parameter with a dedicated literal type, making the supported normal-channel event set explicit and preventing Desktop Agent-scoped or private-channel event names from being registered on app or user channels at compile time.

Because `PrivateChannel` has its own disjoint `PrivateChannelEventTypes` set, its TypeScript interface now inherits the remaining `Channel` members through `Omit<Channel, 'addEventListener'>` and retains its existing listener override and runtime behavior.

The Channel, PrivateChannel, and Events reference documentation has been updated, including a new `ChannelEventTypes` section.

### Related Issue

No linked issue.

## Validation

- [x] `npm run build --workspace packages/fdc3-standard`
- [x] `git diff --check`

- [ ] I acknowledge that a contributor license agreement is required and that I have one in place or will seek to put one in place ASAP.

## Review Checklist

- [ ] **Issue**: If a change was made to the FDC3 Standard, was an issue linked above?
- [x] **Docs & Sources**: If yes, were both documentation (/docs) and sources updated?<br/> *JSDoc comments on interfaces and types should be matched to the main documentation in /docs*
- [ ] **Conformance tests**: If yes, are conformance test definitions (/toolbox/fdc3-conformance) still correct and complete?<br/> *Conformance test definitions should cover all **required** aspects of an FDC3 Desktop Agent implementation, which are usually marked with a **MUST** keyword, and optional features (**SHOULD** or **MAY**) where the format of those features is defined*
- [ ] **Schemas**: If yes, were changes applied to the Bridging and FDC3 for Web protocol schemas?<br/> *The Web Connection protocol and Desktop Agent Communication Protocol schemas must be able to support all necessary aspects of the Desktop Agent API, while Bridging must support those aspects necessary for Desktop Agents to communicate with each other*
- [ ] If yes, was code generation (`npm run build`) run and the results checked in?<br/> *Generated code will be found at `/src/api/BrowserTypes.ts` and/or `/src/bridging/BridgingTypes.ts`*
- [ ] Was code generation (`npm run build`) run and the results checked in?<br/> *Generated code will be found at `/src/context/ContextTypes.ts`*


julianna-ciq:
@kriswest - We noticed that the `Channel.addEventListener(type: string, ...)` was inaccurate. You couldn't just put any string value for "type", there are only a handful of approved "type" values. And then I saw that those types don't match between `Channel` and `PrivateChannel`, which cascaded to a bunch of changes (which is probably why the typedef wasn't tightened up before).

It's up to others if we think that tightening the `addEventListener` typedef is worth it.

kriswest:
You're right - this is not easy to fix with the current set of Event Type objects. We've got [FDC3EventTypes](https://fdc3.finos.org/docs/next/api/ref/Events#fdc3eventtypes) which is correct for DesktopAgent.addEventListener, nothing specific for `Channel.addEventListener` (uses only one of the values from FDC3EventTypes), then [`PrivateChannelEventTypes` ](https://fdc3.finos.org/docs/next/api/ref/Events#privatechanneleventtypes) which has only the `PrivateChannel` specific types and does not also include `contextCleared`, which is the one type `Channel` supports. Worse if you tighten this for `Channel`, `PrivateChannel` extends that and will conflict.

I think we need to make these changes to get this right (close to what you propose):

```ts
// Create a new ChannelEventTypes type
type ChannelEventTypes = "contextCleared";

// Have PrivateChannelEventTypes extend it
type PrivateChannelEventTypes =
  | ChannelEventTypes
  | "addContextListener"
  | "unsubscribe
  | "disconnect";
  
// Update Channel type
class Channel {
  addEventListener(
    type: ChannelEventTypes | null,
    handler: EventHandler
  ): Promise<Listener> {
    // ...
  }
}

// Then override it in PrivateChannel to widen the accepted types
class PrivateChannel extends Channel {
  override addEventListener(
    type: PrivateChannelEventTypes | null,
    handler: EventHandler
  ): Promise<Listener> {
    // ...
  }
}
```

I think you can do that without introducing the BaseChannel type that you have - what we were missing is that PrivateChannel also needs to support contextCleared.

kriswest:
Looks to me like NOT creating `BaseChannel` (because `PrivateChannel` can properly extend `Channel` and only widen the type) will significantly simplify this PR (won't need to add all those `| PrivateChannel` additions as A `PrivateChannel` is also a `Channel`. 

I'm not sure if any changes are needed in the DACP / AgentProxy, its possible this is already correct there - but it should be checked.

julianna-ciq:
@kriswest I've rewritten the typedefs to remove `BaseChannel`, and instead define the `ChannelEventTypes` to include `"contextCleared"`, and have `PrivateChannelEventTypes` extend that. This, of course, means that PrivateChannel now also supports `"contextCleared"`, so I have an implementation for that as well.

kriswest:
> This, of course, means that PrivateChannel now also supports "contextCleared", so I have an implementation for that as well.

Great, that was always intended.

packages/fdc3-agent-proxy/src/channels/DefaultPrivateChannel.ts
```diff
@@ -37,15 +40,37 @@ export class DefaultPrivateChannel extends DefaultChannel implements PrivateChan
         a = new PrivateChannelDisconnectEventListener(this.messaging, this.messageExchangeTimeout, this.id, handler);
         break;
       case null:
-        a = new PrivateChannelNullEventListener(this.messaging, this.messageExchangeTimeout, this.id, handler);
-        break;
+        return this.addAllEventListener(handler);
       default:
         throw new Error('Unsupported event type: ' + type);
```
kriswest: Not your change, but this could be `ChannelError.InvalidArguments`.

Also reported as not covered by a test in the coverage report.

packages/fdc3-agent-proxy/src/channels/DefaultPrivateChannel.ts
```diff
@@ -37,15 +40,37 @@ export class DefaultPrivateChannel extends DefaultChannel implements PrivateChan
...
+    try {
+      await privateChannelListener.register();
+    } catch (error) {
+      await channelListener.unsubscribe();
+      throw error;
+    }
+
+    return {
+      unsubscribe: async () => {
+        await Promise.all([channelListener.unsubscribe(), privateChannelListener.unsubscribe()]);
```
kriswest, line 76: Apparently this isn't covered by testing.

packages/fdc3-agent-proxy/src/channels/DefaultChannel.ts
```diff
@@ -156,14 +157,14 @@ export class DefaultChannel implements Channel {
...
       case 'contextCleared':
-        listener = new EventListener(this.messaging, 'contextCleared', handler);
+        listener = new EventListener(this.messaging, 'contextCleared', this.id, handler);
         break;
       case null:
-        listener = new EventListener(this.messaging, type, handler);
+        listener = new EventListener(this.messaging, type, this.id, handler);
         break;
       default:
         throw new Error('Unsupported event type: ' + type);
```
kriswest: Not your change, but this could be ChannelError.InvalidArguments.

Also reported as not covered by a test in the coverage report.

kriswest (dismissed):
This LGTM - there are just a couple of tweaks that could be applied:
- Test coverage for event unsubscription
- Changing the errors returned on invalid event types to ChannelError.InvalidArguments

kriswest:
Small conflict to fix

kriswest (approved):
LGTM

PR #2066, opened 2026-08-07
Remove unused FDC3 security dependency

julianna-ciq:
## What changed

- removes the unused `@finos/fdc3-security` dependency from `packages/fdc3`
- removes its depcheck baseline entry
- updates the lockfile

## Why

The `@finos/fdc3` roll-up entry point does not import or export `@finos/fdc3-security`, so consumers were installing it without receiving any API from it.

## Validation

- `npm run build` — passes for all workspaces
- `npm test` — passes for all workspaces
- root plus all 15 workspace depcheck invocations — pass


kriswest (approved):
LGTM

PR #2069, opened 2026-08-07
Remove unused CommonJS build dependency

julianna-ciq:
## Summary

- remove `@rollup/plugin-typescript`, which is not loaded by `rollup.config.mjs`
- retain and document `rollup`, which is invoked by the workspace build script

## Validation

- `npm run build`
- `npm test`
- `npx depcheck --skip-missing --quiet` at the repository root and in every workspace


kriswest (approved):
LGTM

PR #2072, opened 2026-08-07
Remove unused schema tslib dependency

julianna-ciq:
## Summary

- remove `tslib`, which is not referenced by source, compiler settings, or emitted code
- retain documented generator dependencies used through scripts or `createRequire()`

## Validation

- `npm run build`
- `npm test`
- `npx depcheck --skip-missing --quiet` at the repository root and in every workspace


kriswest (approved):
LGTM

PR #2091, opened 2026-08-11
feat(workbench): support FDC3 event listeners

### LINKED ISSUE #1674: Add support for listening to events in FDC3 Workbench
## Enhancement Request

FDC3 2.2 added `addEventListener` functions ([`DesktopAgent.addEventListener`](https://fdc3.finos.org/docs/api/ref/DesktopAgent#addeventlistener) and [`PrivateChannel.addEventListener`](https://fdc3.finos.org/docs/api/ref/PrivateChannel#addeventlistener)) to FDC3 which are not currently testable through FDC3 Workbench. Design and implement support for listening to events to FDC3 Workbench.

As events relate to specific concepts (e.g. the User channel changing or listeners being added or removed from private channels), support for listening for events should be integrated into the relevant tabs in the workbench. For example, the User channels tab should use the event to update its field for the currently selected channel (which it currently can't do!), while the private channels tab needs to switch from the deprecated event listeners to the new events and could/should report on the events in the UI. Additional listeners could also report on events through the system log.

Code copy buttons should be provided in each relevant tab with examples for listening to the events.

### Use Case:

Demonstrating the use of the new API events and providing the ability to test them.

### Additional Information

The (user) ChannelStore should (if we're running 2.2 / the function is available) listen for the UserChannelChanged event and update the current user channel, and log a message: https://github.com/finos/FDC3/blob/main/toolbox/fdc3-workbench/src/store/ChannelStore.ts

The PrivateChannelStore uses the deprecated events at: https://github.com/finos/FDC3/blob/e7700c3a44da3d8d3e06af2209ecdd777b3c6d83/toolbox/fdc3-workbench/src/store/PrivateChannelStore.tsx#L187-L252

These should still be used in 2.0 - 2.1, but 2.2 and we should switch to the new events in 2.2 and beyond. The event handlers also seem to ignore any context types in the addContextListener and unsubscribe events and assume all events relate to all context types, which needs looking at.

akira-in-tech:
## Describe your change

Add FDC3 2.2 event support to the Workbench while preserving compatibility with FDC3 2.0 and 2.1 implementations.

- Listen for userChannelChanged events and keep the displayed current user channel in sync.
- Use PrivateChannel.addEventListener for addContextListener, unsubscribe, and disconnect events on FDC3 2.2 and later.
- Retain the legacy Private Channel callbacks for FDC3 2.0 and 2.1.
- Preserve event details.contextType in logs and event history, and broadcast only contexts that match the registered listener type.
- Add event status, history, and copyable examples to the relevant Workbench tabs.
- Add Vitest coverage and a CHANGELOG entry.

### Related Issue

Resolves #1674

## Impact

The change is limited to the FDC3 Workbench. It does not modify the FDC3 API, schemas, conformance tests, context types, or intents.

## Validation

- npm run test --workspace fdc3-workbench: 7 tests passed
- npm run build --workspace fdc3-workbench: passed
- npm run lint --workspace fdc3-workbench: passed with 0 errors
- npm test: passed for all workspaces
- npm run lint: passed for all workspaces


toolbox/fdc3-workbench/src/store/PrivateChannelStore.tsx
```diff
@@ -189,83 +206,147 @@ class PrivateChannelStore {
...
-            }, channelContextDelay?.[key] ?? 0);
-          });
-        }
-      } catch {
-        systemLogStore.addLog({
-          name: 'pcAddContextListener',
-          type: 'error',
-          value: `Failed to add a context listener for '[all]' on channel [${channel.id}]`,
+      Object.entries(channelContexts ?? {})
+        .filter(([, context]) => contextType === null || context.type === contextType)
```
Sziszka90, line 236: I also started working on this issue, but you were quicker. It would be good to have the ability to assign issues to people.

The filter checks only: `contextType === null `but this WebFDC3 implementation can deliver undefined for an All listener. In that case, neither condition matches and every configured context is filtered out. It should work with something like this: `contextType == null || context.type === contextType`.

You can try to test the following way: 

App1: 
1. Open the Workbench.
2. Open the Intents tab.
3. Under Add intent listener, select an intent.
4. Enable Send intent result.
5. Select Channel result.
6. Select Private channel.
7. Add a context result row.
8. Select a context.
9. Click Add listener. App1 should log something like: `An intent listener for [StartCall] has been added`

App2: 
1. Open the Intents tab in App2.
2. Under Raise intent, select the context.
3. Under Intent Type, select the intent type that you used in App1.
4. Click Raise intent.
5. **App2 should display the returned private-channel result.**

toolbox/fdc3-workbench/src/store/PrivateChannelStore.tsx
```diff
@@ -189,83 +206,147 @@ class PrivateChannelStore {
...
   disconnect(channel: PrivateChannel) {
-    this.channelListeners.forEach(listener => {
-      this.removeContextListener(listener.id);
-    });
+    this.channelListeners
+      .filter(listener => listener.channelId === channel.id)
+      .forEach(listener => this.removeContextListener(listener.id));
     this.privateChannelsList = this.privateChannelsList.filter(chan => chan.id !== channel.id);
-    channel.disconnect();
+    this.channelEventListeners.delete(channel.id);
```
Sziszka90: Unsubscribing from the listeners should be considered.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
+* Added FDC3 Workbench support for `DesktopAgent` and `PrivateChannel` events, including automatic user channel updates, private channel event reporting, context-type-aware streaming, FDC3 2.0-2.1 compatibility, and copyable code examples. ([#1674](https://github.com/finos/FDC3/issues/1674))
```
Sziszka90: > better to add to the bottom of the list anyway (creates less conflicts)

toolbox/fdc3-workbench/src/store/PrivateChannelStore.tsx
akira-in-tech: Fixed in c2013e4. The FDC3 2.2 event payload now normalizes an omitted contextType to null, so an all-context listener broadcasts every configured context instead of filtering them all out. I added a regression test that delivers details: {} and verifies both configured contexts are broadcast. The Workbench suite passes with 9 tests.

toolbox/fdc3-workbench/src/store/PrivateChannelStore.tsx
akira-in-tech: Fixed in c2013e4. Event listener handles are now retained per private channel and unsubscribed with Promise.allSettled before a user-initiated disconnect. A remote disconnect also removes the stored handles and starts the same cleanup. The new regression test verifies that all three 2.2 event listeners are unsubscribed before channel.disconnect().

CHANGELOG.md
akira-in-tech: Fixed in c2013e4. I moved the Workbench event-support entry to the bottom of the Unreleased Added list to minimize conflicts with concurrent changelog updates.

akira-in-tech:
Synced the branch with the latest main in 7784423a to pick up the current website/build dependency baseline. The Netlify deploy preview is now green and returns HTTP 200. I also revalidated the Workbench with Node 24.18.0/npm 11.16.0: 9 tests pass, the production build passes, and lint reports 0 errors.

PR #2098, opened 2026-08-12
Bump ESLint to 9.39.5

julianna-ciq:
## Summary

Bump ESLint from ^9.32.0 to ^9.39.5 across the root and every workspace that directly declares it. ESLint 10 was evaluated but is not yet compatible with the repository's current lint rules/plugins.

## Validation

- npm run lint --workspaces --if-present
- npm run build
- tests for all directly affected workspaces


kriswest (approved):
LGTM

PR #2101, opened 2026-08-12
Bump Vite to 8.2.1

julianna-ciq:
## Summary

Bump Vite to 8.2.1 across all five workspaces that directly depend on it and update the shared lockfile.

## Validation

- npm run build
- tests for every directly affected workspace with a test script


kriswest (approved):
LGTM

PR #2111, opened 2026-08-12
Update remaining @finos/fdc3-schema dependencies

julianna-ciq:
## Summary

- bump ts-morph from 24.0.0 to 28.0.0
- remove message-await and replace its progress-only usage with a small local reporter
- update the shared npm lockfile

message-await 2.0.0 cannot be consumed because its published package declares dist/src/index.js as its entry point but does not include a dist directory. Removing it eliminates the remaining outdated dependency without changing generated output.

## Validation

- npm test --workspace packages/fdc3-schema
- npm run build

## Supersedes

Closes #2104. The Dependabot update cannot be used because message-await 2.0.0 was published without its declared entry-point files; this PR removes the dependency instead.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4356,32 +4356,47 @@
       }
     },
     "node_modules/@ts-morph/common": {
-      "version": "0.25.0",
+      "version": "0.29.0",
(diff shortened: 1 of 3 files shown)


kriswest (approved):
LGTM

PR #2117, opened 2026-08-13
fix(toolbox): resolve Vite 8 build warnings in conformance and workbench

### LINKED ISSUE #2115: Investigate and fix vite build warnings
## Vite 8 build warnings

Running `npm run build` from the repo root produces several warnings from Vite 8.2.0 across two workspaces: **fdc3-workbench** and **fdc3-conformance**. These should be addressed before the next release (or before Vite makes `configLoader: 'native'` the default).

---

### 1. `configLoader: 'native'` compatibility warnings

Both projects use patterns in their `vite.config.ts` that are unsupported by the upcoming native config loader.

**fdc3-workbench:**
```
(!) Your Vite config uses features that are unsupported by `configLoader: 'native'`:
  - ESM syntax in a file loaded as CommonJS (vite.config.ts:5:1).
    Use a `.mjs` extension or set `"type": "module"` in the closest package.json
```

**fdc3-conformance:**
```
(!) Your Vite config uses features that are unsupported by `configLoader: 'native'`:
  - `__dirname` (vite.config.ts:18:41). Use `import.meta.dirname` instead
```

**Suggested fix:** Add `"type": "module"` to each project's `package.json` (or rename config to `.mts`), and replace `__dirname` with `import.meta.dirname` in `fdc3-conformance/vite.config.ts`.

---

### 2. `COMMONJS_VARIABLE_IN_ESM` — `@jsdevtools/ono`

Appears twice (once per fdc3-workbench build environment — legacy and modern):

```
[COMMONJS_VARIABLE_IN_ESM] The CommonJS `module` variable is treated as a global variable
in an ECMAScript module and may not work as expected
   ╭─[ ../../node_modules/@jsdevtools/ono/esm/index.js:9:5 ]
   │
 3 │ export { Ono } from "./constructor";
   │ ───┬──
   │    ╰──── This file is considered to be an ECMAScript module because of the `export` keyword here:
   │
 9 │     module.exports = Object.assign(module.exports.default, module.exports);
   │     ───┬──
   │        ╰────
```

**Root cause:** `@jsdevtools/ono` ships an ESM entry (`esm/index.js`) that still uses `module.exports` — a CJS pattern in an ES module. This is an upstream bug in the package.

**Suggested fix:** Check if a newer version of `@jsdevtools/ono` (or its consumer — likely `@apidevtools/json-schema-ref-parser` / `$RefParser`) resolves the issue. Alternatively, add a Vite alias or patch to avoid the broken ESM entry.

---

### 3. Chunk size warning (fdc3-workbench)

```
(!) Some chunks are larger than 500 kB after minification. Consider:
- Using dynamic import() to code-split the application
- Use build.rolldownOptions.output.codeSplitting to improve chunking
- Adjust chunk size limit for this warning via build.chunkSizeWarningLimit.
```

The main bundle is ~1.9 MB minified (526 kB gzip). This is the workbench dev tool so it's less critical for production, but code-splitting heavy dependencies (MUI, jsoneditor, etc.) would help.

---

### 4. `SOURCEMAP_BROKEN` — `fix-source-map-support-global` plugin (fdc3-conformance)

```
[plugin fix-source-map-support-global] [SOURCEMAP_BROKEN] Sourcemap is likely to be incorrect:
a plugin (fix-source-map-support-global) was used to transform files, but didn't generate a
sourcemap for the transformation.
```

**Suggested fix:** Update the `fix-source-map-support-global` plugin to return a sourcemap from its transform hook, or suppress the warning if sourcemaps aren't needed for this output.

---

### 5. `PREFER_BUILTIN_FEATURE` — `@rollup/plugin-inject` (fdc3-conformance)

```
[PREFER_BUILTIN_FEATURE] Rolldown supports `inject` natively. Please refer
https://rolldown.rs/reference/ for more details. It is more performant than passing
`@rollup/plugin-inject` to plugins option.
```

**Suggested fix:** Replace `@rollup/plugin-inject` in the Vite config with Rolldown's built-in `inject` option (available in Vite 8 / Rolldown).

---

### 6. `PLUGIN_TIMINGS` — legacy plugin performance (fdc3-workbench)

```
[PLUGIN_TIMINGS] Your build spent 91% of 6.2s inside plugin hooks (5.6s).
  - vite:legacy-post-process renderChunk (87%, 5.4s, 2 calls)
```

The `@vitejs/plugin-legacy` renderChunk hook dominates build time. This is informational but worth noting if build speed becomes a concern — dropping legacy browser support (or targeting a higher baseline) would cut build time dramatically.

---

### 7. CSS injection entry file selection (fdc3-conformance)

```
[vite-plugin-css-injected-by-js] has identified "open-a.js" as one of the multiple output
files marked as "entry" to put the CSS injection code. However, if this is not the intended
file to add the CSS injection code, you can use the "jsAssetsFilterFunction" parameter to
specify the desired output file.
```

**Suggested fix:** Configure `jsAssetsFilterFunction` in `vite-plugin-css-injected-by-js` to target the correct entry (likely `fdc3-compliance.js`).

---

### Summary table

| # | Warning | Project | Severity |
|---|---------|---------|----------|
| 1 | `configLoader: 'native'` compat | workbench, conformance | Medium — will break in future Vite |
| 2 | `COMMONJS_VARIABLE_IN_ESM` | workbench | Medium — potential runtime bug |
| 3 | Chunk size > 500 kB | workbench | Low — dev tool only |
| 4 | `SOURCEMAP_BROKEN` | conformance | Low — test tool |
| 5 | `PREFER_BUILTIN_FEATURE` | conformance | Low — performance only |
| 6 | `PLUGIN_TIMINGS` (legacy) | workbench | Low — informational |
| 7 | CSS injection entry | conformance | Low — cosmetic |

MFA-G:
# Fix Vite 8 build warnings in `fdc3-conformance` and `fdc3-workbench`

Closes #2115 (warnings **1**, **4**, **5** and **7**).

After this change `toolbox/fdc3-conformance` builds **completely warning-free**, and the `configLoader: 'native'` warning is gone from `toolbox/fdc3-workbench` too.

### 1. `configLoader: 'native'` compatibility (warning #1)

**`fdc3-conformance`** used `__dirname` in `vite.config.ts`, which the native config loader does not provide:

```diff
+const __dirname = import.meta.dirname;
```

`import.meta.dirname` is available from Node 20.11, and the package already requires `>=22`.

**`fdc3-workbench`** loaded its ESM `vite.config.ts` as CommonJS. Added `"type": "module"` to its `package.json`, as suggested in the issue. The workbench has no `.js`/`.cjs` source files that could be affected — all sources are `.ts`/`.tsx` compiled by Vite, and `eslint.config.mjs` is already explicit.

### 2. `PREFER_BUILTIN_FEATURE` — `@rollup/plugin-inject` (warning #5)

Replaced the plugin with Rolldown's native `inject`, which lives under `transform` in `build.rollupOptions`:

```diff
-      plugins: [
-        inject({
-          process: 'process/browser.js',
-          Buffer: ['buffer', 'Buffer'],
-        }),
-      ],
+      transform: {
+        inject: {
+          process: 'process/browser.js',
+          Buffer: ['buffer', 'Buffer'],
+        },
+      },
```

The API is aligned with `@rollup/plugin-inject`, so the mapping is unchanged. `@rollup/plugin-inject` is dropped from `devDependencies`.

### 3. `SOURCEMAP_BROKEN` — `fix-source-map-support-global` (warning #4)

The build sets `sourcemap: true`, so suppressing the warning would have been the wrong trade-off — the plugin really was invalidating the sourcemap for `browser-source-map-support`. It now performs the same single replacement through `magic-string` and returns the generated map:

```ts
const s = new MagicString(code);
s.overwrite(index, index + SOURCE_MAP_SUPPORT_GLOBAL.length, SOURCE_MAP_SUPPORT_GLOBAL_FIXED);
return { code: s.toString(), map: s.generateMap({ hires: true }) };
```

It also returns early when the pattern is absent instead of silently producing identical output, so a future upstream change to `source-map-support` fails loudly rather than shipping a broken global shim. `magic-string` is added as an explicit `devDependency` (it was previously only reachable transitively via `@rollup/plugin-inject`).

### 4. CSS injection entry selection (warning #7)

Only `src/test/index.ts` (the `fdc3-compliance` entry) imports CSS — `import 'mocha/mocha.css'`. With ~20 entries the plugin was picking `open-a.js` arbitrarily, meaning the mocha stylesheet was injected by a mock app rather than by the test runner. Now pinned explicitly:

```ts
cssInjectedByJsPlugin({
  jsAssetsFilterFunction: outputChunk => outputChunk.fileName === 'fdc3-compliance.js',
})
```

Verified in `dist/lib`: after the change the mocha CSS and its `document.head.appendChild` call appear in `fdc3-compliance.js` only, and no longer in `open-a.js`.

## Not addressed

Deliberately left out of this PR, since they are judgement calls rather than fixes:

- **#2 `COMMONJS_VARIABLE_IN_ESM` (`@jsdevtools/ono`)** — comes from a transitive dependency's published ESM build (`@apidevtools/json-schema-ref-parser` → `@jsdevtools/ono`), not from repo code. Fixing it means patching or replacing a dependency.
- **#3 chunk size > 500 kB** and **#6 `PLUGIN_TIMINGS` (legacy plugin)** — both informational, and the issue itself notes #6 is a browser-support policy decision (dropping `@vitejs/plugin-legacy`) rather than a bug.

Happy to follow up on any of those in a separate PR if maintainers have a preference on the direction.

## Validation

Run locally on Node 22.23.2 / npm 12.0.2:

| Check | Result |
|---|---|
| `npm run build` (all workspaces) | pass |
| `npm test` (all workspaces) | pass |
| `npm run lint` (all workspaces) | pass |
| `npx prettier --check` on changed files | pass |
| `npx depcheck --skip-missing` in both toolbox workspaces | pass |
| `vite build` warnings in `fdc3-conformance` | **5 → 0** |
| `vite build` warnings in `fdc3-workbench` | `configLoader` warning gone; remaining are #2/#3/#6 above |

`CHANGELOG.md` updated under `[Unreleased] → Fixed`.


kriswest:
Hi @MFA-G thanks for having a crack at this. Could you take a look at teh Easy CLA comment and hit the [Please click here to be authorized](https://api.easycla.lfx.linuxfoundation.org/v2/repository-provider/github/sign/13399143/121826936/2117/#/?version=2) link to and complete a CAL so we can merge this?

toolbox/fdc3-conformance/package.json
```diff
@@ -27,10 +27,10 @@
     "util": "^0.12.5"
   },
   "devDependencies": {
-    "@rollup/plugin-inject": "^5.0.5",
     "@types/chai": "^4.3.0",
     "@types/mocha": "^10.0.10",
     "http-server": "^14.1.1",
+    "magic-string": "^0.30.17",
```
kriswest: THis is an old verison, current is 1.2.0.

The library is also not wonderfully well maintained (OpenSSF scorecard result is 3.8). Any chance we can eliminate it and use conventional string manipulation without a dependency?

kriswest (commented):
We could lose magic-string from this and a CLA needs completing before it can be merged.

@julianna-ciq could you review and merge if/when this gets updated?

MFA-G:
Thanks @kriswest — good call, `magic-string` is gone in b3fb788.

It was only there to hand back a sourcemap for the `fix-source-map-support-global` transform (returning `{ code }` with no `map` is what Vite 8 warns about). Since that transform rewrites **one expression on a single line** and changes no line numbers, the map is a plain identity map, which is a few lines of plumbing rather than a dependency:

```ts
function identitySourceMap(code: string, id: string) {
  const lineCount = code.split('\n').length;
  // VLQ segments: `AAAA` = [col 0, source 0, line 0, col 0]; `AACA` advances the source line by 1.
  const mappings = ['AAAA', ...Array(Math.max(lineCount - 1, 0)).fill('AACA')].join(';');
  return { version: 3 as const, sources: [id], sourcesContent: [code], names: [], mappings };
}
```

and the transform is back to a simple `code.replace(...)`.

Net effect on `toolbox/fdc3-conformance`: **no devDependency is added at all** — `@rollup/plugin-inject` (which was itself the thing dragging in `magic-string`, plus `estree-walker` and `@rollup/pluginutils`) is removed and replaced by Rolldown's built-in `transform.inject`, and nothing takes its place. The lockfile shrinks accordingly.

Validation on the updated branch:
- `npm run build` in `toolbox/fdc3-conformance` — clean, **zero warnings**, 332 modules, all 22 entries emitted with maps.
- Verified the transform still lands: the bundled `fdc3-compliance.js` contains `globalThis.define||function` and zero occurrences of the old `this.sourceMapSupport` form.
- `npm run build` in `toolbox/fdc3-workbench` — clean apart from the pre-existing chunk-size notice, which is unrelated to this PR.

The EasyCLA is being sorted out. @julianna-ciq ready for another look whenever convenient.

kriswest:
@MFA-G We'd be happy to move this PR along towards being merged, but can't do so unless you complete the CLA check. See comment https://github.com/finos/FDC3/pull/2117#issuecomment-5277610564

kriswest:
Apologies @MFA-G could you rebase and regenerate the lock file (`npm install`) to clear the conflict, otherwise ready to merge IMHO

MFA-G:
Rebased onto `main` and regenerated the lock file in 43eb0e41, @kriswest — the conflict is cleared and the PR shows as mergeable again.

The conflict was only in `package-lock.json`, so I resolved it by taking upstream's version wholesale and re-running `npm install`. The net effect on the lock file is just the removal of the `@rollup/plugin-inject` entries that this PR drops; no other dependency moved. `npm run build` across the workspaces is green, and the `fdc3-conformance` build is still warning-free apart from the pre-existing chunk-size notice.

On the CLA: that one needs a browser sign-in on my side and is not something I can complete from the CLI, so it is still outstanding. I will get it done — please don't let it block your review in the meantime.

kriswest (approved):
LGTM

kriswest:
@MFA-G Approved the PR (warning look to all have been cleared), just that CLA check to pass now!

PR #2157, opened 2026-08-28
Make ContextMetadata traceId optional

julianna-ciq:
## Describe your change

Makes `ContextMetadata.traceId` optional across the API schema, generated BrowserTypes and BridgingTypes, public typings, documentation, and conformance checks. Desktop Agents may generate a trace ID, so callers can construct an `IntentEventPayload` without supplying one.

### Related Issue

N/A — reported as a mismatch between the public `ContextMetadata` type and the generated browser protocol payload.

## Validation

- [x] `npm run build` in `packages/fdc3-schema`
- [x] `npm run build` in `packages/fdc3-standard`
- [x] `npm run build` in `toolbox/fdc3-conformance`
- [x] Verified public `ContextMetadata` without `traceId` is assignable to `BrowserTypes.IntentEventPayload.metadata`
- [x] Verified generated `IntentEvent` runtime conversion accepts metadata without `traceId`


kriswest (commented):
I expected traceId to be required in `ContextMetadata` as it was to be generated by the DA if not provided - and thought that was what was agreed in the working group. It is already optional in `AppProvidableContextMetadata` (https://fdc3.finos.org/docs/next/api/ref/Types#appprovidablecontextmetadata).

The docs are not currently in agreement with sources as they do say the DA *may* provide a traceId - but I think that is the error, not the sources requiring it on output. The conformance tests agree with the sources, whicch you change in this PR. Generation of a UIUD is if it was not provided seems a lesser burden on DAs than apps checking for its existence on using it. So I think we should be making the alternative change and updating the API overview content to require DAs to generate it if not present.

I note that the Agent Proxy is generating traceIds for raiseIntent if one is not provided at: https://github.com/finos/FDC3/blob/cb00631d6a62fa0a419d03caaa64dd2b111bd708/packages/fdc3-agent-proxy/src/intents/DefaultIntentSupport.ts#L177

The reference implementation generates one for a raiseIntent request if not provided: https://github.com/finos/FDC3/blob/cb00631d6a62fa0a419d03caaa64dd2b111bd708/toolbox/fdc3-for-web/fdc3-web-impl/src/handlers/IntentHandler.ts#L72
and for a result if not provided: https://github.com/finos/FDC3/blob/cb00631d6a62fa0a419d03caaa64dd2b111bd708/toolbox/fdc3-for-web/fdc3-web-impl/src/handlers/IntentHandler.ts#L232

I don't see the same on broadcast in the proxy: https://github.com/finos/FDC3/blob/cb00631d6a62fa0a419d03caaa64dd2b111bd708/packages/fdc3-agent-proxy/src/channels/DefaultChannel.ts#L64

The reference implementation handles generation for a broadcast: https://github.com/finos/FDC3/blob/cb00631d6a62fa0a419d03caaa64dd2b111bd708/toolbox/fdc3-for-web/fdc3-web-impl/src/handlers/BroadcastHandler.ts#L475

While I think we should go the other way here and require DAs to generate the traceId if not present - were we to go this way, more work is required IMHO,

kriswest:
Can we get an issue open for this.

kriswest (commented):
How do you feel about closing this in favour of #2198 which makes the traceId generation required for DAs (as I thought originally intended) @julianna-ciq

julianna-ciq:
Closing in favor of #2198

PR #2193, opened 2026-09-09
feat(dacp): contextCleared listener registration, context-clearing docs & conformance tests (#2164, #2165, #2187)

### LINKED ISSUE #2164: [DACP] Support contextCleared listener registration for DesktopAgent and Channel
## Problem

FDC3 defines a [`contextClearedEvent`](https://github.com/finos/FDC3/blob/main/packages/fdc3-schema/schemas/api/contextClearedEvent.schema.json), and the public API permits applications to register either of the following listeners:

```ts
desktopAgent.addEventListener("contextCleared", handler);
channel.addEventListener("contextCleared", handler);
```

However, the DACP [`AddEventListenerRequestPayload`](https://github.com/finos/FDC3/blob/main/packages/fdc3-schema/schemas/api/addEventListenerRequest.schema.json) currently supports only the user-channel-changed event and has no field identifying a Channel.

The reference proxy's `Channel.addEventListener("contextCleared", ...)` implementation registers a local filter for incoming events but sends no listener-registration request to the Desktop Agent. Its Desktop Agent event-listener implementation also rejects `contextCleared` as an unknown event type.

As a result, there is no standard DACP exchange by which the proxy can tell the Desktop Agent that an app has registered a `contextCleared` listener, or identify the Channel to which a Channel-scoped registration applies. A Desktop Agent would have to send all context-cleared events to all connected apps and rely on local filtering, but that behavior is not specified.

## Proposed resolution

Define a complete DACP registration mechanism for `contextCleared` listeners. Possible approaches include:

1. Extend `addEventListenerRequest` with a `CONTEXT_CLEARED` event type and an optional or nullable `channelId`; or
2. Add a Channel-specific event-listener request and response.

The chosen design should define:

- Registration through both `DesktopAgent.addEventListener` and `Channel.addEventListener`.
- Channel scoping for Channel listeners.
- The scope of Desktop Agent listeners, including the relationship to the current User Channel.
- Wildcard listener behavior.
- Unsubscription behavior.
- Whether the Desktop Agent routes events only to registered apps or sends them broadly for proxy-side filtering.

## Acceptance criteria

- Both public forms of `contextCleared` listener registration can be represented over DACP.
- The Desktop Agent has enough information to route the event according to the specified scope.
- Schemas, generated types, the reference proxy, the reference web implementation, and tests are updated consistently.

### Related work

- #1197
- #2042
- #1642

### Area of issue

- API
- FDC3 for Web Browsers
- Desktop Agent Communication Protocol

### LINKED ISSUE #2165: [DACP docs] Add clearContext and contextCleared message exchanges
## Problem

The Desktop Agent Communication Protocol documentation states that DACP encompasses all API calls and events defined by the Desktop Agent API. However, its Channel section does not document the context-clearing messages introduced with `Channel.clearContext()`:

- [`clearContextRequest`](https://github.com/finos/FDC3/blob/main/packages/fdc3-schema/schemas/api/clearContextRequest.schema.json)
- [`clearContextResponse`](https://github.com/finos/FDC3/blob/main/packages/fdc3-schema/schemas/api/clearContextResponse.schema.json)
- [`contextClearedEvent`](https://github.com/finos/FDC3/blob/main/packages/fdc3-schema/schemas/api/contextClearedEvent.schema.json)

The `addEventListener()` section similarly lists `channelChangedEvent` but does not describe `contextClearedEvent`.

The omission also leaves the encoding of the optional public API argument unclear. The public method is:

```ts
channel.clearContext(contextType?: string): Promise<void>
```

but `ClearContextRequestPayload` requires `contextType: string | null`, and the reference proxy normalizes an omitted argument to `null`. Other DACP requests use similar normalization, but no general omitted-versus-null rule is documented.

## Proposed resolution

Update the DACP documentation to include:

- The `clearContextRequest` and `clearContextResponse` exchange.
- The `contextClearedEvent` used to notify listeners.
- Listener registration, routing, and unsubscription behavior for Desktop Agent and Channel listeners.
- The meaning of `channelId` and `contextType` in each message.
- A normative statement that an omitted `clearContext()` argument is encoded as `contextType: null`, if the current required-and-nullable schema is intentional.

It would also be helpful to add a protocol-wide convention describing when an omitted optional API argument is represented by an absent field and when it is normalized to `null` or another sentinel value.

## Acceptance criteria

- Every context-clearing request, response, and event schema is linked from the DACP documentation.
- A complete message sequence is documented.
- Omitted-versus-null encoding is unambiguous.
- The documentation agrees with the schema and reference proxy.

### Related work

- #1197
- #1642

### Area of issue

- Documentation
- FDC3 for Web Browsers
- Desktop Agent Communication Protocol

### LINKED ISSUE #2187: clearContext / getCurrentContext Inconsistency
### Area of Issue

- [ ] App Directory
- [x] API
- [ ] Context Data
- [ ] Intents
- [ ] Desktop Agent Bridging
- [ ] Use Cases
- [ ] Other

### Issue Description

A three-way spec inconsistency blocks writing assertions. On what getCurrentContext returns for a cleared type: the normative MUST list says null (line 174); the prose at line 947 says the fdc3.nothing type is returned instead; line 326 says the cleared type "will be recorded in the field subType of fdc3.nothing". That subType field exists nowhere in the context schemas, the generated types, the proxy or web-impl, and nothing.schema.json does not define it.

kriswest:
## Describe your change

This PR resolves three related issues around clearing context on channels and the `contextCleared` event.

### #2164 — DACP registration mechanism for `contextCleared` listeners

The public API allows registering a `contextCleared` listener via both `DesktopAgent.addEventListener` and `Channel.addEventListener`, but the DACP `AddEventListenerRequestPayload` only supported `USER_CHANNEL_CHANGED` and had no way to identify a Channel. The reference proxy registered Channel `contextCleared` listeners locally and never told the Desktop Agent, and its Desktop Agent listener path rejected `contextCleared` outright.

Using approach 1 from the issue, this PR:

- Adds a `CONTEXT_CLEARED` value to the `FDC3EventType` enum and a required, nullable `channelId` to `AddEventListenerRequestPayload`. `channelId` is the Channel's id for a Channel-scoped registration (via `Channel.addEventListener`) or `null` for a Desktop Agent-level registration (via `DesktopAgent.addEventListener`), whose scope follows the app's current User channel.
- Regenerates `BrowserTypes.ts` from the schema.
- Updates the reference proxy: the Desktop Agent listener now handles `CONTEXT_CLEARED` and translates `contextClearedEvent` into an `FDC3ContextClearedEvent`; the Channel-level listener now registers over DACP (carrying its `channelId`) instead of filtering purely locally.
- Updates the reference web implementation to store each registration's scope, handle `clearContextRequest`, and route `contextClearedEvent` only to matching registered apps (exact channel for Channel-scoped listeners; current User channel for Desktop Agent-level listeners; wildcard listeners included; the originating app is excluded).
- Aligns `FDC3ContextClearedEvent.details.channelId` with the nullable schema payload.
- Adds unit-test coverage in `fdc3-agent-proxy` and `fdc3-web-impl`.

### #2165 — Document context-clearing in the DACP spec

Adds documentation to the Desktop Agent Communication Protocol spec for the `clearContextRequest`/`clearContextResponse` exchange and the `contextClearedEvent`, including listener registration, routing, scope and unsubscription behaviour for both Desktop Agent and Channel listeners, the meaning of `channelId`/`contextType`, and message-sequence diagrams. Also adds a protocol-wide convention describing when an omitted optional API argument is encoded as an absent field versus normalized to `null` (required-and-nullable fields such as `clearContext()`'s `contextType` are normalized to `null`).

### #2187 — Fix `getCurrentContext`-after-clear spec inconsistency

Corrects a three-way inconsistency in the API spec so all sources agree that `getCurrentContext` returns `null` for a cleared type until new context is broadcast, removing the contradictory prose that referenced a non-existent `subType` field on `fdc3.nothing`.

### Conformance tests

- Adds User Channel `clearContext` conformance test definitions (`UCClearContext1`/`UCClearContext2`) mirroring the existing App Channel definitions.
- Implements all four (`ACClearContext1/2`, `UCClearContext1/2`) in `toolbox/fdc3-conformance`, reusing the existing ChannelsApp mock. Each verifies the `contextCleared` event is delivered with the correct `channelId`/`contextType` and that `getCurrentContext` returns `null` for a cleared type. **These pass at 100% against the reference implementation.**
- Fixes a conformance-runner bundling regression where `node:`-prefixed builtins (e.g. mocha's `require('node:util').inherits`) were externalized to empty browser stubs, preventing the test-suite menu from loading.

### Related Issue

resolves #2164
resolves #2165
resolves #2187

## Review Checklist

  - [x] **Docs & Sources**: If yes, were both documentation (/docs) and sources updated?
  - [x] **Conformance tests**: If yes, are conformance test definitions (/toolbox/fdc3-conformance) still correct and complete?
  - [x] **Schemas**: If yes, were changes applied to the Bridging and FDC3 for Web protocol schemas?
    - [x] If yes, was code generation (`npm run build`) run and the results checked in?


toolbox/fdc3-for-web/fdc3-web-impl/src/handlers/BroadcastHandler.ts
```diff
@@ -488,6 +545,19 @@ export class BroadcastHandler implements MessageHandler {
...
+    const channelId = arg0.payload.channelId;
+    const contextType = arg0.payload.contextType ?? null;
+
+    const channel = this.getChannelById(channelId);
+    if (channel) {
+      channel.context = contextType ? channel.context.filter(c => c.context.type !== contextType) : [];
+    }
+
+    this.fireContextClearedEvent(channelId, contextType, sc, from);
+    successResponse(sc, arg0, from, {}, 'clearContextResponse');
```
julianna-ciq: An unknown `channelId` currently falls through this block, emits a `contextClearedEvent`, and returns success even though nothing was cleared. Please reject it consistently with `handleGetCurrentContextRequest` and add a regression test asserting that no event is emitted.

```suggestion
    if (!channel) {
      errorResponse(sc, arg0, from, ChannelError.NoChannelFound, 'clearContextResponse');
      return;
    }

    channel.context = contextType ? channel.context.filter(c => c.context.type !== contextType) : [];
    this.fireContextClearedEvent(channelId, contextType, sc, from);
    successResponse(sc, arg0, from, {}, 'clearContextResponse');
```

website/docs/api/spec.md
```diff
@@ -945,7 +945,7 @@ To facilitate context linking in such situations it is recommended that applicat
...
+Channel interface provides the ability to [`clearContext`](ref/Channel.md#clearcontext) on the channel, either for the specific context type, if provided, or for all contexts on that channel. Applications may listen to the `contextCleared` event through either the Channel or Desktop Agent interface. The event's `channelId` field identifies the channel that was cleared. If a specific type was cleared, the `contextType` field of the event will be set to that type; otherwise, it will be `null`. Once cleared, any apps that join the channel, add new context listeners or call [`getCurrentContext`](ref/Channel.md#getcurrentcontext) will not return null to the caller until new context is broadcast to the channel.
```
julianna-ciq: `will not return null` reverses the intended resolution of #2187. Joining or registering a listener produces no retained context, while `getCurrentContext()` returns `null` for a cleared type. Suggested wording:

```suggestion
The `Channel` interface provides the ability to [`clearContext`](ref/Channel.md#clearcontext) on the channel, either for a specific context type, if provided, or for all contexts on that channel. Applications may listen to the `contextCleared` event through either the Channel or Desktop Agent interface. The event's `channelId` field identifies the channel that was cleared. If a specific type was cleared, the `contextType` field of the event will be set to that type; otherwise, it will be `null`. Once cleared, [`getCurrentContext`](ref/Channel.md#getcurrentcontext) returns `null` for the cleared context, and applications that join the channel or add new context listeners receive no retained context until new context is broadcast to the channel.
```

packages/fdc3-agent-proxy/test/features/app-channels.feature
```diff
@@ -64,6 +64,55 @@ Feature: Channel Listeners Support
     And I call "{channel1}" with "addContextListener" with parameters "{null}" and "{true}"
     Then "{result}" is an error
 
+  Scenario: Adding a contextCleared event listener registers it with the Desktop Agent, scoped to the channel
```
julianna-ciq, line 67: These new scenarios cover only `Channel.addEventListener`. #2164 also requires the `DesktopAgent.addEventListener('contextCleared', ...)` path, and the new translation/registration branches in `DesktopAgentEventListener.ts` remain uncovered. Please add equivalent proxy scenarios—most naturally in `user-channels.feature`—covering: `CONTEXT_CLEARED` registration with `channelId: null`, wildcard registration, translation of an incoming `contextClearedEvent` into `{ type: 'contextCleared', details: { channelId, contextType } }`, and unsubscription. For example, the registration assertion should include:

```gherkin
When I call "{api}" with "addEventListener" with parameters "contextCleared" and "{typesHandler}"
Then messaging will have posts
  | payload.type    | payload.channelId | matches_type            |
  | CONTEXT_CLEARED | {null}            | addEventListenerRequest |
```

website/docs/api/specs/desktopAgentCommunicationProtocol.md
```diff
@@ -139,20 +148,51 @@ Request and response for removing the context listener ([`Listener.unsubscribe()
...
+
+The `contextClearedEvent` is delivered to registered listeners when context is cleared on a channel via [`Channel.clearContext()`](#clearcontext) (see the [`Channel`](#channel) section below). Its payload carries:
+
+- `channelId`: the id of the channel on which context was cleared.
+- `contextType`: the context type that was cleared, or `null` when all context types on the channel were cleared.
+
+**Routing.** The Desktop Agent has enough information from each registration's `type` and `channelId` to route events only to the apps that registered a matching listener, rather than broadcasting events to all connected apps for proxy-side filtering. A `contextClearedEvent` for a given channel MUST be routed to an app when that app has a registration whose `type` is `CONTEXT_CLEARED` (or `null` for a wildcard listener) and whose scope matches — either a Channel-scoped registration for the same channel, or a Desktop Agent-level registration held by an app whose current User channel is the cleared channel. Desktop Agents SHOULD NOT deliver a `contextClearedEvent` back to the app that triggered the clear.
```
julianna-ciq: The implementation and conformance test make origin suppression mandatory in practice, but the DACP only says `SHOULD NOT`, and the public `Channel.clearContext` contract does not mention the exception. Please settle this at the public API level. If origin suppression is intended, make it normative here and document the same behavior on `Channel.clearContext`; otherwise remove the implementation filter and conformance expectation. For the former choice, the DACP wording could be:

```suggestion
**Routing.** The Desktop Agent has enough information from each registration's `type` and `channelId` to route events only to the apps that registered a matching listener, rather than broadcasting events to all connected apps for proxy-side filtering. A `contextClearedEvent` for a given channel MUST be routed to an app when that app has a registration whose `type` is `CONTEXT_CLEARED` (or `null` for a wildcard listener) and whose scope matches — either a Channel-scoped registration for the same channel, or a Desktop Agent-level registration held by an app whose current User channel is the cleared channel. The Desktop Agent MUST NOT deliver a `contextClearedEvent` back to the app instance that triggered the clear.
```

julianna-ciq (changes requested):
The protocol design is directionally sound, but the unknown-channel behavior, cleared-context wording, DesktopAgent-level proxy coverage, and self-notification contract need to be resolved before merge. Details and suggested changes are inline.

I have not duplicated the conformance-runner depcheck fix here because #2198 already carries the package-local configuration for the same shared polyfills; this branch should pick that up when the overlapping change is merged or rebased.

toolbox/fdc3-for-web/fdc3-web-impl/src/handlers/BroadcastHandler.ts
kriswest: Resolving this - although I'm also noting that the addition of an event listener on the unknown channel would be required to receive the event/ That should also fail and I don't think it is. Looking at both issues

toolbox/fdc3-conformance/package.json
```diff
@@ -20,7 +20,9 @@
     "@finos/fdc3": "3.0.0-alpha.2",
     "buffer": "^6.0.3",
     "chai": "^6.2.2",
+    "events": "^3.3.0",
     "mocha": "^11.8.0",
+    "path-browserify": "^1.0.1",
```
julianna-ciq, line 25: I assume this is being side-loaded by Vite using vite.config.ts. If so, that fact should be documented in `toolbox/fdc3-conformance/package.json`. That will address the depcheck errors.

julianna-ciq: Note - looks like this is being handled in #2198

julianna-ciq:
FYI @kriswest — I opened #2213 with the changes from my review (including the depcheck fix), in case taking them as a stacked PR is more convenient.

kriswest:
@julianna-ciq ready for re-review. Integrated your changes with mine and pushed.

julianna-ciq (approved):
lgtm

PR #2197, opened 2026-09-09
fix(dacp): enforce getCurrentContextResponse context/metadata invariant (#2166)

### LINKED ISSUE #2166: [DACP] Define the context/metadata invariant in getCurrentContextResponse
## Problem

The DACP [`GetCurrentContextSuccessResponsePayload`](https://github.com/finos/FDC3/blob/main/packages/fdc3-schema/schemas/api/getCurrentContextResponse.schema.json) requires `context`, but permits `metadata` to be omitted or set to `null` even when a non-null context is returned.

The same `getCurrentContextRequest` and `getCurrentContextResponse` exchange is used to implement both:

```ts
channel.getCurrentContext(contextType?)
channel.getCurrentContextWithMetadata(contextType?)
```

The request does not identify which public method initiated it. Therefore, the Desktop Agent cannot know that it is safe to omit metadata for a particular request. To support `getCurrentContextWithMetadata()`, it must provide the associated metadata whenever it returns a context.

This also conflicts with the Channel compliance requirement to retain each current context together with its associated `ContextMetadata`. The reference proxy currently compensates for missing response metadata by fabricating fallback values such as an unknown source and the response timestamp, which loses the actual provenance of the retained context.

## Proposed resolution

Define and enforce this response invariant:

- When `payload.context` is non-null, `payload.metadata` MUST contain the complete `ContextMetadata` associated with that context.
- When `payload.context` is null, `payload.metadata` MUST be null or omitted, with one canonical representation documented.

Update the JSON Schema with an appropriate conditional constraint and clarify the behavior in the DACP documentation.

## Acceptance criteria

- The relationship between `context` and `metadata` is normative and schema-enforced.
- `getCurrentContextWithMetadata()` never requires fabricated metadata.
- The reference proxy and Desktop Agent implementation follow the invariant.
- Tests cover both non-null and null context responses.

### Area of issue

- API
- FDC3 for Web Browsers
- Desktop Agent Communication Protocol

kriswest:
## Describe your change

Resolves the `getCurrentContextResponse` metadata ambiguity described in #2166.

The same `getCurrentContextRequest`/`getCurrentContextResponse` exchange implements both `Channel.getCurrentContext()` and `Channel.getCurrentContextWithMetadata()`, and the request does not identify which public method initiated it. The schema previously required `context` but allowed `metadata` to be omitted or `null` even when a non-null context was returned. As a result the Desktop Agent couldn't know it was safe to omit metadata, and the reference proxy compensated by fabricating fallback metadata (an `unknown` source, the response timestamp, an empty `traceId`), losing the retained context's real provenance and conflicting with the Channel compliance requirement to retain each context together with its `ContextMetadata`.

### Changes

- **Schema** (`getCurrentContextResponse.schema.json`): the `context`/`metadata` relationship is now a normative, schema-enforced invariant via a conditional (`if`/`then`/`else`) constraint:
  - when `payload.context` is non-null, `payload.metadata` MUST be present and contain the complete `ContextMetadata` for that context;
  - when `payload.context` is `null`, `payload.metadata` MUST be `null` (the single canonical representation).
- **Generated types**: regenerated `BrowserTypes.ts` and updated the field documentation to describe the invariant.
- **Reference proxy** (`DefaultChannel.getCurrentContextWithMetadata`): no longer fabricates fallback metadata. It uses the metadata guaranteed by the invariant (preserving actual provenance) and guards against a non-conformant agent with `ChannelError.MalformedContext`.
- **Reference web implementation**: already honoured the invariant (context and metadata are retained together and returned as a pair, or both `null`); covered by a new test.
- **DACP documentation**: the `getCurrentContext()` / `getCurrentContextWithMetadata()` section now states the invariant normatively and explains why the Desktop Agent must always return metadata alongside a non-null context.
- **Tests**: proxy `getCurrentContextWithMetadata` scenarios cover both the non-null branch (asserting the real `source.instanceId`, `traceId`, `signature` and `custom` are passed through, proving no fabrication) and the null branch; the web implementation gains a null-context ⇒ null-context + null-metadata scenario. The schema constraint was also validated directly across all valid/invalid combinations.
- **Conformance runner fix**: maps `node:`-prefixed builtins (e.g. mocha's `require('node:util').inherits`) to their browser polyfills so the runner loads (the same bundling fix applied on the `contextCleared` branch), allowing the suite to run. **The conformance suite passes 100% (120 tests) against the reference implementation.**

The commit also includes deterministic generator reflow of `ContextTypes.ts` and `BrowserTypes.ts` (produced by `npm run build`), checked in so the generated output stays stable.

### Related Issue

resolves #2166

## Review Checklist

  - [x] **Docs & Sources**: If yes, were both documentation (/docs) and sources updated?
  - [x] **Conformance tests**: If yes, are conformance test definitions (/toolbox/fdc3-conformance) still correct and complete?
  - [x] **Schemas**: If yes, were changes applied to the Bridging and FDC3 for Web protocol schemas?
    - [x] If yes, was code generation (`npm run build`) run and the results checked in?


packages/fdc3-agent-proxy/src/channels/DefaultChannel.ts
```diff
@@ -109,13 +110,23 @@ export class DefaultChannel implements Channel {
 
     const context = response.payload.context;
     if (context) {
+      // Per the getCurrentContextResponse invariant, a non-null context MUST be accompanied by
+      // its complete ContextMetadata, so we use it directly and preserve the actual provenance
+      // rather than fabricating fallback values.
+      throwIfUndefined(
+        response.payload.metadata ?? undefined,
```
julianna-ciq: This guards only the non-null-context/missing-metadata case. `AbstractMessaging.exchange()` does not perform runtime schema validation, so the proxy still silently accepts other violations of the new canonical invariant: `{ context: null }`, `{ context: null, metadata: {...} }`, or an omitted `context`. `getCurrentContext()` also converts an omitted context to `null` and ignores an inconsistent metadata value. Please parse and validate the pair in a shared helper used by both methods, returning `ChannelError.MalformedContext` for every invalid combination. The core check could be structured as:

```ts
const { context, metadata } = response.payload;

if (context === null) {
  if (metadata !== null) {
    throw new Error(ChannelError.MalformedContext);
  }
  return null;
}

if (context === undefined || metadata == null) {
  throw new Error(ChannelError.MalformedContext);
}

const resultMetadata: ContextMetadata = metadata;
return { context, metadata: resultMetadata };
```

packages/fdc3-schema/schemas/api/getCurrentContextResponse.schema.json
```diff
@@ -48,17 +49,33 @@
...
 				}
 			},
 			"required": [
-				"context"
+				"context",
+				"metadata"
 			],
-			"additionalProperties": false
+			"additionalProperties": false,
+			"if": {
```
julianna-ciq, line 64: Please add committed schema-validation tests for this conditional rather than relying only on conforming feature fixtures or manual validation. The matrix should accept a context object with complete metadata and `null`/`null`, and reject at least: context object with null or omitted metadata; null context with object or omitted metadata; and an omitted context. That will ensure later schema generation/refactoring cannot weaken the invariant unnoticed.

packages/fdc3-agent-proxy/test/features/broadcast.feature
```diff
@@ -60,8 +60,8 @@ Feature: Broadcasting
     And I call "{channel1}" with "broadcast" with parameter "{instrumentContext}"
     And I call "{channel1}" with "getCurrentContextWithMetadata" with parameter "fdc3.instrument"
     Then "{result}" is an object with the following contents
-      | context.type    | context.name | metadata.source.appId | metadata.traceId | metadata.signature.signature      | metadata.signature.protected      | metadata.custom.key |
-      | fdc3.instrument | Apple        | test-app              | test-trace-id    | test-signature (signature part) | test-signature (protected part) | value               |
+      | context.type    | context.name | metadata.source.appId | metadata.source.instanceId | metadata.traceId | metadata.signature.signature      | metadata.signature.protected      | metadata.custom.key |
+      | fdc3.instrument | Apple        | test-app              | test-instance              | test-trace-id    | test-signature (signature part) | test-signature (protected part) | value               |
```
julianna-ciq: The contract says all retained app-provided metadata is preserved, including `antiReplay`, but this positive test does not supply or assert it. Please add matching `antiReplay` values to the `ChannelState` fixture and assert them here, for example:

```suggestion
      | context.type    | context.name | metadata.source.appId | metadata.source.instanceId | metadata.traceId | metadata.signature.signature      | metadata.signature.protected      | metadata.custom.key | metadata.antiReplay.iat | metadata.antiReplay.exp | metadata.antiReplay.jti |
      | fdc3.instrument | Apple        | test-app              | test-instance              | test-trace-id    | test-signature (signature part) | test-signature (protected part) | value               | {1234}                 | {2345}                 | test-jti                |
```

julianna-ciq (changes requested):
The schema invariant and removal of fabricated provenance are good changes, but the proxy must enforce both invariant branches and the tests should lock down the schema matrix and complete metadata preservation. Details are inline.

I have not duplicated the conformance-runner depcheck fix here because #2198 already carries the package-local configuration for the same shared polyfills; this branch should pick that up when the overlapping change is merged or rebased.

toolbox/fdc3-conformance/package.json
```diff
@@ -20,7 +20,9 @@
     "@finos/fdc3": "3.0.0-alpha.2",
     "buffer": "^6.0.3",
     "chai": "^6.2.2",
+    "events": "^3.3.0",
     "mocha": "^11.8.0",
+    "path-browserify": "^1.0.1",
```
julianna-ciq, line 25: See comment: https://github.com/finos/FDC3/pull/2193/changes#r3982211064

julianna-ciq: Note - looks like this is being handled in #2198

julianna-ciq:
FYI @kriswest — I opened #2214 with the changes from my review and the CI/depcheck fixes, in case taking them as a stacked PR is more convenient.

kriswest:
Conflicts based on line endings to be resolved in here by @julianna-ciq

julianna-ciq (approved):
lgtm

packages/fdc3-schema/schemas/api/getCurrentContextResponse.schema.json
kriswest: this was done

kriswest (commented):
@Roaders could you approve this one - @julianna-ciq has the last push on it

Roaders (approved):

PR #2199, opened 2026-09-09
fix(fdc3-conformance): fix pre-existing type errors and enable typecheck

### LINKED ISSUE #2194: fdc3-conformance: pre-existing TypeScript type errors are hidden because the build uses esbuild (no typecheck)
`toolbox/fdc3-conformance` is built with Vite/esbuild (`vite build`), which transpiles per-file and does **not** perform type checking. As a result, several genuine TypeScript type errors currently exist in the package's source but are never surfaced by the build or CI, so they can silently rot and mask real regressions.

These were discovered incidentally while adding the `clearContext` conformance tests (PR for #2164/#2165/#2187). They are **out of scope** for that PR and are not regressions introduced by it — they exist on `main` today. This issue tracks fixing them and closing the checking gap.

## How to reproduce

The package's own `tsconfig.json` sets `"moduleResolution": "node"` (node10), which the repo's current TypeScript version rejects when `tsc` is invoked directly:

```
error TS5108: Option 'moduleResolution=node10' has been removed. Please remove it from your configuration.
```

Running a type check with a modern resolution (e.g. `moduleResolution: "bundler"`, `noEmit: true`) surfaces the errors below.

## Type errors found

1. **`src/test/support/intent-support.ts`** — `IntentResult` union member access without narrowing. `intentResult` is typed `Channel | Context | ContextWithMetadata`, but the code reads `.type` and `.id` directly (e.g. lines ~178–179, ~195–196, ~204):
   ```
   Property 'type' does not exist on type 'Channel | Context | ContextWithMetadata'.
     Property 'type' does not exist on type 'ContextWithMetadata'.
   Property 'id' does not exist on type 'Channel | Context | ContextWithMetadata'.
   ```
   Needs a type guard / narrowing before accessing `type`/`id` (or an appropriate cast per branch of the `IntentResultType` switch).

2. **`src/test/basic/fdc3.basic.ts`** — unused import:
   ```
   error TS6133: 'Listener' is declared but its value is never read.
   ```
   The `Listener` import on line 1 should be removed.

3. **`src/test/index.ts`** — missing type declaration for the CSS side-effect import (line 5):
   ```
   error TS2882: Cannot find module or type declarations for side-effect import of 'mocha/mocha.css'.
   ```
   Needs an ambient module declaration (e.g. `declare module '*.css';` in a `.d.ts`) so the `import 'mocha/mocha.css'` typechecks.

> Note: the number of surfaced errors may vary slightly with the `moduleResolution` used to run the check; the three above are the substantive ones and are independent of any recently added code.

## Proposed resolution

1. Fix the individual type errors listed above (narrow the `IntentResult` union, drop the unused import, add a CSS ambient declaration).
2. Update `tsconfig.json` to a `moduleResolution` supported by the repo's TypeScript version (e.g. `"bundler"`), so `tsc` can run against the package.
3. Add a type-check step so this can't regress silently — either:
   - a `typecheck` npm script (`tsc --noEmit -p tsconfig.json`) wired into the package `build` (`tsc --noEmit && vite build`) and/or the monorepo CI, matching how other packages guard types; or
   - at minimum, a standalone `typecheck` script run in CI for `toolbox/fdc3-conformance`.

## Acceptance criteria

- `tsc --noEmit` (or the new `typecheck` script) passes for `toolbox/fdc3-conformance` with no errors.
- The check runs in CI so future type errors fail the build rather than being hidden by esbuild.
- No behavioural change to the conformance tests themselves.

## Area of issue

- FDC3 for Web Browsers
- Conformance Framework / Tooling

jabrailkhalil:
## Description

Closes the typechecking gap for `toolbox/fdc3-conformance` (built with Vite/esbuild, which never type-checks) and fixes the three pre-existing TypeScript errors surfaced when `tsc` is run directly:

1. **`src/test/support/intent-support.ts`** — `IntentResult` is a union (`Context | ContextWithMetadata | Channel | void`); `validateIntentResult` accessed `.type` / `.id` without narrowing. Narrowed per `IntentResultType` branch (`Context`, `Channel`, `PrivateChannel`) before member access.
2. **`src/test/basic/fdc3.basic.ts`** — removed the unused `Listener` import.
3. **`src/test/index.ts`** — added `src/css.d.ts` with `declare module '*.css';` for the `import 'mocha/mocha.css'` side-effect import (`TS2882`).

Additionally:

- `tsconfig.json`: `moduleResolution` changed from `"node"` (node10, removed in the repo's TypeScript version, TS5108) to `"bundler"`; added `noEmit: true` (build is done by Vite).
- `package.json`: added `typecheck` (`tsc --noEmit -p tsconfig.json`), wired into `build` (`npm run typecheck && vite build ...`) and into a new `test` script so the existing per-PR CI workflow (`npm run test --workspaces` in `.github/workflows/coverage.yml`) runs it on every PR. Future type errors fail the build instead of being hidden by esbuild.

## Checklist

- [x] No behavioral change to the conformance tests themselves
- [x] `npm run typecheck --workspace=fdc3-conformance` passes with 0 errors (Node 22, npm workspaces)
- [x] `npm run build --workspace=fdc3-conformance` passes (typecheck + `vite build` + copy)
- [x] Prettier clean on all touched files

Fixes #2194

## Follow-up: clean CI typecheck

The workspace test run builds the dependencies but skips the `@finos/fdc3` roll-up because that package has no test script. The conformance test script now builds that one workspace before running typecheck. This fixes the missing `@finos/fdc3` declarations and cascading type errors from the review without changing any conformance-test behavior.

Reproduced the failure on a fresh checkout with `npm ci` and the workspace test run. With this one-line change, `npm run test --workspace=fdc3-conformance` and `npm run build --workspace=fdc3-conformance` pass. Prettier also passes.

### DIFF
--- a/toolbox/fdc3-conformance/package.json
+++ b/toolbox/fdc3-conformance/package.json
@@ -15,7 +15,7 @@
     "build": "npm run typecheck && vite build && npm run copy && shx cp ../../node_modules/mocha/mocha.css ./dist/lib/mocha.css",
     "serve": "npx http-server ./dist -p 3001 -c-1 --cors",
     "dev": "npm run clean && npm run build && npm run serve",
-    "test": "npm run typecheck",
+    "test": "npm run build --workspace=@finos/fdc3 && npm run typecheck",
     "clean": "rimraf dist"
   },
   "dependencies": {
--- a/toolbox/fdc3-conformance/package.json
+++ b/toolbox/fdc3-conformance/package.json
@@ -11,9 +11,11 @@
   "license": "Apache-2.0",
   "scripts": {
     "copy": "shx mkdir -p ./dist && shx cp -R static/* ./dist",
-    "build": "vite build && npm run copy && shx cp ../../node_modules/mocha/mocha.css ./dist/lib/mocha.css",
+    "typecheck": "tsc --noEmit -p tsconfig.json",
+    "build": "npm run typecheck && vite build && npm run copy && shx cp ../../node_modules/mocha/mocha.css ./dist/lib/mocha.css",
     "serve": "npx http-server ./dist -p 3001 -c-1 --cors",
     "dev": "npm run clean && npm run build && npm run serve",
+    "test": "npm run typecheck",
     "clean": "rimraf dist"
   },
   "dependencies": {
--- a/toolbox/fdc3-conformance/src/css.d.ts
+++ b/toolbox/fdc3-conformance/src/css.d.ts
@@ -1,0 +1,6 @@
+/**
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright FINOS FDC3 contributors - see NOTICE file
+ */
+
+declare module '*.css';
--- a/toolbox/fdc3-conformance/src/test/basic/fdc3.basic.ts
+++ b/toolbox/fdc3-conformance/src/test/basic/fdc3.basic.ts
@@ -1,4 +1,4 @@
-import { Context, DesktopAgent, FDC3_VERSION, Listener, versionIsAtLeast } from '@finos/fdc3';
+import { Context, DesktopAgent, FDC3_VERSION, versionIsAtLeast } from '@finos/fdc3';
 
 import { APIDocumentation } from '../support/apiDocuments';
 import { ContextType, Intent } from '../support/intent-support';
--- a/toolbox/fdc3-conformance/src/test/support/intent-support.ts
+++ b/toolbox/fdc3-conformance/src/test/support/intent-support.ts
@@ -170,13 +170,14 @@ export class RaiseIntentControl {
     switch (expectedIntentResultType) {
       case IntentResultType.Context: {
         if (expectedContextType) {
+          const context = intentResult as Context;
           expect(
             intentResult,
             `The promise received by Test from resolution.getResult() should resolve to a ${expectedContextType} instance`
           ).to.have.property('type');
           expect(
-            intentResult?.type,
-            `The promise received by Test from resolution.getResult() should resolve to a ${expectedContextType} instance. Instead resolved to ${intentResult?.type}`
+            context?.type,
+            `The promise received by Test from resolution.getResult() should resolve to a ${expectedContextType} instance. Instead resolved to ${context?.type}`
           ).to.be.equal(expectedContextType);
           break;
         }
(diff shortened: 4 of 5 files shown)


kriswest:
Needs a manual test in the FDC3 web reference implementation before merging

kriswest (changes requested):
Thrse changes test up fine in the web reference implementation (all conformance tests passed) however the typecheck build is broken. See the errors reported at the end of CI log: https://github.com/finos/FDC3/actions/runs/34476939707/job/103300703015?pr=2199

If you can resolve those build errors this will be good to merge.

jabrailkhalil:
Thanks @kriswest! The typecheck/build blocker is fixed in the current head, 43a55f4: the conformance test script now builds the @finos/fdc3 workspace first so its declarations are available. Locally, `npm run test --workspace=fdc3-conformance` (including typecheck) and `npm run build --workspace=fdc3-conformance` both pass. Could you please re-review?

jabrailkhalil:
The CI workflows for the current head `43a55f456d` are waiting for contributor workflow approval (`action_required`): [Dependency Check](https://github.com/finos/FDC3/actions/runs/34623723968), [Node.js CI Tests and Coverage](https://github.com/finos/FDC3/actions/runs/34623724342), [Static code analysis](https://github.com/finos/FDC3/actions/runs/34623724359). Could a maintainer approve the pending runs when convenient? These runs have not executed their checks yet; I will address any failures once they run.

The typecheck/build correction remains in `43a55f4`, following the reported successful manual web-reference conformance run. Once these checks run, the remaining step is re-review of that correction.

toolbox/fdc3-conformance/package.json
```diff
@@ -11,9 +11,11 @@
   "license": "Apache-2.0",
   "scripts": {
     "copy": "shx mkdir -p ./dist && shx cp -R static/* ./dist",
-    "build": "vite build && npm run copy && shx cp ../../node_modules/mocha/mocha.css ./dist/lib/mocha.css",
+    "typecheck": "tsc --noEmit -p tsconfig.json",
+    "build": "npm run typecheck && vite build && npm run copy && shx cp ../../node_modules/mocha/mocha.css ./dist/lib/mocha.css",
     "serve": "npx http-server ./dist -p 3001 -c-1 --cors",
     "dev": "npm run clean && npm run build && npm run serve",
+    "test": "npm run build --workspace=@finos/fdc3 && npm run typecheck",
```
kriswest: @jabrailkhalil rather than triggering the whole workspace build from fdc3-conformance I think it would make more sense to do that between the lint and test steps of the overall workflow as an explicit step, here: https://github.com/finos/FDC3/blob/2dc194fd87614f4be838b5d352991980087252a9/.github/workflows/coverage.yml#L41-L45

I think you would also need to swap the order of the build and test steps in the release workflow here: https://github.com/finos/FDC3/blob/2dc194fd87614f4be838b5d352991980087252a9/.github/workflows/release.yml#L39-L43

At that point we don't need this test workflow asthe build step already runs typecheck (which is a build time task anyway). Freeing up the the test task here would allow us to setup a task to run the conformance tests in CI (another PR for doing that is already open and would conflict with this).

kriswest (changes requested):
Running the whole workspace build from a component of it seems like a bad idea. Better would be to run the workspace build as an explicit step in the coverage workflow.

jabrailkhalil:
Addressed the requested CI ordering in `cb8bb4c`: the root coverage and release jobs now build the workspaces before running tests, and the package-level script no longer hides build/typecheck inside a test command. Root build and the full test suite pass locally.

kriswest (approved):
LGTM

PR #2214, opened 2026-09-11
Validate retained context response invariants

julianna-ciq:
Addresses the review feedback on #2197 and hardens the retained-context response contract.

- Enforces the context/metadata coupling in the proxy implementation.
- Adds schema and proxy regression coverage for valid and invalid response shapes.
- Reuses the same depcheck alias exclusions as the #2193 follow-up to reduce future merge conflicts.

This PR is intentionally stacked on #2197.


kriswest (approved):
LGTM

kriswest:
@julianna-ciq thanks for this, merged.

BTW your PRs are coming in with line endings changes. Not check if its you or me thats misconfigured but think its on your end (let me know if not). Your git client should be sorting that out.

PR #2202, opened 2026-09-10
fix(example-apps): add trace ID to fallback metadata

julianna-ciq:
## Describe your change

Adds a generated `traceId` to the security demo's fallback `ContextMetadata`. #2198 makes `ContextMetadata.traceId` required, so the existing fallback caused the Netlify TypeScript build to fail with TS2741.

This PR is stacked on #2198 and targets `2163-appprovidable-metadata-optional`.

## Testing

- `npm run build`

## Contributor License Agreement

- [x] I acknowledge that a contributor license agreement is required and that I have one in place or will seek to put one in place ASAP.

## Review Checklist

- [x] **Issue**: Addresses the build regression in #2198.
- [ ] **CHANGELOG**: Not applicable; this is a build fix for the parent PR.
- [ ] **API changes**: No.
- [ ] **Context types**: No.
- [ ] **Intents**: No.

PR #2198, opened 2026-09-09
fix(dacp): clarify and align optional AppProvidableContextMetadata in request payloads (#2163)

kriswest:
## Describe your change

Resolves the inconsistency described in #2163, where the public `broadcast()`, `open()`, `raiseIntent()` and `raiseIntentForContext()` methods accept an **optional** `AppProvidableContextMetadata` argument but the corresponding DACP request payloads **required** their `metadata` field, leaving independent implementations unsure whether to omit `payload.metadata` or synthesise `metadata: {}`.

### Chosen resolution

The preferred option from the issue: make `metadata` **optional** on the wire and omit it when the app supplies no metadata argument.

- **Schemas**: `metadata` is no longer required in `broadcastRequest`, `openRequest`, `raiseIntentRequest` and `raiseIntentForContextRequest`. Each field's description now clarifies that it carries only the app-provided portion of the metadata and MUST be omitted when the app supplied no argument. Generated types regenerated (all four payloads now `metadata?: AppProvidableContextMetadata`).
- **Reference proxy**: omits the field (`...(metadata && { metadata })`) across all four calls instead of sending `metadata ?? {}`.
- **Reference web implementation**: already tolerated an absent `metadata` (all handlers use `payload.metadata ?? {}`), covered by tests.

### traceId is now the Desktop Agent's responsibility

Previously the proxy fabricated a `traceId` (via `uuid v4()`) on `raiseIntent`/`raiseIntentForContext`, which meant those requests always carried a metadata object even with no app input — inconsistent with `broadcast`/`open`.

- The proxy no longer generates a `traceId`; it forwards only app-provided metadata (or omits it entirely).
- It is now a **normative MUST** that the Desktop Agent provides a `traceId` in delivered `ContextMetadata`: it forwards an app-supplied `traceId` unchanged, and generates one when the app did not supply it, so receiving apps can always rely on a `traceId` being present. The reference web implementation already did this for broadcast, open and intent delivery.
- `ContextMetadata.traceId` is now a **required** property (overriding the optional `AppProvidableContextMetadata.traceId`), matching the wire schema which already required `traceId` on delivered event metadata. Spec, `spec.md` compliance requirements, `ContextMetadata.ts` and the `Types` reference doc were updated to MUST wording.

### Documentation

[...]

PR #2215, opened 2026-09-11
Test omitted metadata in proxy requests

julianna-ciq:
Adds regression coverage ensuring open and raise-intent proxy requests omit optional metadata when none is supplied.

This PR is intentionally stacked on #2198. The separate Netlify fallback-metadata fix was merged as #2202.


kriswest (approved):
LGTM

PR #2219, opened 2026-09-14
Clarify open timeout behavior when context is provided

julianna-ciq:
## Summary

- clarify that `fdc3.open` only waits for a launched app to initialize FDC3 when context is provided
- allow calls without context to resolve as soon as the app launches, without implying that FDC3 is available in the opened app
- document passing `{ type: 'fdc3.nothing' }` when callers expect channel support or otherwise want to confirm FDC3 availability
- distinguish `OpenError.ApiTimeout` (context supplied, app does not initialize FDC3) from `OpenError.AppTimeout` (app initializes but does not add the matching context listener)
- update the reference web implementation and add unit and conformance coverage for both context and no-context behavior

## Validation

- `npm test --workspace toolbox/fdc3-for-web/fdc3-web-impl` (123 tests passed)
- `npm run build --workspace toolbox/fdc3-conformance`
- `docusaurus build`
- Prettier check on all changed TypeScript and documentation files
- ESLint on all changed TypeScript files


kriswest (approved):
LGTM, thanks for the updates

PR #2232, opened 2026-09-18
fix(fdc3-schema): stop bundling the Convert validation runtime into every getAgent() consumer

### LINKED ISSUE #1901: Bundle bloat: importing from @finos/fdc3 pulls the entire @finos/fdc3-schema runtime (~240 KB)
## Minor Issue

Upgrading `@finos/fdc3` from 2.1.x to 2.2.3 grew our minified ESM bundle by **181 KB**.
Marking `@finos/*` as `sideEffects: false` at resolve time in our bundler narrowed the
regression to **158 KB** but didn't fix it — the remainder is structural (see details below).

### Area of Issue

- [ ] App Directory
- [ ] API
- [ ] Context Data
- [ ] Intents
- [ ] Desktop Agent Bridging
- [ ] Use Cases
- [x] Other

### Issue Description

Importing only `getAgent` / `fdc3Ready` from `@finos/fdc3` pulls in the entire
`@finos/fdc3-schema` runtime (~240 KB), even when the consumer never references
`BrowserTypes` or `BridgingTypes`.

#### Root cause

`@finos/fdc3-schema/dist/src/index.js`:

```js
import * as BrowserTypes from '../generated/api/BrowserTypes.js';
import * as BridgingTypes from '../generated/bridging/BridgingTypes.js';
export { BrowserTypes, BridgingTypes };
```

The `import * as` namespace binding is opaque to bundlers — any reachability of
`BrowserTypes` retains the whole module. The two generated modules are ~146 KB and
~94 KB of quicktype output (`Convert` classes plus type-guard helpers).

The umbrella `@finos/fdc3` then re-exports both namespaces, so any consumer that imports
from it makes the schema reachable:

```js
// @finos/fdc3/dist/src/index.js
import { BrowserTypes, BridgingTypes } from '@finos/fdc3-schema';
export { BridgingTypes, BrowserTypes };
```

Within `@finos/fdc3-get-agent`, the modules that introduce a runtime import of
fdc3-schema (and therefore force it into any reachable graph) are:

- `strategies/HelloHandler`
- `strategies/IdentityValidationHandler`
- `ui/AbstractUIComponent`
- `ui/DefaultDesktopAgentChannelSelector`
- `ui/DefaultDesktopAgentIntentResolver`

(`@finos/fdc3-agent-proxy/listeners/PrivateChannelEventListener` is in the same
position.)

#### Options for cutting it down

1. **Add `"sideEffects": false`** to every `@finos/*` package. None of the six 2.2.3
   packages declare it today.

2. **Replace `import * as` with named re-exports** in
   `@finos/fdc3-schema/dist/src/index.js` so bundlers can drop symbols at export-name
   granularity instead of keeping the entire 146 KB / 94 KB namespaces because
   anything in them was reached.

3. **Add an `exports` map with sub-paths** to `@finos/fdc3-schema` so consumers can
   opt in to the narrowest slice (e.g. types vs. validators) without bundlers having
   to prove the rest is unused.

4. **Split the generated `Convert` class out of the type-only exports.** Most of the
   240 KB is the `Convert` runtime (162 static methods on `BrowserTypes` alone) plus
   its type-guard helpers. Consumers wanting only TypeScript types pay the full
   runtime cost. A separate entry point for validators would let consumers that don't
   need runtime schema validation skip it entirely.

patrickboneill:
resolves #1901

## Summary

Importing from `@finos/fdc3` pulled the entire `@finos/fdc3-schema` runtime (~240KB) into every consumer's bundle, even ones that only call `getAgent()`. The bulk of that weight is the quicktype-generated `Convert` class (full JSON validation/serialization for every message type) plus the `isValid<X>` predicates that call into it — nothing internal to this repo uses either; `fdc3-get-agent`, `fdc3-agent-proxy` and `fdc3-web-impl` only ever use the fast `is<X>` predicates and the plain message interfaces.

This follows the direction discussed by @kriswest in the issue: split the validation runtime (`Convert` class + `isValid<X>`) out of the generated files, since it's the part nothing internal needs but that made the whole namespace unavoidably reachable (and therefore un-tree-shakeable) for every consumer.

* `generate-type-predicates.ts` now splits each quicktype-generated file into the slim types file (interfaces + fast `is<X>` + `<X>_TYPE` consts) and a companion `*Validation.ts` file (`Convert` class + its runtime helpers/type map + `isValid<X>`), so bundlers simply never see the validation runtime unless something imports it.
  * `BrowserTypes.ts` compiled JS: 155KB → 40KB
  * `BridgingTypes.ts` compiled JS: 95KB → 8KB
* `fdc3-schema` and `fdc3`'s `index.ts` export the new `BrowserTypesValidation` / `BridgingTypesValidation` namespaces for consumers that do need full validation. This is the one breaking surface: `BrowserTypes.isValidX` / `BrowserTypes.Convert` move to `BrowserTypesValidation.isValidX` / `BrowserTypesValidation.Convert`.
* Added `"sideEffects": false` to all 9 `@finos/*` packages (none had it), after auditing every package's `src/` for module-load side effects (global mutation, DOM/CSS effects, side-effect-only imports) — none found.
* Declared the previously-phantom `@finos/fdc3-schema` dependency in `fdc3-agent-proxy` and `fdc3-web-impl`, which were relying on npm workspace hoisting for it.

Regenerating the schema output also picked up an already-merged schema change (#1810 - `contextType`/`contextTypes`) that the committed generated files hadn't been regenerated for yet since that PR merged; that's pre-existing drift, not a new schema change.

## Bundle size verification

Bundled `import { getAgent } from '@finos/fdc3'; getAgent();` with esbuild (`--bundle --minify`), before vs. after this change:

| | Before | After | Change |
|---|---|---|---|
| Minified | 187.8 KB | 66.6 KB | **−64.5%** |
| Gzipped | 26.9 KB | 15.0 KB | **−44.3%** |

## Test plan

- [x] `npm run build` from repo root — full monorepo build passes
- [x] `packages/fdc3-schema` — `vitest run` (7/7 passing)
- [x] `packages/fdc3-get-agent` — `vitest run` (38/38 passing)
- [x] `packages/fdc3-agent-proxy` — `vitest run` passing
- [x] `packages/fdc3-standard` — `vitest run` (16/16 passing)
- [x] `packages/fdc3-security` — jest (45/45 passing)
- [x] `toolbox/fdc3-for-web/fdc3-web-impl` — `vitest run` (135/135 passing)
- [x] `eslint` clean on all touched hand-written source
- [x] Runtime smoke test confirming `BrowserTypes.Convert` is now `undefined`, `BrowserTypesValidation.Convert`/`isValidX` work correctly (including rejecting a malformed payload), and `BrowserTypes.isX` fast predicates are unaffected


kriswest:
@hughtroeger & @enewsome can you take a look at this and let us know if its sufficient to deal with the issue you raised? Any review comments?

kriswest:
@patrickboneill could you resolve the conflict with packages/testing/package.json by rebasing on main and running an `npm install` afterwards?

CHANGELOG.md
Roaders, line 1: I think we need an additional changelog entry noting the breaking change of moving functions from `BrowserTypes` to `BrowserTypesValidation`

patrickboneill: https://github.com/finos/FDC3/pull/2232/changes/3a044e7bd4f0951e37002915718550fec33db293 Added an update to the changelog

